PR #2, opened 2016-07-26
Add Rx.js and hateoas helpers

OzzieOrca:
@adammeyer let me know your thoughts. I was trying to make it easier to grab and follow links through the hateoas API. Let me know if you think we should use JSONPath more. Maybe we should make some wrappers/shortcuts for it so we don't have to add the `[0]` everywhere and make it easier to grab links. I've also added a little bit to `payment` branch but not a whole lot yet.
- Add Rx.js
- Use Observable map() and subscribe() instead of promise then()
- Use forkJoin() instead of $q.all()
- Change apiService to class
- Have http helper methods change the method and pass the object on
- Add zoom and followLocation options to api config
- Map response object to zoom object
- Add hateoasHelpers service to pull links and  data out of response object using paths and zoom

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
       "json": "github:systemjs/plugin-json@^0.1.2",
       "jsonpath": "npm:jsonpath@^0.2.6",
       "lodash": "npm:lodash@^4.13.1",
+      "rxjs": "npm:rxjs@^5.0.0-beta.10",
       "scss": "github:KevCJones/plugin-scss@^0.2.11",
       "ui-router-extras": "npm:ui-router-extras@^0.0.14"
     },
--- a/src/app/cart/cart.component.js
+++ b/src/app/cart/cart.component.js
@@ -21,7 +21,7 @@ class CartController{
 
   loadCart(){
     this.cartService.get()
-      .then((data) => {
+      .subscribe((data) => {
         this.$log.info('cart', data);
       });
   }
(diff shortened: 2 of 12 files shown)


adammeyer:
I like the idea of adding the zoom to options on the api and having the service to parse it.  Good idea!  It looks like it should take in an array, but I noticed the cart service is still sending a string.  What am I missing there?

I think we should continue to use the zoom object.  I'm not a fan of making multiple requests via the hateoas links.

What are the benefits of using rx.js?  Looks like it's the same as angular, just replacing then with subscribe.  Will it add any value?  If there isn't a need for it, I would just say stick with the Angular promises.

OzzieOrca:
I didn't go through the rest of the app and add the `zoom` directly to the root object. The cart service is still using `params` and is passing a string there.

OzzieOrca:
Here's some reasons why I like Rx.js:
- It will make migration to Angular 2 easier as `$http` uses Observables in the new framework
- It is a lot more flexible and more concise
- It can make catching errors and retrying requests easier (potentially with a sign in step inbetween)
- It can send off multiple requests but keep their data processing in order and ignore previous data (search box that searches as you type) [mpdx example](https://github.com/CruGlobal/mpdx/pull/990/files#diff-deaebea26c64a78a2071fc322bfadd7bR106)
- It has lots of nice convenient [operators](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) to process data on the stream: `debounce`, `retry`, `reduce`, `mergeMap` etc.
- It can map one response into many items in the stream (`mergeMap`) and you can process and handle those in the stream

adammeyer:
Thanks for the reasons.  👍

PR #3, opened 2016-07-26
Cart

adammeyer:


adammeyer:
@OzzieOrca This is ready for your review.  Thanks!

src/app/main/main.component.js
```diff
@@ -45,11 +47,7 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
       template: '<checkout></checkout>'
     });
 
-  $locationProvider.html5Mode({
```
adammeyer: I removed this so it will run on s3 without any configuration changes needed.

src/app/cart/cart.component.js
```diff
@@ -11,9 +11,7 @@ let componentName = 'cart';
 
 class CartController{
 
-  /* @ngInject */
```
OzzieOrca: I think you need to keep this for minification

src/app/main/main.component.js
```diff
@@ -19,15 +19,17 @@ let componentName = 'main';
...
     this.cartService = cartService;
     this.$log = $log;
+    this.$state = $state;
   }
 
   addItemToCart(id){
     this.cartService.addItem(id)
       .subscribe((response) => {
         this.$log.info('Added new item', response.data);
+        this.$state.reload();
```
OzzieOrca: This won't work on AEM right?

adammeyer: Correct.  But we won't use the main component on AEM.

OzzieOrca: Oh right that's in the main component...

PR #4, opened 2016-07-27
Switch Give_Front-end to checkout branch and use SCSS from there

OzzieOrca:
@adammeyer What do you think of this? Looks like David Sudarma has continued to make css changes. Running `jspm update github:CruGlobal/Give_Front-end` will update the css to the latest. I also like using it to copy templates since the files are already in intellij.


adammeyer:
Looks good. 👍

PR #5, opened 2016-07-28
checkout step 1 validation

adammeyer:


src/app/checkout/step-1/step-1.component.js
```diff
@@ -67,6 +69,6 @@ export default angular
     controller: Step1Controller,
     templateUrl: template.name,
     bindings: {
-      changeStep: '&'
```
adammeyer: @OzzieOrca I wasn't able to get the function to change the step, so I changed it to bind the variable.  Any downsides to this?

adammeyer:
@OzzieOrca I'm merging this now, but let me know if you have any changes and I'll correct them.  thanks!

src/app/checkout/step-1/step-1.component.js
OzzieOrca: It adds watchers... Not a big deal. I think you need to call it like [this](https://github.com/CruGlobal/give-web/pull/6/files#diff-723383532014b6b0bdf61e792b226b2cR64). 

So you'd use it like this:
```
<my-component change-step="$ctrl.parentComponentFunction(param1)"></my-component>
```
Parent component controller:
```
parentComponentFunction(someParam1){
  //do something
}
```
Child component:
```
this.changeStep({param1: 'value})
```

For some reason it uses named parameters so the key in the child component has to match what you put in the template. I guess it has to parse the function in the template as a string and then call it.

adammeyer: I gave that a try this morning, but I couldn't figure it out.  I think there must be something missing.

PR #10, opened 2016-08-05
Better payment validation

OzzieOrca:
Most important changes were updating ccp, refactoring validation around its new api, and using ccp to mask bank account numbers.

My last commit might seem like it redoes some of the things I did in the first commit as I was figuring out the best way to do things.


adammeyer:
👍  Good to merge.  Will anything need to be updated for step-1 validation?

OzzieOrca:
I don't think so. This just moved some checks into separate validators to be able to show individual error messages and updated how some of the validation was done.

PR #11, opened 2016-08-05
Remove ui-router, create html pages for routes

adammeyer:
Remove ui-router and main component.
Created cart.html and checkout.html to simulate AEM setup.


src/app/cart/cart.component.js
```diff
@@ -9,6 +10,10 @@ import designationsService from 'common/services/api/designations.service';
 import template from './cart.tpl';
 import templateModal from 'app/productConfig/productConfigModal.tpl';
 
+//local dev
+import localDevNav from 'common/localDev/nav/local-dev-nav.component';
```
OzzieOrca: The cart component shouldn't depend on dev stuff right? Otherwise it'll end up in production code. The main module pulled those components together before. We could use a main component that supports transclusion to easily choose what components go inside.

OzzieOrca:
Also what's the motivation for this?

adammeyer:
I created a localDev component.  That way the others will stay free of the local dev stuff.

I would like the dev env to function like site will in AEM, using links instead of UI router.

build/tasks/release.js
```diff
@@ -6,33 +6,18 @@ var $ = require('gulp-load-plugins')({
 var paths = require('../paths');
 
 gulp.task('cache-bust', function () {
-  var cacheBust = "var systemLocate = System.locate; System.locate = function(load) { var cacheBust = '?bust=' + " + Math.round(new Date() / 1000) +"; return Promise.resolve(systemLocate.call(this, load)).then(function(address) { if (address.indexOf('bust') > -1 || address.indexOf('css') > -1 || address.indexOf('json') > -1) return address; return address + cacheBust; });}\n"
   return gulp.src('dist/app/app.js')
     .pipe($.insert.prepend("window.prod = true;\n"))
-    .pipe($.insert.prepend(cacheBust))
```
OzzieOrca: The `window.prod` isn't used anywhere and should be removed. I'm using `angular-environment` instead. If you want to remove the cache-bust stuff, this whole task should be removed.

OzzieOrca:
I think having a place to test the compiled bundles in isolation would be good eventually. I think most of this is figuring out what HTML to serve and where and what it should include. We can pick and choose which JS bundles to include as much as we want. Part of this complication is that we have a main/dev bundle that currently has the dummy header (maybe it doesn't need to be here but I guess it makes it look more like prod) and our random dev buttons that should go away anyways (but maybe not).

### Here's what I think we need:
#### 1. A dev environment that works locally and on S3
This will have our dev/main component and that will import all of our other root components and switch between them. To switch between them we could either:
- **A. Create individual HTML files** These files would have duplicated code importing our dev/main component with just a single line that is different (`<checkout></checkout>` instead of  `<cart></cart>` for example). They would only be used for dev/testing.
- **B. Use `ui-router`** We could have one main HTML file and add states to the router config. I'd argue that this is cleaner.

#### 2. Something that mirrors production and contains no dev code
This could be:
- **A. HTML files in this project** A folder of HTML files in this project that are copied to the dist folder that testers can navigate to on S3. These files would include `common.js` and then their respective bundle but not the main/dev component. Maybe these files or our dev files (number 1 above) should be in a subfolder on S3. We can't use `ui-router` for this as production can't use it.
- **B. AEM Pages** These HTML files would include the same thing as A. This would be closer to prod and we can test AEM specific things (analytics, custom content, injecting a cart link into the header, etc). Idk how soon we can/want to set this up.

I think what I'm arguing against is doing 1A **AND** 2A. I'm just trying to think through this and put together the bigger picture. That's 2 HTML files per component and configuring them differently (1 for dev and 1 for prod) and keeping them all updated as things change and creating tooling for both. There's no way around making separate HTML somewhere for number 2 as long as we are using multiple pages on AEM. I think using ui-router would make implementing number 1 cleaner and I don't think it adds much complexity. Either way, we need to document (probably in the readme) what bundles the release process outputs and how to use them.

**So these were my first thoughts when thinking about this PR:** (I tried to summarize them better above)

- It seems like you just renamed the `main` component to `localDev` and moved the component html (`<checkout></checkout>` or whatever) from ui-router to a root HTML file. All the JS is still interdependent. If you are relying on something included in localDev (say `import 'angular-bootstrap';`) but forget to import it in the actual component, it is going to fail on AEM no matter if you use this setup or the previous one. If we want to mirror production, we can't have the dev/main component involved.
- That's going to be a lot of manually created HTML files (and duplicated HTML) in this project as we add more root components (AEM routes). Maybe this kind of production testing should actually happen inside AEM with AEM importing compiled bundles only.
- I think we should get pushstate working on S3. I think adding an object to the router config is cleaner than creating lots of HTML files that aren't going to be used anywhere other than testing. Either a hack like something [here](http://stackoverflow.com/questions/16267339/s3-static-website-hosting-route-all-paths-to-index-html) since the HTML is never going live on S3 and we don't care about SEO there or using cloudfront or something else like Firebase Hosting (Free and provides pushstate support). This would allow us to still use `ui-router` to switch between components in dev faster and modify our main/dev component with one main file and we could still use the same relative routes that AEM will use for the pages (`give.cru.org/checkout` or `give.cru.org/checkout.html` or whatever). I agree that we should use `href`s to change pages instead of `ui-sref` which won't work in production.
- It looks like you removed the common bundle. I think that would be good to keep either way as it prevents all the common dependencies from being reloaded on every single page and it can be loaded once and cached and all the other bundles can use it without having to have all those deps duplicated in their individual bundles.
- We should figure out what to do with the CSS at some point. I guess this would be a better question for Sudarma. Should we bring the givesite specific SCSS into this project and compile it here or should that stay outside and we use the same CDN in this project and on AEM. Some of the generic Cru SCSS is probably already hosted on AEM somewhere.

Hopefully all this made sense. Sorry it got so long. I think I communicate better by writing things out and then trying to summarize it. Let me know what your thoughts are on it. I'll try to be around earlier in the morning to work on these PRs and help Brian get started in case he runs into anything. 😃

adammeyer:
Thanks @OzzieOrca!  My thoughts:

- When we get into reading JWT cookies for authorized user data (sometime this week), the s3 environment won't really be useful anymore.  Cookies require .cru.org domain.  I'm going to start on building the AEM components today.  So we and QA can do testing on AEM dev.

- For CSS, I think we'll need to pull the global CSS from cru.org and just use the give CSS in this project.  For AEM, I'll only want to import give css so there isn't redundancy.

- I'm fine with closing this and using what we have, but I think two things need to be solved.

 1. Linking, in AEM links will be something like "checkout.html".  How can we bridge the differences in the ui-router linking and AEM?  Maybe call the routes in UI-Router, "checkout.html" instead of just "checkout"?

 2. Before starting on this PR, I tried to get the common bundle to work.  I wasn't able to, it always required main.js to be included for System JS to run.

I'll close this PR for now.  Let me know your thoughts on these two items.  thanks!

PR #12, opened 2016-08-05
Better systemjs tooling

OzzieOrca:
Main motivation was to get code coverage working.

All of the js is compiled through systemjs using systemjs-plugin-babel. There is no longer a need to use `/dist` as a temp directory so I moved the release task to output the bundles directly in `/dist`.

One slightly annoying side effect is that the compiled `.tpl.js` and `.css` files are output in `/src`. CSS shouldn't matter much since is all external for this project. I think it could be fixed by using a systemjs loader to run scss and ngHtml2js but I haven't found a good solution yet.


adammeyer:
If the tpl.js is in src, will they be omitted from the s3 deploy?

OzzieOrca:
They'll be imported and bundled together with the rest of the JS.

I still need to fix the final bundle in this branch. It seems to be reaching outside the bundle files for stuff.

adammeyer:
Sounds good.  This looks good to me, so feel free to merge when ready.

PR #13, opened 2016-08-08
gift config modal validation

adammeyer:
Simple validation for amount (must be greater than $1)
Add max length of 250 for comments

### DIFF
--- a/src/app/productConfig/productConfigModal.tpl.html
+++ b/src/app/productConfig/productConfigModal.tpl.html
@@ -14,7 +14,7 @@
     </div>
   </div>
 </div>
-<form>
+<form novalidate name="$ctrl.itemConfigForm" ng-submit="$ctrl.addToCart()">
   <div class="modal-body pt0 p">
     <div class="container-fluid">
       <div class="panel panel-default panel-plain mt">
@@ -34,7 +34,7 @@
               <!--<label><input id="" name="custom_option" type="radio" value="Custom003"> Custom Ministry Option 3</label>-->
             <!--</div>-->
 
-            <div data-toggle="buttons">
+            <div data-toggle="buttons" validation="$ctrl.itemConfigForm.amount" show-error="asClass">
               <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 50" ng-class="{'active': $ctrl.itemConfig.amount === 50}"><span class="number">$50</span></label>
               <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 100" ng-class="{'active': $ctrl.itemConfig.amount === 100}"><span class="number">$100</span></label>
               <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 250" ng-class="{'active': $ctrl.itemConfig.amount === 250}"><span class="number">$250</span></label>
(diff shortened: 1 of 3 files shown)


src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -43,9 +43,11 @@ <h4 class="panel-title border-bottom-small" translate>
               <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 5000" ng-class="{'active': $ctrl.itemConfig.amount === 5000}"><span class="number">$5000</span></label>
               <label class="btn btn-default-form u-textLeft custom-amount">
                 <div class="form-group form-group-default u-inline">
-                  <input class="form-control form-control-subtle number" min="0" step="any" type="text" ng-model="$ctrl.itemConfig.amount">
+                  <input class="form-control form-control-subtle number" min="1" type="number" name="amount" ng-model="$ctrl.itemConfig.amount" required>
                 </div>
               </label>
+              <span class="help-block" validation="$ctrl.itemConfigForm.amount" show-error="required" translate>You must enter an amount</span>
+              <span class="help-block" validation="$ctrl.itemConfigForm.amount" show-error="min" translate>Amount must be at least $1</span>
```
OzzieOrca: Are we ever going to accept non USD currencies? Not a blocker for this PR unless you can think of a generic error message.

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -125,12 +115,22 @@ <h4 class="panel-title border-bottom-small mt" translate>
           <div class="panel-body pt0">
             <ul class="list-unstyled">
               <li>
-                <a class="commentLink" href="" ng-click="showRecipientComments = !showRecipientComments" translate>Send a Message to {{$ctrl.productData.displayName}}</a>
-                <textarea class="staff-comments give-modal-comments" rows="3" ng-model="$ctrl.itemConfig['recipient-comments']" ng-if="showRecipientComments || $ctrl.itemConfig['recipient-comments']" style="display:block;"></textarea>
+                <a class="commentLink" href="" ng-click="showRecipientComments = !showRecipientComments"><translate>Send a Message to</translate> {{$ctrl.productData.displayName}}</a>
```
OzzieOrca: Will the translate directive not work? It seems to [support interpolation](https://angular-gettext.rocketeer.be/dev-guide/annotate/#interpolation).

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -125,12 +115,22 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
+                <textarea class="staff-comments give-modal-comments"
+                        rows="3"
+                        maxlength="250"
+                        ng-model="$ctrl.itemConfig['recipient-comments']"
+                        ng-if="showRecipientComments || $ctrl.itemConfig['recipient-comments']"
+                        style="display:block;"></textarea>
```
OzzieOrca: Does this need `display: block`?

src/app/productConfig/productConfigModal.tpl.html
adammeyer: Okay, I'll switch it back.  Thanks.

src/app/productConfig/productConfigModal.tpl.html
adammeyer: Yes, the css is hiding it because it used a jquery library to show it.

OzzieOrca:
Besides those minor questions, 👍

src/app/productConfig/productConfigModal.tpl.html
OzzieOrca: There's a second spot that has display block too.

OzzieOrca: Ah. Ya I've seen that elsewhere... K leave it

PR #14, opened 2016-08-08
Append .html to all routes

OzzieOrca:
@adammeyer How's this look? I still need to configure S3 but it would help to get this on there.


OzzieOrca:
Actually I could probably play with this on the dev branch. I'll do that for a bit.

adammeyer:
👍

PR #15, opened 2016-08-09
international address lines

adammeyer:
Adds 4 text boxes for non-US addresses.  Combines them all into the street address on save.

### DIFF
--- a/src/app/checkout/step-1/step-1.tpl.html
+++ b/src/app/checkout/step-1/step-1.tpl.html
@@ -172,7 +172,7 @@
                 <select class="form-control  form-control-subtle"
                         name="addressCountry" required
                         ng-model="$ctrl.donorDetails['mailing-address']['country-name']"
-                        ng-options="v['display-name'] | uppercase as v['display-name'] for v in $ctrl.countries"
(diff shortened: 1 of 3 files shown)


src/app/checkout/step-1/step-1.tpl.html
```diff
@@ -172,7 +172,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Mailing Address<
                 <select class="form-control  form-control-subtle"
                         name="addressCountry" required
                         ng-model="$ctrl.donorDetails['mailing-address']['country-name']"
-                        ng-options="v['display-name'] | uppercase as v['display-name'] for v in $ctrl.countries"
+                        ng-options="v['display-name'] | uppercase as v['display-name'] for v in $ctrl.countries | orderBy:'\'display-name\''"
```
OzzieOrca: Does this need to be double quoted? It just looks odd. I'm assuming you had an issue with passing the variable to orderBy since it has a dash.

OzzieOrca:
Looks good

src/app/checkout/step-1/step-1.tpl.html
adammeyer: correct.  The dash is dumb and breaks it, so I needed the double quotes.

PR #16, opened 2016-08-10
Add ngMessages

OzzieOrca:
I decided to use [ngMessages](https://docs.angularjs.org/api/ngMessages) to show the validation errors because it allows you to only show one message at a time and order which message shows up if there are multiple errors. Order is based on position in the DOM. I was having to return true if the value was empty or the wrong length in some of my specific payment validation functions to prevent them from showing too early.

I updated the productConfigModal and step 1 also.

I got rid of my validation directive and decided to make a filter that takes a ngModelController and returns true/false depending on whether the messages should be shown. This is then used in `ng-class` for the `has-error` and `ng-if` for showing/hiding the error messages.

https://github.com/CruGlobal/give-web/commit/b876eec533f2f62a942c9012df5b4065134a58be changes `maxlength` to `ng-maxlength` to allow input larger than maxlength but then show an error message. I think this gives a better user experience and they don't lose what they type but I doubt these will be hit very often and it adds complexity to the template. I did it thinking validation was missing entirely, forgetting that the input would stop the typing past that point. It would be easy to reverse that commit if you don't like it.

Let me know what you think.


src/app/checkout/step-1/step-1.tpl.html
```diff
@@ -176,72 +203,93 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Mailing Address<
...
+                    <div class="help-block" ng-message="required" translate>You must select a state / region</div>
+                  </div>
                 </div>
               </div>
             </div>
             <div class="col-sm-6">
-              <div class="form-group is-required" validation="$ctrl.detailsForm.addressPostalCode" show-error="asClass">
+              <div class="form-group is-required" ng-class="{'has-error': ($ctrl.detailsForm.addressPostalCode | showErrors)}">
                 <label translate>Zip / Postal Code</label>
                 <input type="text" class="form-control  form-control-subtle" name="addressPostalCode" ng-model="$ctrl.donorDetails['mailing-address']['postal-code']" required max="20">
```
adammeyer: max should be ng-maxlength here.  I must have miss-typed that one.

adammeyer:
Looks good! 👍

PR #17, opened 2016-08-10
Rename apiService to cortexApiService.

Omicron7:
This PR drops the cortex namespace from the default apiUrl and adds it to the cortex specific API calls.

Login process requires issuing api calls to the apiUrl + '/cas/' + path. A casApiService will be added to facilitate this.


OzzieOrca:
👍

adammeyer:
👍

PR #19, opened 2016-08-11
Add credit card validation and save details to cortex

OzzieOrca:
I've got the validators in place although I'm not sure what format the year is supposed to be in (2 digits or 4 digits). I couldn't find a Jira task with the specific validators required like I've seen with the bank accounts and contact info.

~~I seem to be getting a 400 bad request to CCP when I try to submit the details.~~ I added the CCP staging key to fix this.

### DIFF
--- a/src/app/checkout/step-2/credit-card/credit-card.tpl.html
+++ b/src/app/checkout/step-2/credit-card/credit-card.tpl.html
@@ -6,23 +6,22 @@
 
     <div class="row">
       <div class="col-sm-6">
-        <div class="form-group is-required" ng-class="{'has-error': ($ctrl.creditCardPayment.accountType | showErrors)}">
+        <div class="form-group is-required" ng-class="{'has-error': ($ctrl.creditCardPaymentForm.cardNumber | showErrors)}">
           <label>Card #</label>
-          <input type="text" name="cardNumber" class="form-control  form-control-subtle" ng-model="$ctrl.creditCardPayment.cardNumber" required ng-minlength="13" ng-maxlength="16">
-          <div ng-messages="$ctrl.creditCardPaymentForm.cardNumber.$error" role="alert" ng-if="$ctrl.creditCardPaymentForm.$submitted || $ctrl.creditCardPaymentForm.cardNumber.$touched">
+          <input type="text" name="cardNumber" class="form-control  form-control-subtle" ng-model="$ctrl.creditCardPayment.cardNumber" required>
+          <div ng-messages="$ctrl.creditCardPaymentForm.cardNumber.$error" role="alert" ng-if="($ctrl.creditCardPaymentForm.cardNumber | showErrors)">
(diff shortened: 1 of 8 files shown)


src/app/checkout/step-2/credit-card/credit-card.tpl.html
```diff
@@ -1,124 +1,153 @@
-<div class="mb">
+<form novalidate ng-submit="$ctrl.submitPayment(creditCardPayment.$valid)" name="$ctrl.creditCardPaymentForm">
```
adammeyer: Is creditCardPayment a form?

src/app/checkout/step-2/credit-card/credit-card.tpl.html
```diff
@@ -1,124 +1,153 @@
...
+                  <input type="text" class="form-control  form-control-subtle">
+                </div>
+              </div>
+            </div>
+            <div class="row">
+              <div class="col-sm-6">
+                <div class="form-group">
+                  <label>State</label>
+                  <div class="form-group">
+                    <select class="form-control  form-control-subtle" id="suffix">
```
adammeyer: Remove the suffix id

src/app/checkout/step-2/credit-card/credit-card.tpl.html
```diff
@@ -1,124 +1,153 @@
...
+      <div class="panel panel-default">
+        <div class="panel-body">
+          <div class="mb">
+            <h4 class="panel-title  border-bottom-small  visible">New Billing Address</h4>
+            <div class="row">
+              <div class="col-sm-12">
+                <div class="form-group">
+                  <label>Country</label>
+                  <div class="form-group">
+                    <select class="form-control  form-control-subtle" id="suffix">
```
adammeyer: Remove the suffix id

src/app/checkout/step-2/credit-card/credit-card.tpl.html
```diff
@@ -1,124 +1,153 @@
...
+                      <option>Peoples Republic of China</option>
```
adammeyer: Are you populating this country list?  You can use the geographies function from the order services to do that.

adammeyer:
Looks good! 👍   Just a few comments.

src/app/checkout/step-2/credit-card/credit-card.tpl.html
OzzieOrca: I actually haven't done anything with the billing address yet. It is all hidden currently. I guess I should work on that next. I think I'd prefer to tackle that in another PR.

src/app/checkout/step-2/credit-card/credit-card.tpl.html
OzzieOrca: I'm not using `on-submit` since the submit button is outside the form and component. It just got left there. I'll remove it from both the credit card and bank account templates.

PR #20, opened 2016-08-11
Check for undefined in showErrors

OzzieOrca:
I was seeing occasional undefined errors when switching payment types.


adammeyer:
👍

PR #21, opened 2016-08-11
Validate bank account number for min digits

OzzieOrca:
The user may have only entered spaces or dashes or something else.


adammeyer:
👍

PR #22, opened 2016-08-11
Add devpub.cru.org to list of staging domains

OzzieOrca:
@adammeyer Should this point at the staging apiUrl? I guess at this point it doesn't matter much as the development and staging urls are the same. If it is not listed, it will default to development.


adammeyer:
Yes, devpub and uatpub should both point to staging cortex.

OzzieOrca:
I added that. What are those domains for/what's the difference?

adammeyer:
I believe uatpub will be the same as stage.give.  I believe that's just an alias for uatpub.

adammeyer:
👍

PR #24, opened 2016-08-11
Step 2 component: add changePaymentType and more tests

OzzieOrca:
I was seeing an `undefined` error if you switched `paymentTypes` while `submitted` was true due to the form for the new payment type watching `submitted` and submitting that form before it was fully initialized.


adammeyer:
👍

PR #25, opened 2016-08-12
Add bank account component tests

OzzieOrca:
Also added the first fixture. I'm not sure where the best place to define them is but I think it makes sense for them to go in a folder where they are relevant to the services/components they help test.


adammeyer:
👍

PR #27, opened 2016-08-15
Add tests to paymentValidationService

OzzieOrca:
- Move CCP initialization to function
- Test CCP initialization with different keys
- Test getCardNumberErrorMessage


adammeyer:
👍

PR #28, opened 2016-08-16
Compile template in bank account spec and add tests

OzzieOrca:
Adds test for: 
- validity of entire form
- $validators
- $parsers
- component inputs
- component outputs
- savePayment failures


adammeyer:
👍

PR #29, opened 2016-08-16
Checkout step 1 corrections

adammeyer:
Fixed:
Primary Title does not have an empty selector
City and State (US) should not be Address 3 and 4 (International) - the US City and State are being shown as Address 3 and 4 when non-US Country is selected.
Country selector does not default to United States


OzzieOrca:
👍

PR #30, opened 2016-08-16
Checkout empty cart

adammeyer:
Hide checkout page if no cart items
Checkout step 1 loading indicators


OzzieOrca:
👍

I think I'll make a component for the loading indicators in another PR. It won't do much but I think 1 line of HTML is better than having 6 everywhere.

adammeyer:
Thanks, that would be great.

PR #31, opened 2016-08-16
Display payment details on review page (Step 3)

OzzieOrca:
- Add capitalize filter
- Load current payment details in step 3 and add method in order service to request them from cortex
- Change `<a>` tags to `<button>`


adammeyer:
👍

PR #32, opened 2016-08-16
Add loading component

OzzieOrca:
@adammeyer what do you think? It does add complexity in that is has to be imported but it's fewer, deduplicated lines of HTML. I think we would only have to do that in each root component.

### DIFF
--- a/src/common/components/loading/loading.scss
+++ b/src/common/components/loading/loading.scss
@@ -1,0 +1,3 @@
+.loading-wrap {
+  transform: translate(-50%, -50%);
+}
--- a/src/common/components/loadingOverlay/loadingOverlay.scss
+++ b/src/common/components/loadingOverlay/loadingOverlay.scss
@@ -11,11 +11,3 @@
   height: calc(100% + 20px);
   background-color: rgba(255, 255, 255, 0.7);
 }
-
-.loading-background > .loading-wrap {
-  position: static;
-}
-
-.error-block {
-  color: #a94442;
-}
--- a/src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
+++ b/src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
@@ -95,11 +95,8 @@
         <div class="panel-body">
           <h4 class="panel-title  border-bottom-small  visible">Billing Address</h4>
           <display-address address="$ctrl.donorDetails.mailingAddress"></display-address>
-          <div class="loading-wrap" ng-if="!$ctrl.donorDetails"><!-- TODO: this seems to be misaligned -->
-            <span class="loading loading1"></span>
-            <span class="loading loading2"></span>
-            <span class="loading loading3"></span>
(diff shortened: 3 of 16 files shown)


adammeyer:
I like it.  Especially for certain pages where this will be used multiple times.
@Omicron7 FYI

Omicron7:
I've got something similar in the PR I'm working on. The one I've got also adds an overlay to block UI elements and grey them out. Used more for overlaying inputs fields after clicking submit and waiting for a response.

OzzieOrca:
Ok cool. I just wanted to consolidate the existing code and add loading indicators to the review checkout step when the panels are empty and loading. Does yours do an overlay on/in a certain element or the whole page. We can use yours if it is general enough.

Omicron7:
Wherever it's included, it overlays the parent element. It has the one requirement of the parent having position=relative style.

src/common/components/loadingOverlay/loadingOverlay.scss
```diff
@@ -11,11 +11,3 @@
   height: calc(100% + 20px);
   background-color: rgba(255, 255, 255, 0.7);
 }
-
-.loading-background > .loading-wrap {
-  position: static;
-}
-
-.error-block {
```
OzzieOrca: @Omicron7 you said this should be removed correct?

src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
```diff
@@ -95,11 +95,8 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Credit Card Paym
         <div class="panel-body">
           <h4 class="panel-title  border-bottom-small  visible">Billing Address</h4>
           <display-address address="$ctrl.donorDetails.mailingAddress"></display-address>
-          <div class="loading-wrap" ng-if="!$ctrl.donorDetails"><!-- TODO: this seems to be misaligned -->
-            <span class="loading loading1"></span>
-            <span class="loading loading2"></span>
-            <span class="loading loading3"></span>
-            <span class="loading loading4"></span>
+          <div style="position: relative">
```
OzzieOrca: @riceguitar I added a few of these to get the loading indicators to be aligned right. You have any suggestions for making it cleaner?

src/common/components/loading/loading.scss
```diff
@@ -0,0 +1,3 @@
+.loading-wrap {
+  transform: translate(-50%, -50%);
```
OzzieOrca: @Omicron7 This should fix the loading indicators not being aligned on their centers. @riceguitar suggested it. I removed your `position: static;` in the loadingOverlay.scss. I'm not sure how much of the alignment stuff in your `loading-background` is still needed.

src/common/components/loadingOverlay/loadingOverlay.scss
Omicron7: I think .error-block can be removed. It was just for testing before we had actual styles form errors.

PR #33, opened 2016-08-16
Registered user sign in

Omicron7:
Adds casApi and sessionService for managing login and session information.
Adds signInForm component which will add a signIn form to other components.
Adds sign-in.html for the registered user / guest checkout during the checkout process.

@OzzieOrca Could you look at the single spec I added (casApi.spec). It copies the cortexApi spec, but sin't passing, and I don't know why.


OzzieOrca:
To fix your spec, you need to add `import appConfig from 'common/app.config';` and its module to the cas api service. The angular environment isn't configured correctly without it and throws this error locally: `TypeError: Cannot read property 'development' of undefined`.

src/app/cart/cart.component.js
```diff
@@ -16,9 +18,10 @@ let componentName = 'cart';
 class CartController{
 
   /* @ngInject */
-  constructor($uibModal, cartService) {
+  constructor($uibModal, cartService, sessionService) {
     this.$uibModal = $uibModal;
     this.cartService = cartService;
+    this.session = sessionService.current;
```
OzzieOrca: I'm assuming this will be used later...

src/app/signIn/signIn.component.js
```diff
@@ -0,0 +1,32 @@
...
+
+import template from './signIn.tpl';
+
+let componentName = 'signIn';
+
+class SignInController {
+
+  /* @ngInject */
+  constructor( $log, $location ) {
+    this.$log = $log;
```
OzzieOrca: Currently unused but logging errors here would be nice. It can stay if that's the intention eventually.

src/common/services/cortexApi.service.js
```diff
@@ -66,8 +65,8 @@ class CortexApi {
     return this.http(request);
   }
 
-  serializePath(path){
-    if(isArray(path)){
+  static serializePath( path){
```
OzzieOrca: I like the idea of this being static (I think I've seen intellij suggest it) but I'm not sure how to test it as angular doesn't publish the class itself when using the injector, just the instance. Do you have any thoughts?

src/common/services/casApi.service.js
```diff
@@ -0,0 +1,69 @@
+import angular from 'angular';
+import 'angular-environment';
+import {Observable} from 'rxjs/Observable';
+import 'rxjs/add/observable/from';
+import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/do';
+
+let serviceName = 'casApiService';
+
+class CasApi {
```
OzzieOrca: How much is this going to be used? If it is for a couple of requests, do you think it would be better to do them directly in the session service? There might be many reasons we are going to CAS though. It just seems like a lot of duplicate code from the cortex api service to just change the url prefix.

src/common/services/casApi.service.js
```diff
@@ -0,0 +1,69 @@
...
+  /*@ngInject*/
+  constructor( $http, envService ) {
+    this.$http = $http;
+    this.envService = envService;
+  }
+
+  http( config ) {
+    config.params = config.params || {};
+    if ( config.followLocation ) {
+      config.params.followLocation = true;
```
OzzieOrca: This was cortex specific. Does CAS have that feature?

OzzieOrca:
Looks good :) I made a few comments

src/app/cart/cart.component.js
Omicron7: It's used in the template current to switch an anchor tag from checkout to sign-in depedning on current session.role.

src/app/signIn/signIn.component.js
Omicron7: I'll pull it, it was mainly for debugging.

src/common/services/casApi.service.js
Omicron7: I'm not sure what all the possible CAS endpoints are yet, or if they might issue redirects. I left it in.

src/common/services/cortexApi.service.js
Omicron7: I yanked it back out to fix the specs. I originally did it because of the intellij suggestion.

src/common/services/casApi.service.js
Omicron7: I don't know all the cas endpoints yet, I do know there will be sign-in, sign-out, forgot password and possibly the security questions. But nothing other than sessionService will use it, so it makes sense to just do $http in the service.

src/common/services/casApi.service.js
OzzieOrca: I meant this is a cortex api feature https://touchpoint-developers.elasticpath.com/1.16.0/API-Reference/Follow-Location (might need to sign in to view).
> The FollowLocation query parameter reduces the number of requests required when performing POST operations. FollowLocation instructs Cortex to retrieve the resource referenced in the Location header and return it in the response body which eliminates the need to make another call to retrieve the resource. FollowLocation can also be used with zoom to further reduce the number of HTTP requests required.

I don't think it's a generic http thing... I couldn't find it quickly on Google anyways.

Omicron7: I'll yank it out when I remove casApi.service in next PR. You can at least login now.

adammeyer:
👍  Looks great!  I really like the way you did the loading indicators.

PR #34, opened 2016-08-16
Add loading indicators to review page

OzzieOrca:
- Disable submit button until all data is defined and has been shown.
- Move the example search requests to the designations service (unrelated but was cluttering the checkout component where it will never be used)

I'll replace the loading HTML with a loading component later when Brian's is merged.


adammeyer:
👍

PR #38, opened 2016-08-17
Display credit card payment info on review page

OzzieOrca:
- Also remove sending the bank account display-account-number as the server generates that

Cortex seems to be returning the encrypted card number in the `card-number` field instead of the masked one. I think they are working on fixing that.


adammeyer:
👍

OzzieOrca:
I fixed my tests

Omicron7:
👍

PR #39, opened 2016-08-17
Move css processing outside of systemjs 

OzzieOrca:
- Create dist/give.css and dist/give.min.css
- Use gulp-preprocess to replace dev imports in index.html with production ones
- Remove systems-plugin-css and systems-plugin-scss
- Output css to dist/ instead of src/
- Remove cache-bust task
- Remove css imports in js
- Remove src/\*\*/\*.css and src/\*\*/\*.css.map from .gitignore (@adammeyer @Omicron7 means you guys need to remove your css files after you pull this commit)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@ jspm_packages
 .sass-cache
 
 src/**/*.tpl.js
-src/**/*.css
-src/**/*.css.map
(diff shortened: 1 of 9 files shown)


adammeyer:
@OzzieOrca Would this be a good time to use the cru.min and only import give css?

adammeyer:
nm, that's not complete yet.  👍

OzzieOrca:
Ya https://github.com/CruGlobal/give-web/blob/a1f9fc10ffce20fb1efc608fa900552e299b8b87/src/app/main/main.scss#L1 is still pulling from Give_Front-end which includes the cru styles... We need to move that in at some point and edit it.

PR #40, opened 2016-08-17
include cartService in product config component

adammeyer:


OzzieOrca:
Is this for future use? I don't see it used there. Also you need to delete those .css and .css.map files. It's my fault there there from the old css tooling.

adammeyer:
It's used in the modal.  So it has to be in the component too.

OzzieOrca:
Oh gotcha. Since they have to be added to the module that makes sense 👍 
You could probably squash these commits.

PR #41, opened 2016-08-17
Change how session.role is handled to allow checking expiry time.

Omicron7:
Update cart to use $location instead of ngSwitch.


OzzieOrca:
👍

PR #42, opened 2016-08-18
Implement storing CCV in sessionStorage and submitting the purchase form on step 3

OzzieOrca:
- Store encrypted CCV into session storage after step 2 credit card payment is submitted
- Add methods to order service to store, retrieve, and clear CCV to/from sessionStorage
- Add getPurchaseForms, submit, and submitWithCcv methods to order service
- Add check for no current payments when loading them in step 3
- In step 3, write submitOrder function to choose and submit the purchase form depending on current payment type
- Fix ng-disabled on Submit your Gift button to check if at least one payment is loaded
- gulp test only needs to run html
- Fix wrong input name in bank account spec

#### Notes:
- Submitting the bank account form will most likely fail as it check for a billing address. Sending a billing address to cortex hasn't been implemented yet. (The `enhancedpurchaseaction` for credit cards doesn't seem to have that kind of validation which is probably a bug)
- From a previous PR: the credit card number on the review page is showing the encrypted version. The api is sending the wrong data back in that field. The backend guys should know about it.

### DIFF
--- a/src/app/checkout/step-2/credit-card/credit-card.spec.js
+++ b/src/app/checkout/step-2/credit-card/credit-card.spec.js
@@ -1,23 +1,277 @@
 import angular from 'angular';
 import 'angular-mocks';
+import omit from 'lodash/omit';
+import size from 'lodash/size';
 import module from './credit-card.component';
 
-describe('checkout', function() {
-  describe('step 2', function() {
-    describe('credit card', function() {
+import cartResponse from 'common/services/api/fixtures/cortex-cart-paymentmethodinfo-forms.fixture.js';
+
+describe('checkout', () => {
+  describe('step 2', () => {
+    describe('credit card', () => {
       beforeEach(angular.mock.module(module.name));
       var self = {};
 
-      beforeEach(inject(function($rootScope, $componentController) {
-        var $scope = $rootScope.$new();
+      beforeEach(inject(($rootScope, $componentController, $httpBackend, $compile) => {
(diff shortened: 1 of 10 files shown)


src/app/checkout/step-2/credit-card/credit-card.spec.js
```diff
@@ -1,23 +1,278 @@
...
+          let expectedPostData ={
+            'card-number': '***encrypted***',
+            'card-type': 'VISA',
+            'cardholder-name': 'Person Name',
+            'expiry-month': 12,
+            'expiry-year': 19
+          };
+          function dataValidator(data){
+            data = angular.fromJson(data);
+            return isEqual(omit(data, 'card-number'), omit(expectedPostData, 'card-number')) && data['card-number'].length >= 100;
```
adammeyer: Would angular.equals work here?

adammeyer:
Looks great! 👍

src/app/checkout/step-2/credit-card/credit-card.spec.js
OzzieOrca: Probably. You like the angular functions better? I guess it's one less import.

adammeyer: I don't mind, I was just thinking less imports the better.

OzzieOrca: I swapped those out

PR #44, opened 2016-08-18
Session expiry

Omicron7:
Updates sessionService.getRole() to check current 'give-session' cookie before returning the 'cortex-session' role. This accounts for users who are idle more than give-session expiration.

Fixes $location.url() issues by using $window.location.href instead.


adammeyer:
👍

PR #45, opened 2016-08-18
Refactor country and region api requests into a geographies service

OzzieOrca:
@adammeyer let me know what you think since you originally wrote this. I made it so you just pass the region function a country and it figures out the link internally.


OzzieOrca:
Also should Canada be using the international address format? I noticed it has regions. Idk how different addresses are there. That doesn't affect this PR either way.

adammeyer:
:+1 Looks good.  Just to confirm, fixtures are only used for testing, correct?

adammeyer:
I originally had it showing the regions dropdown for any country with regions, but that got shot down by QA.  So now it's just US.

OzzieOrca:
Ya they're imported in the `.spec` files only. Interesting... ok

PR #47, opened 2016-08-19
Use enhancedpurchaseform for all checkout flows

OzzieOrca:
Still need to evaluate whether the app should to check for `needinfo` errors. Some of the time I wasn't seeing `enhancedpurchaseform` return those errors like `purchaseform` does. I guess I'll do that in another PR if I need to.


adammeyer:
👍

PR #49, opened 2016-08-19
More designation info

adammeyer:


OzzieOrca:
👍 I think you could squash these commits when merging

PR #50, opened 2016-08-22
Donation config corrections

adammeyer:
This are corrections from QA.

1) In the AMOUNT section – the last option is for "other amount" – for some reason it is getting pre-populated with whichever amount is being selected (for example "50"). If you try to change it, weird up/down arrows show up and make it feel like you have to click one at a time to increase the amount. See how United Way handles their amounts: https://donate.unitedway.org/get-involved/ways-to-give/donate – that's the standard behavior we want.
2) GIFT FREQUENCY - change "One Time" to "Single Gift"
3) OPTIONAL COMMENTS - drop the word "Comments" from the title
4) "Special Handling instructions for Processing the Gift" should read "Special Handling Instructions for Processing This Gift"

### DIFF
--- a/src/app/productConfig/productConfigModal.tpl.html
+++ b/src/app/productConfig/productConfigModal.tpl.html
@@ -48,8 +48,7 @@
                 </div>
               </label>
               <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error" ng-if="($ctrl.itemConfigForm.amount | showErrors)">
-                <div class="help-block" ng-message="required" translate>Amount must be at least $1</div>
-                <div class="help-block" ng-message="pattern" translate>Amount must be at least $1</div>
+                <div class="help-block" ng-message="pattern" translate>Amount must be a valid amount and at least $1</div>
               </div>
             </div>
           </div><!-- // panel-body -->
--- a/src/app/productConfig/productConfigModal.tpl.html
+++ b/src/app/productConfig/productConfigModal.tpl.html
@@ -47,7 +47,7 @@
                 </div>
               </label>
               <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error" ng-if="($ctrl.itemConfigForm.amount | showErrors)">
-                <div class="help-block" ng-message="required" translate>You must enter an amount</div>
+                <div class="help-block" ng-message="required" translate>Amount must be at least $1</div>
                 <div class="help-block" ng-message="min" translate>Amount must be at least $1</div>
               </div>
             </div>
--- a/src/app/productConfig/productConfigModal.tpl.html
+++ b/src/app/productConfig/productConfigModal.tpl.html
@@ -61,8 +61,8 @@
             <div data-toggle="buttons">
               <label class="btn btn-radio btn-wide"
                      ng-repeat-start="f in $ctrl.productData.frequencies | orderBy: $ctrl.frequencyOrder"
-                     ng-class="{'active': !f.selectAction}"
-                     ng-click="$ctrl.changeFrequency(f.selectAction)">
+                     ng-class="{'active': f.name === $ctrl.productData.frequency}"
+                     ng-click="$ctrl.changeFrequency(f)">
                 <span class="giftsum-about giftsum-title" ng-switch="f.name">
                   <span ng-switch-when="NA" translate>Single Gift</span>
                   <span ng-switch-default>{{f.display}}</span>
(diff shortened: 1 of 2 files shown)


src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -35,20 +35,21 @@ <h4 class="panel-title border-bottom-small" translate>
...
-              <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 250" ng-class="{'active': $ctrl.itemConfig.amount === 250}"><span class="number">$250</span></label>
-              <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 500" ng-class="{'active': $ctrl.itemConfig.amount === 500}"><span class="number">$500</span></label>
-              <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 1000" ng-class="{'active': $ctrl.itemConfig.amount === 1000}"><span class="number">$1000</span></label>
-              <label class="btn btn-radio" ng-click="$ctrl.itemConfig.amount = 5000" ng-class="{'active': $ctrl.itemConfig.amount === 5000}"><span class="number">$5000</span></label>
+              <label class="btn btn-radio" ng-click="$ctrl.changeAmount(a)" ng-class="{'active': $ctrl.itemConfig.amount === a}" ng-repeat-start="a in $ctrl.selectableAmounts">
+                <span class="number">${{a}}</span>
+              </label>
+              <span ng-repeat-end></span>
```
OzzieOrca: How come normal `ng-repeat` doesn't work here? Is that `span` required between each label?

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -35,20 +35,21 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                  <input class="form-control form-control-subtle number" type="text" name="amount"
+                         ng-model="$ctrl.customAmount"
+                         ng-change="$ctrl.itemConfig.amount = $ctrl.customAmount"
+                         ng-pattern="/^[1-9]\d*(\.\d+)?$/">
                 </div>
               </label>
               <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error" ng-if="($ctrl.itemConfigForm.amount | showErrors)">
                 <div class="help-block" ng-message="required" translate>Amount must be at least $1</div>
-                <div class="help-block" ng-message="min" translate>Amount must be at least $1</div>
+                <div class="help-block" ng-message="pattern" translate>Amount must be at least $1</div>
```
OzzieOrca: This error could be misleading if someone types `1.` Maybe adding something about it being a valid number would be good.

src/app/productConfig/productConfigModal.tpl.html
adammeyer: A space is, but normal ng-repeat puts the labels too close together.

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -63,7 +64,10 @@ <h4 class="panel-title border-bottom-small mt" translate>
                      ng-repeat-start="f in $ctrl.productData.frequencies | orderBy: $ctrl.frequencyOrder"
                      ng-class="{'active': !f.selectAction}"
                      ng-click="$ctrl.changeFrequency(f.selectAction)">
-                <span class="giftsum-about giftsum-title">{{f.display}}</span>
+                <span class="giftsum-about giftsum-title" ng-switch="f.name">
+                  <span ng-switch-when="NA" translate>Single Gift</span>
+                  <span ng-switch-default>{{f.display}}</span>
+                </span>
               </label>
               <span ng-repeat-end></span>
```
OzzieOrca: Using `ng-repeat-start` and `ng-repeat-end` is here too...

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -126,7 +130,7 @@ <h4 class="panel-title border-bottom-small mt" translate>
                         style="display:block;"></textarea>
               </li>
               <li>
-                <a class="commentLink" href="" ng-click="showDSComments = !showDSComments" translate>Special Handling Instructions for Processing the Gift</a>
+                <a class="commentLink" href="" ng-click="showDSComments = !showDSComments" translate>Special Handling Instructions for Processing This Gift</a>
```
OzzieOrca: Does the `This` need to be capitalized?

adammeyer: I copied it from the task.  I'll check.

src/app/productConfig/productConfigModal.tpl.html
OzzieOrca: Oh interesting. I guess ideally that would be done in CSS

OzzieOrca:
Besides comments, 👍

src/app/productConfig/productConfigModal.tpl.html
adammeyer: idk.  What they have is valid html.  ng-repeat just removes the spacing between each element.  I think this is the best way to resolve that.

PR #51, opened 2016-08-22
Session Modal Service

Omicron7:
Initial implementation of a sessionModalService, signIn Modal and sign in button component.
Includes non-functional components for signUp and resetPassword.

@OzzieOrca @adammeyer I'd like some initial feedback on these while I work on implementing the signUp functionality.


src/common/components/resetPasswordModal/resetPasswordModal.tpl.html
```diff
@@ -0,0 +1,12 @@
+<div class="col-md-12 col-xs-12">
+  <p>Enter your registered email address below and we'll send a new password to that email.</p>
+  <form>
+    <div class="form-group">
+      <label for="email">Email</label>
+      <input type="email" class="form-control form-control-subtle" placeholder="Email">
+    </div>
+    <button type="submit" class="btn btn-block btn-primary">Reset My Password</button>
+  </form>
+  <p class="mt- text-center">Never mind
```
OzzieOrca: I feel like this needs a period or something after to separate it from the link. They are 2 different phrases.

src/common/components/signInButton/signInButton.component.js
```diff
@@ -0,0 +1,66 @@
...
+    'gettext',
+    template.name,
+    sessionService.name,
+    sessionModalService.name
+  ] )
+  .component( componentName, {
+    controller:  SignInButtonController,
+    templateUrl: template.name,
+    bindings:    {
+      pageReload: '<'
```
OzzieOrca: Is this going to be coming in from a variable in some outer scope or just hardcoded as a string on each btn instance such as `<sign-in-button page-reload="false">`? If it's the second case, I think using the `@` binding might be better.

src/common/components/resetPasswordModal/resetPasswordModal.tpl.html
Omicron7: Thats copied directly from http://crugive.site/

src/common/components/signInButton/signInButton.component.js
Omicron7: It's the later case, but '@' binding doesn't work with boolean values, changes them to strings.
The page-reload binding was something I added in the hope that other components become more session aware and don't need a complete page reload. We may never use this field, don't know yet.

src/common/components/resetPasswordModal/resetPasswordModal.tpl.html
OzzieOrca: Gotcha but I still think that. Idk if my opinion has any weight.

OzzieOrca:
I think this looks good. I like the component reuse and interaction and using the BehaviorSubject to broadcast changes in state to all the buttons.

src/common/components/signInButton/signInButton.component.js
OzzieOrca: Alright ya. Whatever you think is best

adammeyer:
👍

PR #52, opened 2016-08-23
Implement billing address functionality

OzzieOrca:
- Cleanup bank account spec to use spies and fake methods instead of calling real service methods that use $http
- Load geographies into credit card component
- Allow user to enter and save US and international billing address info in credit card component
- Load donor details in credit card component to display mailing address
- Copy mailing address to billing address if "Same as Mailing Address is checked" in credit card component
- Load billing address in step 3 component
- Create display-address component to format either a US address or an international one
- Move getDonorDetails, updateDonorDetails and addEmail functions from cart service to order service
- Add billing address save and load methods to order service
- Add address formatting functions to order service to handle packing and unpacking international addresses all into "street-address"
- Stop sending credit card type as it seems to be computed on the server


OzzieOrca:
I'm not sure if you guys want to code review this now and get it merged as is or wait until it is fully done. This still has a couple issues that I'm not sure how to resolve. I'm hoping to get some better answers from my messages on slack and hipchat:

1. Currently the billing address assigned to the order and displayed throughout the checkout process is the user's default address (or the first one ever entered). I still need to use the cortex selector to choose the address for the order. If you guys have implemented something using a selector before and have any tips/docs that would be  great. Otherwise I'll try to figure it out.
2. Cortex currently won't let the app save an international billing address. It requires all fields to be filled out but since this is currently implemented following how the mailing address works by putting all lines in the `street-address` field, it returns an error response saying they cannot be blank.
3. Cortex addresses seem to require that a first and last name be provided with them. There is no UI element for this. Copying the name from donor details won't work cuz the donor could be paying with someone else's credit card. We could either split the cardholder name into 2 inputs or try to figure out how to split the cardholder name string into a first and last name.

src/app/checkout/step-2/credit-card/credit-card.spec.js
```diff
@@ -1,25 +1,33 @@
...
 
-      beforeEach(inject(($rootScope, $componentController, $httpBackend, $compile) => {
+      beforeEach(inject(($rootScope, $httpBackend, $compile) => {
         self.outerScope = $rootScope.$new();
         self.$httpBackend = $httpBackend;
 
         self.outerScope.submitted = false;
         self.outerScope.onSave = () => {};
         let compiledElement = $compile(angular.element('<checkout-credit-card submitted="submitted" on-save="onSave(success)"></checkout-credit-card>'))(self.outerScope);
+        // TODO: figure out how to skip calling $onInit when using $compile. For now, just expect the requests generated by calling $onInit. Adding a spy to $onInit would be best but I couldn't figure out how to access the controller before $onInit is called
```
Omicron7: Can you mock $onInit to a noop for purposes of testing?

Omicron7:
Looks good 👍 

Comments on above questions:
#1 It may have been decided in the Monday Sprint planning that tracking multiple addresses isn't required and changes to the address always update your default. I'd check with Jason/Zandre

Regarding #3 above. Can Donor's pay with someone else's payment information? For tax purposes, I'd assume only the entity/person who donated would get a tax deduction. Meaning that donors could only use their funds for donation. I'd default to Donor first/last.

src/app/checkout/step-2/credit-card/credit-card.spec.js
OzzieOrca: I can't seem to get a hold of it before it is run. It is nonexistent before the scope apply and then has already been run after

OzzieOrca:
K thanks.

For #3, couldn't they use their spouses card or something? Idk that there will always be an exact match. I can suggest that when talking to Jason and the other UX guys.

adammeyer:
@OzzieOrca Regarding #2.  Post a comment in JIRA and tag Jason and Randy in it.  See what their input is.

#3, What about putting what the type in for "Name on Card" as the billing name?

OzzieOrca:
@adammeyer for #3 that probably would work but that one input would either need to be split in half on a space somewhere or have 2 inputs cuz cortex is expecting the fname and lname separately. Someone said the ERT got around it by splitting the string (I guess authorize.net does the same thing) but that wasn't ideal cuz you can't always tell where the fname stops and the lname begins

adammeyer:
@OzzieOrca That makes sense.  I would post that in JIRA as well, tag the same people.  Ask how that issue should be addressed.

OzzieOrca:
I'll resolve these issues in another PR. They are currently blocked but might have a solution in using the profile credit card payment form or modifying it to use in the order endpoint. I'm still waiting to hear back from Dave and there might be some backend work that has to be done.

PR #55, opened 2016-08-23
Remove semi annual

adammeyer:


src/app/productConfig/productConfig.modal.js
```diff
@@ -19,7 +19,7 @@ class ModalInstanceCtrl{
   }
 
   frequencyOrder(f){
-    let order = ['NA', 'MON', 'QUARTERLY', 'ANNUAL', 'SEMIANNUAL'];
+    let order = ['NA', 'MON', 'QUARTERLY', 'SEMIANNUAL', 'ANNUAL'];
```
OzzieOrca: Should `SEMIANNUAL` still be here? I just don't know the details of what you're doing.

adammeyer: I'm not sure.  I thought I should leave it in just in case.  We may use something like this for gifts already setup as semiannual.  Just new gifts can't be setup this way.

OzzieOrca: Gotcha. Is this modal going to be used for editing gifts?

adammeyer: Probably not.

OzzieOrca: Do what you think is best. It just seemed a little weird sorting by it and then never showing it.

adammeyer: Okay, removed.

PR #57, opened 2016-08-23
Check needinfo for errors before submitting order

OzzieOrca:
I added basic error messages but they aren't styled very well and I'm not sure the best place to display them. I've messaged Sudarma to get his thoughts for showing generic error messages. I'm not sure who else I should talk to about UX.

### DIFF
--- a/src/common/services/api/order.service.js
+++ b/src/common/services/api/order.service.js
@@ -189,10 +189,11 @@ class Order{
       }
     })
       .map((data) => {
-        return map(data.needInfo, 'name');
+        let errors = map(data.needInfo, 'name');
+        return (errors && errors.length > 0) ? errors : undefined;
       })
       .do((errors) => {
-        this.$log.error('The user was presented with these `needinfo` errors. They should have been caught earlier in the checkout process.', errors);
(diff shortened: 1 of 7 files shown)


adammeyer:
@OzzieOrca David is the right guy.

src/common/services/api/order.service.js
```diff
@@ -179,6 +181,21 @@ class Order{
...
+      path: ['carts', this.cortexApiService.scope, 'default'],
+      zoom: {
+        needInfo: 'order:needinfo'
+      }
+    })
+      .map((data) => {
+        return map(data.needInfo, 'name');
+      })
+      .do((errors) => {
+        this.$log.error('The user was presented with these `needinfo` errors. They should have been caught earlier in the checkout process.', errors);
```
adammeyer: Is it standard practice to log errors in production?

OzzieOrca: My thought was to pipe `$log` to rollbar eventually but I've never actually done that. Do you guys have any thoughts?

adammeyer: Good idea.  I like it.

adammeyer:
👍

OzzieOrca:
I've updated this and moved the error messages to the top.

Omicron7:
:thumbsup:

PR #58, opened 2016-08-24
Search

adammeyer:
This is the initial search results page.  No UI yet, so I just made a temp one.  I'll revisit and add better pagination once the UI is available.

The real search form will be on the home page.  But for testing, I have made one in the "Debug Tools" section.


OzzieOrca:
👍, not sure why `epSelector.js` is called that. That name confused me and makes me think of using the cortex api `selector` to choose one address out of a list of addresses or something.

adammeyer:
how about urlTermSelector?

OzzieOrca:
I guess this could be used more generically than I thought since it could probably grab any AEM "query param" (idk what to call those url segments). Maybe use whatever those are called in AEM. I'm not sure I like the `Selector` part cuz it's not really choosing anything out of a list. Maybe just `urlTerms` or something cuz I think calling `urlTerms.get()` makes sense when you read that line.

I've gotten too picky... Sorry

PR #59, opened 2016-08-24
Sign Up Modal

Omicron7:
Implements Sign Up Modal.

### DIFF
--- a/src/common/components/signUpModal/signUpModal.tpl.html
+++ b/src/common/components/signUpModal/signUpModal.tpl.html
@@ -20,7 +20,7 @@
         </div>
       </div>
       <div class="col-xs-6">
-        <div class="form-group" ng-class="{'has-success': $ctrl.signUpForm.last_name.$valid, 'has-error': ($ctrl.signUpForm.last_name.$error | showErrors) }">
+        <div class="form-group" ng-class="{'has-success': $ctrl.signUpForm.last_name.$valid, 'has-error': ($ctrl.signUpForm.last_name | showErrors) }">
           <label for="last_name" translate>Last Name</label>
           <input id="last_name"
                  name="last_name"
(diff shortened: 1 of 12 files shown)


src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -1,53 +1,125 @@
...
+             name="password"
+             type="password"
+             class="form-control form-control-subtle"
+             ng-model="$ctrl.password"
+             minlength="8"
+             ng-pattern="/^(?=.*\d)(?=.*[a-zA-Z]).{6,}$/"
+             required>
+      <div ng-if="($ctrl.signUpForm.password | showErrors)" ng-messages="$ctrl.signUpForm.password.$error">
+        <span class="help-block" ng-message="required" translate>Password is required.</span>
+        <span class="help-block" ng-message="pattern" translate>Must include 1 letter and 1 number.</span>
```
OzzieOrca: I think there should be a `minlength` message here also

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -1,53 +1,126 @@
 <div class="col-md-12 col-xs-12">
-  <form>
+  <div class="alert alert-danger" role="alert" ng-messages="$ctrl.signUpErrors" ng-if="$ctrl.hasError">
+    <div ng-message="400" translate>An error occurred during sign up.<br />Please Check your Email Address and Password and try again.</div>
+    <div ng-message="403" translate>An error occurred during sign up.<br />This is most likely caused by the use of a reserved domain name in your email address. You might try to
```
OzzieOrca: What does a `reserved domain name in your email address` mean? I feel like if I'm not sure what it means, users will be confused by the error...

OzzieOrca:
👍

src/common/components/signUpModal/signUpModal.tpl.html
Omicron7: 403 FORBIDDEN - Email domain is reserved (cru.org, ccci.org, uscm.org...). There is an on boarding process for corporate email domains which can't be bypassed through give site.

These error messages are included in https://jira.cru.org/browse/EP-726 for Jason Chandler to modify.

Omicron7: I contemplated making it the same as 409, since if you are signing up with a corporate email, chances are you already have an account.

PR #62, opened 2016-08-24
Add Rollbar

OzzieOrca:
This adds the generic rollbar error handling but also pipes Angular's `$log` to rollbar with stack traces.

I made a `give-web devs` team on Rollbar but I haven't added you guys yet. Feel free to add yourselves or I can. I just didn't want you guys getting random errors as I test it.

### DIFF
--- a/src/common/rollbar.config.js
+++ b/src/common/rollbar.config.js
@@ -1,24 +1,54 @@
 import { rollbarAccessToken } from 'common/app.constants';
 import rollbar from 'rollbar-browser';
+import stacktrace from 'stacktrace-js';
 
 /* @ngInject */
-function rollbarConfig(envServiceProvider, $provide) {
+function rollbarConfig(envServiceProvider, $provide, $windowProvider) {
+  if($windowProvider.$get().__karma__ !== undefined) return; // Don't decorate $log in test env
+
   let rollbarConfig = {
     accessToken: rollbarAccessToken,
     captureUncaught: true,
     captureUnhandledRejections: true,
-    environment: envServiceProvider.get()
+    environment: envServiceProvider.get(),
+    enabled: !envServiceProvider.is('development'), // Disable rollbar in development environment
+    verbose: envServiceProvider.is('development') // Log rollbar errors to console in development environment
   };
   let Rollbar = rollbar.init(rollbarConfig);
 
   /* @ngInject */
   $provide.decorator('$log', ($delegate) => {
+    // Add rollbar functionality to each $log method
     angular.forEach(['log', 'debug', 'info', 'warn', 'error'], (ngLogLevel) => {
-      let originalFunction = $delegate[ngLogLevel];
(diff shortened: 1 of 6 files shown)


src/common/rollbar.config.js
```diff
@@ -0,0 +1,60 @@
...
+  // eslint-disable-next-line angular/window-service
+  if(window.__karma__ !== undefined) return; // Don't decorate $log in test env
+
+  let rollbarConfig = {
+    accessToken: rollbarAccessToken,
+    captureUncaught: true,
+    captureUnhandledRejections: true,
+    environment: envServiceProvider.get(),
+    enabled: !envServiceProvider.is('development'), // Disable rollbar in development environment
+    verbose: envServiceProvider.is('development') // Log rollbar errors to console in development environment
```
OzzieOrca: This may be unwanted and just clutter up our development consoles. It would help us see which errors are getting sent to rollbar though. What do you guys think/prefer?

Omicron7:
👍 Looks good to me. I'm fine with development environments outputting stack traces locally.

adammeyer:
👍 
There are many json paths errors in the console now when using the minified version of the app.  Those should probably be resolved before this is merged.

OzzieOrca:
@adammeyer I'm not sure how to resolve the JSONPath errors. I feel like the library must have been made for Node and browsers were an afterthought.

https://github.com/dchester/jsonpath/blob/master/lib/grammar.js#L104-L105 uses `fs` (which seems to do network requests in the browser) to load [`module.js`](https://github.com/dchester/jsonpath/blob/master/include/module.js) and [`action.js`](https://github.com/dchester/jsonpath/blob/master/include/action.js). I have no idea why it needs those files as it seems to work fine without them. Without hosting those 2 files on AEM or forking JSONPath and editing it, idk how to get rid of those errors. Maybe there's a way to add stuff to a cache `fs` has but I haven't seen it.

adammeyer:
@OzzieOrca Okay, that's fine for now.  Just didn't want it to start populating rollbar with errors we already knew existed.

src/common/rollbar.config.js
```diff
@@ -0,0 +1,99 @@
...
+
+  /* @ngInject */
+  $provide.decorator('$log', ($delegate) => {
+    // Add rollbar functionality to each $log method
+    angular.forEach(['log', 'debug', 'info', 'warn', 'error'], (ngLogLevel) => {
+      let rollbarLogLevel = ngLogLevel === 'warn' ? 'warning' : ngLogLevel;
+
+      let originalFunction = $delegate[ngLogLevel]; // Call below to keep angular $log functionality
+
+      $delegate[ngLogLevel] = (...args) => {
```
adammeyer: @OzzieOrca what does this "...args" do?

OzzieOrca: Gets the args as a real array. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters

adammeyer: Okay, cool!

PR #64, opened 2016-08-25
Fixes issue with signIn Modal closing on invalid user/pass

Omicron7:
We shouldn't do anything onFailure (close the modal). Let the signInForm handle failure state. If the user wants to close the modal, then can do it manually.


adammeyer:
👍

PR #65, opened 2016-08-25
Added missing images and updated paths.

Omicron7:


src/app/checkout/step-2/bank-account/bank-account.tpl.html
```diff
@@ -82,7 +82,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Bank Account Pay
             <div class="panel-body">
               <div class="row">
                 <div class="col-sm-8">
-                  <img src="/assets/img/check-sample.png" alt="Credit Card Icon"/>
+                  <img src="/assets/images/check-sample.png" alt="Credit Card Icon"/>
```
adammeyer: Because this will be hosted on AEM, these are going to have to be absolute paths.  I can set a window variable for the domain, then they can be referenced via ng-src.  Thoughts?

OzzieOrca: You can add another variable to angular environment if you want. You'd have to inject it and put it on the template scope which is annoying but I think you'd have to do that with any global angular service/constant unless you put it on window.

adammeyer: I'll put the variable on Window from AEM, but it could then be converted to an angular constant for use in the templates.

OzzieOrca: The variable will just be the s3 urls right? Do you think adding interaction between AEM and angular is better? I was thinking just add hardcoded strings for the buckets...

adammeyer:
added the imgDomain env var.  @OzzieOrca @Omicron7 Please review.

@OzzieOrca FYI: the check sample is never shown because of the bank-explain-wrap class.  Didn't know if you where aware of that.

Omicron7:
👍

PR #66, opened 2016-08-25
Replace json path lib

adammeyer:


adammeyer:
Sorry for all the commits, I'll squash those.

adammeyer:
This will make the common file so much smaller, yay! 😄  🎉  🎈

OzzieOrca:
Yay 😄

OzzieOrca:
I mean other than killing the coverage by 4% looks good. Maybe I should figure out how to ignore that directory...

Omicron7:
👍 
We should figure out how to ignore any file without coverage. Never see a coverage drop again.

PR #67, opened 2016-08-25
Img dir rename

adammeyer:
UX is going to use img from now on and phase out images.


OzzieOrca:
K that's fine. Why do you have so many merge commits?

adammeyer:
I'll squash them

PR #70, opened 2016-08-26
Load CCP keys dynamically

OzzieOrca:
Fetch CCP key from remote URL and fallback to hardcoded key if there is an error.

- Add CCP service which returns an Observable containing CCP when the keys are loaded and it has been initialized
- Add ccpKeyUrl to angular environment


adammeyer:
👍

PR #71, opened 2016-08-26
Update modals to reflect current redesign.

Omicron7:
html and CSS class changes are complete. CSS needs one fix https://jira.cru.org/browse/EP-1182


adammeyer:
👍

PR #73, opened 2016-08-26
Added test coverage to sessionService.

Omicron7:


src/common/services/session/sessionModal.controller.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      close:   angular.noop,
+      dismiss: angular.noop
+    };
+    state = 'sign-in';
+    $ctrl = $controller( module.name, {
+      $uibModalInstance: uibModalInstance,
+      state:             state
+    } );
+    spyOn( uibModalInstance, 'close' );
+    spyOn( uibModalInstance, 'dismiss' );
```
OzzieOrca: This works fine but to not have to use the `angular.noop` above, you could use `createSpy` or `createSpyObj`:

```
uibModalInstance = {
    close:   jasmine.createSpy('close'),
    dismiss: jasmine.createSpy('dismiss')
};
```
or
```
uibModalInstance = jasmine.createSpyObj('uibModalInstance', ['close', 'dismiss']);
```
I think the first parameter in each jasmine function is just used in error reporting...

src/common/services/session/sessionModal.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+  describe( 'open', () => {
+    it( 'should be defined', () => {
+      expect( sessionModalService.open ).toBeDefined();
+    } );
+
+    it( 'should open \'sign-in\' by default', () => {
+      sessionModalService.open();
+      expect( $uibModal.open ).toHaveBeenCalled();
+      expect( $uibModal.open.calls.count() ).toEqual( 1 );
```
OzzieOrca: You could use `toHaveBeenCalledTimes` to combine these 2 lines.

src/common/services/session/sessionModal.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    } );
+
+    it( 'should allow options', () => {
+      sessionModalService.open( 'sign-up', {backdrop: false, keyboard: false} );
+      expect( $uibModal.open ).toHaveBeenCalled();
+      expect( $uibModal.open.calls.count() ).toEqual( 1 );
+      let options = $uibModal.open.calls.argsFor( 0 )[0];
+      expect( options.resolve.state() ).toEqual( 'sign-up' );
+      expect( options.backdrop ).toEqual( false );
+      expect( options.keyboard ).toEqual( false );
```
OzzieOrca: You could do something like this but since there's that `state()` function call, idk that you can  incorporate it in the `toHaveBeenCalledWith` so it might be cleaner to leave it as is. I'll share the code anyways...
```
expect($uibModal.open).toHaveBeenCalledWith(jasmine.objectContaining({
    backdrop: false,
    keyboard: false
}));

**Comment by OzzieOrca** [comment]
👍  Looks great. Your code is totally readable so my comments are optional and just come from the fact that I've been looking through the jasmine docs for the past few weeks. Feel free to take anything you like. Testing can kinda turn into how many convenience functions can you remember which is not always the best thing or the best use of time... I know rspec has tons of functions. Thankfully jasmine doesn't have quite that many :)

**Comment by adammeyer** [comment]
👍

File: src/common/components/signInForm/signInForm.spec.js [hunk 4]
```diff
@@ -0,0 +1,83 @@
+import angular from 'angular';
+import 'angular-mocks';
+import {Observable} from 'rxjs/Observable';
+import 'rxjs/add/observable/of';
```
OzzieOrca: Unused import?

OzzieOrca:
👍

src/common/components/signInForm/signInForm.spec.js
Omicron7: Actually, should be importing 'from'.

Omicron7:
👍

PR #74, opened 2016-08-30
Return order info after submitting purchase

OzzieOrca:
- Remove old info logs
- Log purchase info on submit
- Send empty object on submit without CCV in the body instead of undefined so $http sends a content type
- On submit, followLocation

This should help testers see some data for now and will help my thank you component.


adammeyer:
👍

PR #76, opened 2016-08-30
payment methods inside profile

denlight:
task is blocked for now due to a non-functional end-point


src/app/profile/payment-methods/payment-method/payment-method.tpl.html
```diff
@@ -0,0 +1,58 @@
+  <div class="panel panel-default panel-with-gray-background">
+    <div class="panel-heading" role="tab" id="paymentHeading{{index}}">
+      <div class="payment-summary-link p- collapse-indicator-row {{$ctrl.isCollapsed ? 'collapsed' : ''}}" role="button">
```
adammeyer: use ng-class for the collapsed class

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
```diff
@@ -0,0 +1,58 @@
...
+        <div class="row">
+          <div class="col-md-8">
+            <p>
+              <img class="hidden-xs payment-icon" src={{"/assets/img/cc-icons/"+''}}>
```
adammeyer: There needs to be an image in the src here, looks like that part is missing.

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
```diff
@@ -0,0 +1,58 @@
+  <div class="panel panel-default panel-with-gray-background">
+    <div class="panel-heading" role="tab" id="paymentHeading{{index}}">
```
adammeyer: I don't think the id is necessary here

index.html
```diff
@@ -11,6 +11,7 @@
       <base href="/">
       <!-- @if ENV='dev' -->
       <link rel="stylesheet" href="dist/give.css" type="text/css">
+      <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.css">
```
adammeyer: Is this being used?  I don't see any FA icons in the markup.

denlight: @adammeyer , inside the modal/bank account there is a question mark at the bottom.

denlight: @sorry, not at the bottom. In the middle 'Locate your Routing and Account Numbers ? '

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
```diff
@@ -0,0 +1,87 @@
+<div class="modal-content give-modal-content">
```
adammeyer: I believe model-content class is added by the UI-bootstrap JS.  Double check that there isn't already a div with this class as a parent when the modal is open.

denlight: @adammeyer , you are right. The reason why I did a duplicate 'modal-content' div is because of the styles. Modal form styles are attached to '.modal-content.give-modal-content'. I couldn't find a way to append a class to the 'modal-content' div. Here https://angular-ui.github.io/bootstrap/#/modal I have found a way to add class to a parent modal container or a backdrop, but not 'modal-content'. So I hacked it. It is ugly but I am not sure how else to go around it. Im open for suggestions though =)

adammeyer: Take a look at what was done here: https://github.com/CruGlobal/give-web/blob/9f7548cf86cfd9c7b1303a3f75b7e82f00019b68/src/common/services/session/sessionModal.service.js#L21

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
```diff
@@ -0,0 +1,86 @@
...
+
+          <div class="mb_x">
+            <div class="row">
+              <div class="col-md-6">
+                <a
+                  ng-class="{'on':$ctrl.paymentType == 'bankAccount'}"
+                  ng-click = "$ctrl.switchPaymentType('bankAccount')"
+                  class="tab-btn btn btn-lg btn-default btn-block">
+                  Bank Account
+                  <img src="/assets/img/icon-bank.png" alt="Bank Account Icon">
```
adammeyer: For images, you'll need to add the imgDomain variable so they show correctly on production.

Here is an example: https://github.com/CruGlobal/give-web/blob/master/src/app/checkout/step-2/step-2.tpl.html#L12

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
```diff
@@ -0,0 +1,86 @@
...
+        <!-- credit card form -->
+        <div ng-if="$ctrl.paymentType === 'creditCard'" class="tab-indiv cart-tab show">
+          <div class="row">
+            <div class="col-md-12 col-xs-12">
+              <checkout-credit-card submitted="$ctrl.submitted" on-save="$ctrl.onSave(success,error)"></checkout-credit-card>
+            </div>
+          </div>
+        </div><!-- credit card form -->
+        <!-- loading -->
+        <div class="loading-wrap" ng-if="$ctrl.submitted">
```
adammeyer: Use the loadingOverlay component here, check the search page or home page login for an example.

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
```diff
@@ -0,0 +1,58 @@
+  <div class="panel panel-default panel-with-gray-background">
+    <div class="panel-heading" role="tab">
+      <div ng-class="{'collapsed': $ctrl.isCollapsed}" class="payment-summary-link p- collapse-indicator-row" role="button">
+        <div class="row">
+          <div class="col-md-8">
+            <p>
+              <!-- <img class="hidden-xs payment-icon" src={{"/assets/img/cc-icons/"+''}}> at the moment there is no way for me to know what image I need-->
+              <span class="payment-nickname">Visa</span>
```
adammeyer: This should show the payment type from API

src/app/profile/payment-methods/payment-methods.tpl.html
```diff
@@ -0,0 +1,23 @@
...
+          <div class="row pb">
+            <div class="col-md-6">
+              <h3>Your Payment Methods</h3>
+            </div>
+            <div class="col-md-6 text-right">
+              <a class="btn btn-primary" ng-click="$ctrl.addPaymentMethod()">Add New Payment Method</a>
+            </div>
+          </div>
+          <div class="row">
+            <div ng-repeat="(index,paymentMethod) in $ctrl.paymentMethods" class="panel-group panel-collapse" id="accordion" role="tablist" aria-multiselectable="true">
```
adammeyer: id can be removed here

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
denlight: @adammeyer I used it first but it looked a little wonky. The background (overlay) was too short and wasn't covering the whole modal. So I used a hardcoded version without a background like Scotty did in his credit card component. I was thinking about making a loading component a little more dynamic where we can turn off/on the background. what do you think about that ?

src/app/profile/payment-methods/payment-methods.tpl.html
```diff
@@ -0,0 +1,23 @@
+<div class="screenContent account-management">
+  <div class="container container-edge-small container-edge-extrasmall">
+    <div class="container">
```
adammeyer: This seems weird to have 2 containers.  Is that correct?

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
adammeyer: You may need to set the element above the loadingOverlay to position:relative.  Give that a try and let me know.

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
denlight: Nice! That will do it =)

adammeyer:
Thanks @denlight, looks good.  I'll have @OzzieOrca take a quick look too.  Is it correct that this endpoint isn't working?  And that is why the data is hard coded in?  I thought this one was and the manage reoccurring gifts one was not.  But there are so many, it's hard to keep track of them all! :)

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,34 @@
...
+
+class PaymentMethodController{
+
+  /* @ngInject */
+  constructor(){
+    this.isCollapsed = true;
+  }
+
+  getExpiration() {
+    return `EXPIRES ${this.model.expiry_month}/${this.model.expiry_year}`;
```
OzzieOrca: I think you can do this in the view. Doing translation is easier there. Was there a reason you did it here?

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  getLastFourDigits(){
+    let cardNumber = this.model.card_number;
+    return `ending in ****${cardNumber.substr(cardNumber.length-4,this.model.card_number.length-1)}`;
```
OzzieOrca: I know you might not have seen this in the api yet but the api should only give you the last 4 digits or already have prefixed a bunch of starts to it. I think I've seen it both ways. If it ever gives you a full number, I'd make that known since that's a security issue. Again I think the `ending in` belongs in the view so it can be translated.

src/app/profile/profile.tpl.html
```diff
@@ -0,0 +1,7 @@
+<div class="screenContent">
+  <div class="container container-edge-small container-edge-extrasmall">
+    <div class="container">
```
OzzieOrca: `container` `container` again

src/app/profile/payment-methods/payment-method/forms/add-payment-method/modal.tpl.html
```diff
@@ -0,0 +1,86 @@
...
+    <div class="container-fluid" tabindex="-1" role="dialog">
+      <div class="row">
+        <div class="col-xs-12">
+
+          <div class="alert alert-danger" role="alert" ng-if="$ctrl.error">
+            <p>{{$ctrl.error}}</p>
+          </div>
+
+          <div class="mb_x">
+            <div class="row">
```
OzzieOrca: It might be nice to share some of this html with what the step-2 component has. It might take some refactoring in the checkout component or adding an intermediate component though since there are next step buttons there... I'll think about that more

index.html
OzzieOrca: Ya that question mark has been missing for a while. AEM seems to include font-awesome. I guess this is fine for development.

OzzieOrca:
Looks good. Using the checkoutBankAccount and checkoutCreditCard components as-is won't work as they submit the payment info to the orders endpoint and not the profile endpoint. This is my fault not yours. I guess I need to pull the logic to save the payment info to the api out into the parent component. I could probably pass the data through that `on-save` callback. I'm working on loading the existing payment methods for the checkout component so maybe I'll figure out what's best as part of that.

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
denlight: @adammeyer , I'm not sure what you mean. At the moment there is nothing that comes from API since the end-points are not returning what I add. It is dummy data for now.

src/app/profile/payment-methods/payment-methods.tpl.html
denlight: @adammeyer , sorry, that is a side effect of a copy / paste syndrome =)

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
OzzieOrca: I guess he meant you don't have dummy data for it. Just don't forget about it and leave it hardcoded when you have real api data.

src/app/main/main.component.js
```diff
@@ -40,6 +49,21 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
       url: "/checkout.html",
       template: '<checkout></checkout>'
     })
+    .state('profile', {
+      url: "/profile",
+      abstract: true,
+      template: '<profile></profile>',
+      onEnter: (sessionService,sessionModalService,$state) => {
```
adammeyer: The UI router is only used for the local dev env, it's not used once the app is integrated with the CMS.  So, you'll want to implement this using the session enforcement service @Omicron7 made.

Here is an example on the checkout page: https://github.com/CruGlobal/give-web/blob/master/src/app/checkout/checkout.component.js#L39

src/app/main/main.component.js
```diff
@@ -40,6 +49,21 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
       url: "/checkout.html",
       template: '<checkout></checkout>'
     })
+    .state('profile', {
+      url: "/profile",
+      abstract: true,
```
Omicron7: ui-router is only included in our development environment, so we can't use it to enforce or resolve state changes. It's kludgy, but in production, each component in /app is actually an independent angular application and not a single application like we have in development (This is a design requirement of Adobe Experience Manager and analytics). This also requires that state management is handled inside each base component (parent/child ui-router states won't work in production).

src/app/main/main.component.js
```diff
@@ -40,6 +49,21 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
...
+      template: '<profile></profile>',
+      onEnter: (sessionService,sessionModalService,$state) => {
+        if(sessionService.getRole() != 'REGISTERED') {
```
Omicron7: I've written a sessionEnforcerService to facilitate session enforcement and expiration in components. You can see an example of it used in [checkout](https://github.com/CruGlobal/give-web/blob/master/src/app/checkout/checkout.component.js#L39)

src/app/main/main.component.js
denlight: Thanks for letting me know @Omicron7 , I will look at your examples and take care of my PR

src/common/services/api/profile.service.js
```diff
@@ -29,7 +29,7 @@ class Profile{
     return this.cortexApiService.get({
         path: ['profiles', this.cortexApiService.scope, 'default'],
         zoom: {
-          paymentMethods: 'paymentmethods:element[]'
+          paymentMethods: 'paymentmethods:element[],paymentmethods:element:bankaccount,paymentmethods:element:bankaccount:element,paymentmethods:element:creditcard,paymentmethods:element:creditcard:element'
```
denlight: @OzzieOrca , @adammeyer ,guys, do you think it would be good to make zoom.paymentMethods have dynamic value? In case where I need recurring gifts passed along with payment methods I add more loading time when you don't really need extra data for checkout. Let me know what you guys think. Thanks

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.tpl.html
```diff
@@ -0,0 +1,13 @@
+
```
denlight: @OzzieOrca , @adammeyer , do you guys think it would be good to make bank account/credit card forms handle editing as well? In my example bellow I use the forms and virtually pass some payment method data that could be populated into the form.

src/common/services/api/profile.service.js
OzzieOrca: These aren't used for checkout but I imagine they'll be used in self service for managing gifts and other things. What do you think of making 2 service methods? `getPaymentMethods` and something like `getPaymentMethodsWithDonations`. That query is kinda slow...

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.tpl.html
OzzieOrca: yes I think adding an input binding to those would be valuable

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -56,6 +57,17 @@ class CreditCardController {
       });
   }
 
+  populateData(){
```
denlight: @OzzieOrca , @adammeyer would that work for data population?

PR #77, opened 2016-08-30
home sign in component

adammeyer:
For the AEM home page

### DIFF
--- a/src/app/homeSignIn/homeSignIn.component.js
+++ b/src/app/homeSignIn/homeSignIn.component.js
@@ -1,0 +1,62 @@
+import angular from 'angular';
+import includes from 'lodash/includes';
+
+import commonModule from 'common/common.module';
+import loadingOverlay from 'common/components/loadingOverlay/loadingOverlay.component';
+import sessionService from 'common/services/session/session.service';
+import sessionModalService from 'common/services/session/sessionModal.service';
+
+import template from './homeSignIn.tpl';
+
+let componentName = 'homeSignIn';
+
+class HomeSignInController {
+
+  /* @ngInject */
+  constructor(sessionService, sessionModalService) {
+    this.sessionService = sessionService;
+    this.sessionModalService = sessionModalService;
+  }
+
+  $onInit() {
+    this.isSigningIn = false;
+    this.showSignInForm = true;
+
+    if ( includes( ['IDENTIFIED', 'REGISTERED'], this.sessionService.getRole() ) ) {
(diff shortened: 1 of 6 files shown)


src/app/homeSignIn/homeSignIn.component.js
```diff
@@ -0,0 +1,62 @@
...
+
+class HomeSignInController {
+
+  /* @ngInject */
+  constructor(sessionService, sessionModalService) {
+    this.sessionService = sessionService;
+    this.sessionModalService = sessionModalService;
+  }
+
+  $onInit() {
```
OzzieOrca: Can you test this? I think it'd be pretty easy. You might have to mock sessionService.

src/app/homeSignIn/homeSignIn.component.js
```diff
@@ -0,0 +1,62 @@
...
+        this.isSigningIn = false;
+        this.showSignInForm = false;
+      }, ( error ) => {
+        this.isSigningIn = false;
+        this.errorMessage = error.data.error;
+      } );
+  }
+
+  signUp() {
+    this.sessionModalService.signUp();
```
OzzieOrca: Can you test this too? I think it would just be a single spy to check if it has been called.

OzzieOrca:
Looks good, I'm just curious, where this will be used? Is this in the navbar or part of a panel somewhere? I guess idk what the home page looks like.

adammeyer:
It's going to be the sign in part about half way down on the home page.  http://crugive.site/homepage/

Omicron7:
👍

src/app/homeSignIn/homeSignIn.spec.js
```diff
@@ -0,0 +1,93 @@
...
+  describe( 'signUp', () => {
+    let deferred;
+    beforeEach( inject( function ( _$q_ ) {
+      deferred = _$q_.defer();
+      spyOn( $ctrl.sessionModalService, 'signUp' ).and.callFake( () => deferred.promise );
+      $ctrl.signUp();
+    } ) );
+
+    it( 'opens signUp Modal', () => {
+      expect( $ctrl.sessionModalService.signUp ).toHaveBeenCalled();
```
OzzieOrca: I don't think you need the calling strategy if you're only testing that it has been called and can leave it as `spyOn( $ctrl.sessionModalService, 'signUp' )`.

OzzieOrca:
The home page is cool. Hadn't seen it yet.
👍

PR #79, opened 2016-08-31
manage-giving component and child 'gift' component

denlight:


OzzieOrca:
These `give` and `gift` component names are really general. Are they specific to the user management part of the site? Would something like `manage-gifts`, `manage-partners`, `partner-details` or `giving-report` be more descriptive? I'm not quite sure all what these components represent do don't just blindly take one of my names. Maybe they are meant to be general but we haven't used them in any of the product selection/cart/checkout areas of the site yet...

OzzieOrca:
Cool. You can just rename the PR...

PR #82, opened 2016-09-01
Various fixes for review page (step 3)

OzzieOrca:
- Add submit button to cart summary panel for step 3
- Review template fixes: change headings, move change btns into panel headings, add donor/org name, remove "ministry" from designation number, change image placeholder size
- Fix issue with a single needinfo error where it tries to map the keys in the single object instead of treating it as an item in an array

Should fix issues from https://jira.cru.org/browse/EP-1199. ~~I'm still waiting to hear how they want the names formatted and~~ item 2 isn't fixed per comment there.

The interaction between the step 3 component and the cart summary components got a little more complicated than I wanted. Let me know if you have any suggestions.


adammeyer:
👍

PR #83, opened 2016-09-01
Add gift start date to review page

OzzieOrca:
Bob noticed this and [commented on Jira](https://jira.cru.org/browse/EP-484?focusedCommentId=42902&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-42902). The start date year doesn't seem to be stored. I'm not sure if we should guess and show the year for the next occurrence of that month. I asked that on Jira.


adammeyer:
👍

PR #85, opened 2016-09-01
Forgot Password, Reset Password and modalState service.

Omicron7:
This adds a Forgot Password Modal, Reset Password Modal and a new service modalState.

modalStateService has a provider that allows registering of named modals and  the method to display them. The service attaches a module.run method, which looks at the url #hash and displays the modal at app start. The service also has methods for setting and getting additional query param properties. This allows deeplinking to modals.

example: resetPasswordModal registers 'reset-password' names modal. If the application starts with #reset-modal in the url, the service will display the reset-password modal.

### DIFF
--- a/src/common/components/forgotPasswordModal/forgotPasswordModal.component.js
+++ b/src/common/components/forgotPasswordModal/forgotPasswordModal.component.js
@@ -1,0 +1,80 @@
+import angular from 'angular';
+import 'angular-gettext';
+import sessionService from 'common/services/session/session.service';
+import template from './forgotPasswordModal.tpl';
+
+let componentName = 'forgotPasswordModal';
+
+class ForgotPasswordModalController {
+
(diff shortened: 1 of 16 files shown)


OzzieOrca:
👍

src/common/components/forgotPasswordModal/forgotPasswordModal.component.js
```diff
@@ -0,0 +1,80 @@
...
+            break;
+          default:
+            this.errors['unknown'] = true;
+        }
+      } );
+  }
+
+  resetPasswordUrl() {
+    // Create in memory <a> to do url manipulation and set it to current location
+    let url = this.$document[0].createElement( 'a' );
```
adammeyer: nice.

adammeyer:
👍

PR #86, opened 2016-09-01
Added signIn tests. 

Omicron7:
This changes signIn so that it navigates to checkout whenever the session changes to REGISTERED. Navigating to this page with a REGISTERED session will automatically redirect to checkout. Signing in with the Sign In Button will also cause a redirect to checkout.


OzzieOrca:
This is the full page sign in before checkout right? I feel like I get it confused at first with all the modals. Maybe we could come up with a better name for this component. Idk. Anyways, 👍

PR #88, opened 2016-09-02
move search query to query string

adammeyer:


src/app/searchResults/searchResults.component.js
```diff
@@ -12,12 +11,15 @@ let componentName = 'searchResults';
...
   /* @ngInject */
-  constructor(designationsService) {
+  constructor($location, designationsService) {
+    this.$location = $location;
     this.designationsService = designationsService;
+  }
 
-    let searchTerm = urlTerm.get();
+  $onInit(){
+    let searchTerm = this.$location.search().q;
```
OzzieOrca: @adammeyer I don't think this works without the router running. I tried that with mpdx, broke it in production, and reverted back to this ugly thing https://github.com/CruGlobal/mpdx/blob/master/app/assets/javascripts/angular/services/urlParameter.js Hopefully there's something cleaner than that out there.

PR #89, opened 2016-09-02
Added deep linking to productConfig modal.

Omicron7:
Also removed modalState params. Didn't work correctly and is actually easier to use $location.


adammeyer:
👍

PR #90, opened 2016-09-02
show first name on home page instead of email

adammeyer:


OzzieOrca:
👍 First name on* homepage?

PR #91, opened 2016-09-03
Add array handling and nested zoom support to cortexApiService and hateoasHelperService

OzzieOrca:
- Add docs for cortexApiService


Omicron7:
👍

PR #92, opened 2016-09-03
Url query term lib

adammeyer:


adammeyer:
@OzzieOrca You where right about $location.search not working without router.  I made this lib to take care of that.  I'm merging now because the home page search has already been updated to use the query param.  Let me know if you have any input.

OzzieOrca:
Could use some tests but ya looks good.

PR #93, opened 2016-09-03
Thank You component

OzzieOrca:
This is based on https://github.com/CruGlobal/give-web/pull/91 and uses its nested features.

There were a couple of things I couldn't find on the purchases endpoint. I'll ask about these on Tuesday:
- Gift start date for recurring gifts
- Last 4 digits of credit card number. This also still isn't available for the review page.
- The single gift total seems to be missing from `ratetotals`. For now I'm using `monetary-total` which I'm not sure is right.


src/app/thank-you/account-benefits/account-benefits.tpl.html
```diff
@@ -0,0 +1,15 @@
...
+    <p>Benefits of registering for an account include:</p>
+    <ul>
+      <li>Save time during checkout</li>
+      <li>Update your contact info</li>
+      <li>View your donation history</li>
+      <li>Manage your recurring gifts</li>
+      <li>Change your bank account or credit card information</li>
+    </ul>
+    <a data-toggle="modal" data-target="#signupModal" class="btn btn-primary btn-block">Register Your Account</a>
+    <a data-toggle="modal" data-target="#signinModalRec">Register (User Match Alternate Modal)</a>
```
adammeyer: For now, remove this one and link the button to show the register modal from sessionModal.service.

adammeyer:
:+1 Looks great scotty.  Missing some translate tags on thank-you.tpl.html.

build/tasks/release.js
```diff
@@ -49,6 +49,7 @@ gulp.task('bundle', function () {
   var bundles = [
     'app/cart/cart.component',
     'app/checkout/checkout.component',
+    'app/thank-you/thank-you.component',
```
adammeyer: Is there a reason you used a hyphen here instead of camel case?

OzzieOrca: I like the naming conventions here https://angular.io/styleguide#!#naming but even I haven't been consistent with them. What do you think? The 3 of us should be on the same page but it doesn't matter to me whether that's camelCase of kebab-case for filenames. It looks like about half our filenames have dashes and the other half don't...

adammeyer: I noticed because this is the first AEM component that isn't going to using camel case.  I'd say to stick with camel case.  @Omicron7 do you have any thoughts?

Omicron7: I went with camelCase on everything since it looked like that was what the project was using initially.

OzzieOrca: It kinda looks like I started doing all the services in camelCase and all the components in kebab-case. I don't know why I did that. I must have made the first ones that way and just copied it... Opps

OzzieOrca:
Added translate tags and linked the Register your account btn to the sessionModalService

adammeyer:
Great!  Are you going to change the thank-you component to thankYou?

OzzieOrca:
Ya I can. Should I go through and mass rename all the other files in the app? I guess that doesn't belong in this PR.

adammeyer:
I would do that in a separate PR.  I only see display-address as the other one.

OzzieOrca:
Renamed them to camelCase

Omicron7:
👍

adammeyer:
👍

PR #94, opened 2016-09-05
Change submit label on add/edit

Omicron7:


src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -161,7 +161,7 @@ <h4 class="panel-title border-bottom-small mt" translate>
       <button class="btn btn-default btn-md btn-block-xs" type="button" ng-click="$dismiss()" translate>Cancel &amp; Return</button>
     </div>
     <div class="col-md-6 col-sm-6 text-right">
-      <button class="btn btn-md btn-primary btn-block-xs" type="submit" translate>Add to Gift Cart</button>
+      <button class="btn btn-md btn-primary btn-block-xs" type="submit">{{$ctrl.submitLabel|translate}}</button>
```
adammeyer: Do you need the translate here?  Looks like it is already translating in the controller.

Omicron7: gettext() in the controller is a NOOP. It exists only to allow the gulp/grunt tasks to find the translatable strings and add them to the PO file. translate in the element here actually does the translation.

adammeyer:
👍

PR #95, opened 2016-09-06
Search results

adammeyer:


adammeyer:
Sorry for the many commits, I'll squash those.

src/app/searchResults/searchResults.component.js
```diff
@@ -12,29 +12,52 @@ let componentName = 'searchResults';
...
   }
 
   $onInit(){
-    let searchTerm = urlQueryTerm.get('q');
-    if(searchTerm){
-      this.designationsService.createSearch(searchTerm)
-        .subscribe((id) => {
-          this.searchId = id;
-          this.getPageResults(1);
+    var params = this.$location.search();
```
OzzieOrca: Isn't this going to fail without the router?

adammeyer: @Omicron7 found a solution, he committed it directly to master. https://github.com/CruGlobal/give-web/commit/2302727087cdb011a0d6f1cdf308099f81d1dc35

src/common/services/api/designations.service.spec.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  describe('productSearch', () => {
+    it('should send a request to API and get results', () => {
+      self.$httpBackend.expectGET('https://cortex-gateway-stage.cru.org/search?keyword=steve').respond(200, searchResponse);
+      self.designationsService.productSearch({
+        keyword: 'steve'
+      })
+        .subscribe((data) => {
+          expect(data.length).toEqual(searchResponse.response.docs.length);
```
OzzieOrca: I think you should check to make sure at least the first one is mapped correctly and has the keys you expect.

src/app/searchResults/searchResults.component.js
```diff
@@ -12,29 +12,52 @@ let componentName = 'searchResults';
...
+
+    this.searchParams.keyword = params.q;
+    this.searchParams.first_name = params.fName;
+    this.searchParams.last_name = params.lName;
```
OzzieOrca: What's the purpose of copying these if `requestSearch()` copies them again? Do they need to be there even if the type isn't `people`?

src/app/searchResults/searchResults.component.js
```diff
@@ -12,29 +12,52 @@ let componentName = 'searchResults';
...
+    this.searchParams.type = params.type;
+    this.searchParamsInitial = angular.copy(this.searchParams);
+
+    this.requestSearch();
+  }
+
+  requestSearch(type){
+    if(angular.isDefined(type)){
```
OzzieOrca: Is `requestSearch()` called by anything other than `$onInit` when `this.searchParams` is set? Do you think this first block of logic would make more sense with what you currently have in `$onInit`? I'm a little confused why it seems like you are copying the first and last name into initial and then back.

OzzieOrca:
There are some translate tags missing but mostly looks good.

src/app/searchResults/searchResults.component.js
OzzieOrca: Ah ok cool

src/app/searchResults/searchResults.component.js
adammeyer: So if they choose ministries, it doesn't show the first and last name in the fields.  I thought that would be confusing to the user.

adammeyer: Yes, requestSearch can be called from the view.

src/app/searchResults/searchResults.component.js
Omicron7: We didn't have $locationProvider configured properly on production. Wasn't an issue with routers at all. You still need to use $window.location to reload or change the url, but we can use $location to read/write query params now.

OzzieOrca: Nice

PR #97, opened 2016-09-06
Changed deep linking to query params.

Omicron7:
Fragment was lost during 302 redirect. This changes deep linking to use a query parameter.
Adds urlFor helper for generating url's with modals attached.


OzzieOrca:
👍

PR #101, opened 2016-09-07
checkout step 3 loading indicator

adammeyer:


src/app/checkout/step-3/step-3.component.js
```diff
@@ -90,14 +92,17 @@ class Step3Controller{
...
         this.orderService.clearCardSecurityCode();
         this.onSubmitted();
         this.$window.location.href = 'thank-you.html';
       },
       (error) => {
+        this.submittingOrder = false;
         this.$log.error('Error submitting purchase:', error);
         this.onSubmitted();
         // TODO: show error message
+        this.errors = [error];
```
twinge: Does this resolve the TODO comment from the line above, or does more need to be done to show errors?

adammeyer: No, this line shows the default generic error message.  https://github.com/CruGlobal/give-web/blob/c9b4268903d7de7b5c30aea813d56548c470758d/src/app/checkout/step-3/step-3.tpl.html#L6

But I think we should at some point give the user a more detailed message of why their order failed.  Like API down vs credit card declined etc...

OzzieOrca:
👍

PR #102, opened 2016-09-07
In thank you component, avoid parsing billing address if it is not a credit card payment.

OzzieOrca:
I was getting undefined errors when loading a bank account payment


Omicron7:
:thumbsup:

PR #104, opened 2016-09-08
Payment icon

denlight:


denlight:
@adammeyer , @OzzieOrca I made a little component for payment method's icon. I basically has one input for a card Type(card-name). And if 'card-name' is undefined (in case of bank account) it will return a bank icon. Tell me what you think and I will refactor it if needed to match the requirements. Thanks

OzzieOrca:
@denlight I think I kinda replaced this with my paymentMethodDisplay component in https://github.com/CruGlobal/give-web/pull/106. Let me know if you see any features from this PR that should be brought over. I'm not sure if all the card type strings I'm checking for are exactly what the server is sending (same case and everything). I need to check the rarer ones.

OzzieOrca:
@denlight are you fine with me closing this?

PR #105, opened 2016-09-09
Fixes lost q param after closing modal.

Omicron7:
This fixes https://jira.cru.org/browse/EP-1237
This completely removes the code that was removing all query params when a modal closed. It will now be up to the individual modal components to remove query params they care about.


OzzieOrca:
👍

PR #106, opened 2016-09-12
Checkout existing payment methods

OzzieOrca:
This allows `REGISTERED` users to select one of their existing payment methods, if one exists.
https://jira.cru.org/browse/EP-1001
https://jira.cru.org/browse/EP-251

It currently does not:
- Allow a `REGISTERED` user to add a new payment method. The "Add New Payment Method" button and modal are not implemented yet. That will come in a seperate PR.
- Ask for a CCV if a credit card payment method is selected. Will come in another PR also once I figure out what the UI should be. Also I'm not sure if this prompt needs to happen for every credit card payment or only single gifts since I think only authorize.net needs it.

@denlight This should add two features you'll probably need for the profile payments methods management:
- The [paymentMethodDisplay component](https://github.com/CruGlobal/give-web/blob/a28f21904a43c771eb41786add947fe40c1caf42/src/common/components/paymentMethods/paymentMethodDisplay.component.js)

screen shot 2016-09-11 at 10 12 37 pm
- [getPaymentMethods](https://github.com/CruGlobal/give-web/blob/a28f21904a43c771eb41786add947fe40c1caf42/src/common/services/api/profile.service.js#L28) from the profile service to fetch existing payment methods for a user. I initially was using the profile endpoint for the payment methods in checkout but ended up using paymentmethodinfo in the order endpoint. This should both fetch the payment methods and sort them according to how David Sudarma suggested (bank accounts first, then credit cards with soonest expiration dates last)

Let me know if you have any issues using these.


wrandall22:
To answer your question, CCV is only required/displayed when giving single credit card gifts.

src/common/services/paymentHelpers/paymentValidation.spec.js
```diff
@@ -4,7 +4,7 @@ import ccp from 'common/lib/ccp';
 import { ccpStagingKey } from 'common/app.constants';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/of';
-import module from './paymentValidation.service';
+import module from './paymentValidation.service.js';
```
adammeyer: why is the .js extension needed here?

adammeyer:
👍  looks great!

src/common/services/paymentHelpers/paymentValidation.spec.js
OzzieOrca: I changed the folder it was in and missed it. Nice catch.

src/common/services/paymentHelpers/ccp.service.spec.js
```diff
@@ -5,7 +5,7 @@ import ccp from 'common/lib/ccp';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/of';
 
-import module from './ccp.service';
+import module from './ccp.service.js';
```
adammeyer: @OzzieOrca I noticed it here too

OzzieOrca: k thanks

PR #108, opened 2016-09-12
Session Enforcement

Omicron7:
Added timeout for give-session expiration. This allows detection of the cookie expiring so we can enforce REGISTERED role on pages.


src/common/services/session/session.service.js
```diff
@@ -110,28 +116,50 @@ function session( $cookies, $rootScope, $http, envService ) {
...
+      sessionTimeout = undefined;
+    }
+    // Decode give-session cookie and set timeout based on expiration
+    if ( angular.isDefined( encodedSession ) ) {
+      let giveSession = jwtDecode( encodedSession ),
+        timeout = new Date( giveSession.exp * 1000 ) - Date.now();
+      if ( timeout > 0 ) {
+        sessionTimeout = $timeout( timeout );
+        sessionTimeout.then( () => {
+          updateCurrentSession( $cookies.get( 'cortex-session' ) );
```
OzzieOrca: When this timeout has occurred, do we need to manually change the user's role from REGISTERED to IDENTIFIED? This doesn't fire off a cortex request that will come back with a different cookie so the cookie will still contain REGISTERED. This will be read by `sessionService.getRole()` in your enforcer service and the user will still look registered. Does this sound right or am I missing something?

OzzieOrca:
Looks good other than my question/comment.

src/common/services/session/session.service.js
Omicron7: Browsers delete the expired cookies, so $cookies.get() returns undefined. getRole() uses this to downgrade the role. The cortex-session role will still say 'REGISTERED', but getRole() will return to correct role.
I have tested it (allowed a session to actually expire while on the page) and it worked correctly.

OzzieOrca: Ok sounds good.

PR #109, opened 2016-09-12
Validate Credit Card Expiration Date

OzzieOrca:
Makes sure the card is not expired.
https://jira.cru.org/browse/EP-1251


adammeyer:
👍

PR #110, opened 2016-09-12
Fix gift comment collapsing

OzzieOrca:
In productConfig modal, disable click actions on comment headings when there is content in the box.
https://jira.cru.org/browse/EP-1201
With the old code, if you opened the comment, typed something, and tried to close it by clicking the heading an odd number of times while there was still content then it would instantly be closed if there was no content in the box.

This might not be the cleanest way to fix it but this keeps all the code in the view as it was before.


Omicron7:
👍

PR #113, opened 2016-09-14
Make step 3 review page errors better

OzzieOrca:
- Separate needinfo errors from submission errors since needinfo errors disable the submit button
- Show specific submission errors for things that fail clientside

@Omicron7 this should fix the issue with the submit button being disabled that you mentioned

### DIFF
--- a/src/app/checkout/step-3/step-3.component.js
+++ b/src/app/checkout/step-3/step-3.component.js
@@ -66,12 +66,12 @@ class Step3Controller{
   checkErrors(){
     this.orderService.checkErrors()
       .subscribe((data) => {
-        this.errors = data;
+        this.needinfoErrors = data;
       });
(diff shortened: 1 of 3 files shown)


src/app/checkout/step-3/step-3.component.js
```diff
@@ -81,6 +81,7 @@ class Step3Controller{
   }
 
   submitOrder(){
+    this.submissionError = '';
```
Omicron7: If submissionError isn't needed, you can `delete this.submissionError` instead of setting to empty string.

OzzieOrca: It's essentially pick your favorite version of false... I can `delete` it. I was just trying to remove the error from the top once a new request is sent.

PR #117, opened 2016-09-19
Add tests to step 1, homeSignIn, accountBenefits, and paymentMethodDisplay

OzzieOrca:
- Refactor step 1 component to be more testable
- Remove unneeded arg when testing loadDonorDetails in creditCardForm component
- Remove unused toPromise operator from designations service


adammeyer:
👍

PR #118, opened 2016-09-19
User matching

Omicron7:


src/app/thankYou/accountBenefits/accountBenefits.component.js
```diff
@@ -1,26 +1,51 @@
...
+      this.hasUserMatch = this.donorDetails['registration-state'] !== 'COMPLETED';
+    }
+  }
+
+  doUserMatch() {
+    if ( this.sessionService.getRole() === Roles.registered ) {
+      this.sessionModalService.userMatch();
+    }
+    else {
+      this.sessionModalService.userMatch().then( () => {
```
OzzieOrca: Is this supposed to be sign in? Or userMatch twice?

src/app/thankYou/accountBenefits/accountBenefits.component.js
```diff
@@ -1,26 +1,51 @@
...
 
+  $onChanges( changes ) {
+    // donorDetails is undefined initially
+    if ( changes.donorDetails && angular.isDefined( changes.donorDetails.currentValue ) ) {
+      // Show account benefits if registration state is NEW or MATCHED
+      this.hasUserMatch = this.donorDetails['registration-state'] !== 'COMPLETED';
+    }
+  }
+
+  doUserMatch() {
```
OzzieOrca: Could add tests

src/app/thankYou/accountBenefits/accountBenefits.component.js
```diff
@@ -1,26 +1,51 @@
...
+class AccountBenefitsController {
   /* @ngInject */
-  constructor(sessionModalService){
+  constructor( sessionModalService, sessionService ) {
     this.sessionModalService = sessionModalService;
+    this.sessionService = sessionService;
+    this.hasUserMatch = false;
   }
 
+  $onChanges( changes ) {
```
OzzieOrca: Could add tests

src/app/thankYou/thankYou.tpl.html
```diff
@@ -12,7 +12,8 @@ <h2 class="panel-name" translate>
                   </h2>
 
                   <p translate>
-                    We are processing your gift now. We'll send an email confirmation to to <strong>{{$ctrl.email}}</strong> as well as a tax-deductible receipt by postal mail.
+                    We are processing your gift now. We'll send an email confirmation to to
```
OzzieOrca: I noticed that there are 2 `to`s here. Could you fix that since you touched this line?

src/common/components/accountBenefitsModal/accountBenefitsModal.component.js
```diff
@@ -0,0 +1,47 @@
...
+    this.sessionService = sessionService;
+  }
+
+  $onInit() {
+    this.modalTitle = this.gettext( 'Register Your Account for Online Access' );
+  }
+
+  registerAccount() {
+    if ( this.sessionService.getRole() === Roles.registered ) {
+      // No need to sign if if we already are
```
OzzieOrca: `sign in if`?

src/common/components/accountBenefitsModal/accountBenefitsModal.component.js
```diff
@@ -0,0 +1,47 @@
...
+      this.onSuccess();
```
OzzieOrca: Could be tested

src/app/thankYou/thankYou.component.js
```diff
@@ -43,6 +49,14 @@ class ThankYouController{
         if(this.purchase.paymentMeans.self.type === 'elasticpath.purchases.purchase.paymentmeans'){ //only credit card type has billing address
           this.billingAddress = this.orderService.formatAddressForTemplate(this.purchase.paymentMeans['billing-address'].address);
         }
+
+        // Display Account Benefits Modal when GUEST checkout
```
OzzieOrca: Does this show the modal immediately on checkout? I thought that's what the button in the accountBenefits panel was for... It just seems like there are multiple ways to show the accountBenefits with duplicate code.

src/app/thankYou/accountBenefits/accountBenefits.component.js
```diff
@@ -1,26 +1,51 @@
...
+    // donorDetails is undefined initially
+    if ( changes.donorDetails && angular.isDefined( changes.donorDetails.currentValue ) ) {
+      // Show account benefits if registration state is NEW or MATCHED
+      this.hasUserMatch = this.donorDetails['registration-state'] !== 'COMPLETED';
```
OzzieOrca: I think this variable name is confusing... this is just a flag to determine if user match stuff needs to be shown, not if it has a match or has been matched right? I would think if the account `hasUserMatch` or is `COMPLETED` this wouldn't need to be shown. The boolean value seems to be the opposite of the variable name.

src/common/components/accountBenefitsModal/accountBenefitsModal.tpl.html
```diff
@@ -0,0 +1,17 @@
+<div class="col-xs-12">
```
OzzieOrca: Do you think it would be easy to use the same html and/or component as the accountBenefits component? I guess the button would have to change. Is this modal for use outside of the thank you component too?

Omicron7:
I agree, I don't think they both need to be there, but this is direct requirements from Jason Chandler.

src/app/thankYou/accountBenefits/accountBenefits.component.js
Omicron7: Good catch.

src/app/thankYou/thankYou.component.js
Omicron7: Yes, it immediately shows the modal when the user is a PUBLIC or has been matched. I was told the Account Benefits inside the thank-you page are if the user accidentally dismisses the modal. They both need to be present.

src/common/components/accountBenefitsModal/accountBenefitsModal.tpl.html
Omicron7: I thought about re-use, but the html was small enough and functionality differed enough to keep them separate.

OzzieOrca: Alright that's fine

src/app/thankYou/thankYou.component.js
OzzieOrca: Gotcha. Interesting... Should the panel be shown if you are public also? Cuz I don't think that is currently a condition here: https://github.com/CruGlobal/give-web/pull/118/files#diff-1d741c69c93c23b65b01e053a605ecf4R22 Although maybe `registration-state` will never be completed for a public user

Omicron7: PUBLIC user's always have a registration-state of MATCHED (They weirdly match themselves), so it will be shown.

OzzieOrca: Alright... Huh. If that's the case do we need the public check on the next line? (I probably shouldn't have started this thread on a comment...)

PR #131, opened 2016-09-22
s3 cache headers

adammeyer:
This will cache the dev env for 5 mins and stage/prod for 1 hour in user's browser and cloud front.


Omicron7:
Will this mean that it'll take up to an hour for changes in a recent deploy to show up for users, or does cloudfront automatically invalidate caches on new deploys/files?

adammeyer:
I was going to add the invalidation in another branch, but I can do that here.

adammeyer:
@Omicron7 I added the invalidation script

PR #132, opened 2016-09-22
Change addItem path to match latest EP.

Omicron7:
This changes the addItem path to 'itemfieldlineitem'. This will break until https://github.com/CruGlobal/give-ep-extensions/pull/220 is merged to stage.

### DIFF
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -115,7 +115,7 @@ function cart(cortexApiService){
     data.quantity = 1;
 
     return cortexApiService.post({
-      path: ['itemfieldlineitems', 'items', cortexApiService.scope, id],
+      path: ['itemfieldlineitem', 'items', cortexApiService.scope, id],
       data: data
     });
   }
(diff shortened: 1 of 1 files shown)


src/common/services/api/cart.service.js
```diff
@@ -115,7 +115,7 @@ function cart(cortexApiService){
     data.quantity = 1;
 
     return cortexApiService.post({
-      path: ['itemfieldlineitems', 'items', cortexApiService.scope, id],
+      path: ['itemfieldlineitem', 'items', cortexApiService.scope, id],
```
dsgoers: This needs to be `itemfieldslineitem`.

OzzieOrca:
I think this can go. I'm getting `Cannot find 'itemfieldlineitems' resource family.` errors when trying to click Add to Gift Cart on the productConfig modal.

mattdrees:
Yeah, sorry for not communicating well, but the EP renaming change went live early this evening.

PR #135, opened 2016-09-23
json nav

adammeyer:


src/common/components/nav/mobileNav.tpl.html
```diff
@@ -0,0 +1,158 @@
...
+        </a>
+      </li>
+    </ul>
+
+
+    <div class="tab-content">
+      <div role="tabpanel" class="tab-pane active" ng-if="mobileTab === 'nav'">
+        <div class="main-menu-container">
+          <ul class="main-menu" id="main-menu" ng-style="{'left': (($ctrl.mobileMenuLevel[0] * -100) + '%')}">
+            <li ng-repeat="i in $ctrl.menuStructure" ng-class="{'has-children': i.children.length}" ng-if="!i.hideInNav">
```
Omicron7: Could this be made into a component? It looks like it's hard coded to work only a specific number of levels, which might break if nav changes on AEM? You should be able to pass bindings of any information about current level. @OzzieOrca might have some ideas.

src/common/components/nav/nav.component.js
```diff
@@ -0,0 +1,70 @@
...
+      this.menuStructure = structure;
+    });
+  }
+
+  getNav() {
+    return Observable.from(this.$http({
+      method: 'GET',
+      url: this.navFeed,
+      headers: {
+        'Accept': undefined
```
Omicron7: can you pass 'application/json' here to have $http automatically parse the JSON?

src/common/components/nav/mobileNav.tpl.html
adammeyer: I made a component, let me know what your thoughts.

PR #138, opened 2016-09-26
Don't require CCV for existing credit card payments since we aren't collecting it

OzzieOrca:
- Store 'existing payment method' to ccv in session storage so step 3 knows it doesn't need to send a CCV

https://jira.cru.org/browse/EP-1327

### DIFF
--- a/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
+++ b/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
@@ -84,6 +84,7 @@ class ExistingPaymentMethodsController {
   selectPayment(){
     this.orderService.selectPaymentMethod(this.selectedPaymentMethod)
       .subscribe(() => {
+          this.orderService.storeCardSecurityCode('existing payment method');
           this.onSubmit({success: true});
         },
         (error) => {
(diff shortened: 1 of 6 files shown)


src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -84,6 +84,7 @@ class ExistingPaymentMethodsController {
   selectPayment(){
     this.orderService.selectPaymentMethod(this.selectedPaymentMethod)
       .subscribe(() => {
+          this.orderService.storeCardSecurityCode('existing payment method');
```
Omicron7: This should probably be made an exported const and used that way instead of hard coding the string everywhere.
`export const existingPaymentMethod = 'existing_payment_method';`
`include {existingPaymentMethod} from './filename'`

OzzieOrca: Alright. Where do you think the const should go? In the order service? Or app.constants.js?

Omicron7: I'd stick it in orderService since it's specific to that.

PR #140, opened 2016-09-26
Poll session expiration more frequently.

Omicron7:
I've been running into issues where the Sign In modal wasn't presenting after taking lunch. It seems that Chrome either doesn't count time, or ignores the timeout while the computer is "asleep". This changes the timeout to a 30 second or less (depending on give-session remaining time). This should catch and display Sign-In modal after extended periods of inactivity better now. Testing in browser with shortened timeouts functioned correctly.

I've got no clue how to accurately test this in a spec.


OzzieOrca:
It seems like it should still be firing as the restriction on background tabs only reduces the firing to once per second which shouldn't affect a long timeout. https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setTimeout#Timeouts_in_inactive_tabs_clamped_to_>1000ms

Idk. This is weird. I guess this works. I wonder if the browser doesn't delete the cookies when they expire if the tab is inactive. That could prevent https://github.com/CruGlobal/give-web/pull/140/files#diff-4da804074ed050928738f4f5581a3d9bR160 from calling `updateCurrentSession` and then a new timeout wouldn't be created since the diff between now and the expiry date is in the past (negative) https://github.com/CruGlobal/give-web/pull/140/files#diff-4da804074ed050928738f4f5581a3d9bR148.

You should be able to mock the date in jasmine and then tick forward. http://jasmine.github.io/2.4/introduction.html#section-Mocking_the_Date That might lead to some fun scope apply stuff. It still might be complicated to test... What do you think?

Omicron7:
The restriction on inactive tabs clamping to 1 sec are for timeouts that are for less than 1000ms.

What actually appears to be happening is that the browsers don't accurately track time when the computer is asleep. From what I can tell, JS doesn't make any guarantees that setTimeout/setInterval will wait for a concurrent number of time, just that it'll be that much time or more.

I read into the not all browsers remove expired cookies, but thought I determined that's been fixed in modern ones. I'll add a check though to see if it's expired based on exp in the JWT and manually delete it if it is.

I looked into the jasmine time mocking, it didn't look like it'd help with setTimeout though. Few stackoverflows I could find said you test setTimeout, you'd need to disable asynchronous testing, which we don't want to do. The difficulty for me is that the methods we need to test and add coverage to are all private, and can only be called from a browser timeout.

OzzieOrca:
This seems to work:
```js
describe('setTimeout', () => {
  beforeEach(function() {
    jasmine.clock().install();
  });

  afterEach(function() {
    jasmine.clock().uninstall();
  });

  it('setTimeout should work', () => {
    let timeoutFinished = false;
    setTimeout(() => {
      timeoutFinished = true;
    }, 3000);
    jasmine.clock().tick(3000);
    expect(timeoutFinished).toEqual(true);
  });
  it('setTimeout should not fire if the time is not long enough', () => {
    let timeoutFinished = false;
    setTimeout(() => {
      timeoutFinished = true;
    }, 3000);
    jasmine.clock().tick(2999);
    expect(timeoutFinished).toEqual(false);
  });
});
```

You could also publish those private methods and prefix them with a `_` to indicate they are private, just for testing.

Omicron7:
@OzzieOrca Mind reviewing this again since I added the specs and checks for browsers that don't delete cookies?

PR #143, opened 2016-09-28
[#EP-1249] Prevent multiple order submissions

Omicron7:
This prevents an order for getting submitted multiple times as well as displays a progress (loading) over cart-summary while submitting. This prevents clicking "Edit Cart" while an order is submitting as well.


src/app/checkout/step-3/step-3.component.js
```diff
@@ -132,6 +137,8 @@ export default angular
       cartData: '<',
       submit: '<',
       onSubmitBtnChangeState: '&',
-      onSubmitted: '&'
+      onSubmitted: '&',
+      onSubmittingOrder: '&',
+      submittingOrder: '<'
```
OzzieOrca: Do you think this would be better staying local (by not having the input)? It just took me a bit to figure why it was going out the output and coming right back in the input of the component. It is nice that the onSubmittingOrder call can set it everywhere with one call though. I'm fine with either.

PR #145, opened 2016-09-28
donation start day calculation

adammeyer:


src/app/cart/cart.component.js
```diff
@@ -60,8 +60,17 @@ class CartController {
     this.$window.location.href = this.sessionService.getRole() === 'REGISTERED' ? 'checkout.html' : 'sign-in.html';
   }
 
-  donationStartDate( month, day ) {
-    return new Date( 0, Number( month ) - 1, Number( day ) );
+  donationStartDate( config ) {
```
adammeyer: @OzzieOrca @Omicron7 Before I write tests, I wanted to get your feedback on this function.  It's getting the start day for each donation in the cart.  Do think this I should do this as part of the "loadCart" function instead?  I'm open to suggestions.

src/common/services/api/cart.service.js
```diff
@@ -134,6 +135,23 @@ function cart(cortexApiService){
...
     });
   }
+
+  function giftStartDate(day){
+    return cortexApiService.get({
+      path: ['nextdrawdate'],
+      cache: true
+    }).map((data) => {
+      let drawDate = data['next-draw-date'].split('-');
+      drawDate = new Date(drawDate[0], (drawDate[1] - 1), drawDate[2]);
```
Omicron7: I believe if next-draw-date is 'YYYY-MM-DD', you can just pass it to new Date();

src/app/cart/cart.component.js
OzzieOrca: So when you load the cart, do you have all the data you need? You just take the configured day from the EP cart and figure out the next time that day occurs on or after the next Siebel draw day right? I think this could be good as part of the cart service as it could be used throughout checkout. Although I could see this date getting off if they go through checkout slowly and the Siebel draw day changes.

src/app/productConfig/productConfig.modal.js
```diff
@@ -134,12 +129,20 @@ class ModalInstanceCtrl {
       } );
   }
 
-  daysInMonth( month ) {
-    var daysInMonth = new Date( 2001, month, 0 ).getDate();
-    return range( 1, daysInMonth + 1 ).map( function ( n ) {
+  daysInMonth() {
+    return range( 1, 29 ).map( function ( n ) {
```
OzzieOrca: Should this be 29? What happens when February only has 28 days?

src/app/cart/cart.component.js
Omicron7: Once the recurring order is completed, we have an actual next-draw-date from siebel. I think this is just for new/none completed recurring items. I do agree, it would probably help to add it to the cartService so any orders already have it.

src/common/services/api/cart.service.js
OzzieOrca: I was thinking the same thing but how come `new Date('2014-12-01')` gives me `Sun Nov 30 2014 16:00:00 GMT-0800 (PST)` and `new Date('2014-12-02')` gives me `Mon Dec 01 2014 16:00:00 GMT-0800 (PST)`? The day seems off by one... I usually use [moment](http://momentjs.com/) for dates...

Omicron7: I believe "new Date(string)" uses GMT/UTC to create the date (unless string contains timezone info).  toString or printing to console converts it to your timezone before printing.

adammeyer: Correct, it assumes it's GMT.  That's why I did it this way.

src/app/productConfig/productConfig.modal.js
adammeyer: This returns 1 through 28.  When I had 28 there it returned through 27.

src/common/services/api/cart.service.js
adammeyer: This way imports as the current local day.

src/app/cart/cart.component.js
adammeyer: It is part of the cart service.  This code it pulling it and showing it.  I was wondering if you had any feedback on what I came up with to prevent digest loops.

src/app/productConfig/productConfig.modal.js
OzzieOrca: Oh right. The upper bound of range is exclusive. It just reads weird.

src/app/cart/cart.component.js
OzzieOrca: Can the cart service call `this.cartService.giftStartDate` and map the cart line items to include it? I'm a little confused why sometimes we are using `gift-start` and sometimes have to go compute it from `recurring-day-of-month`.

adammeyer: That was my first idea, but since there is an async http call involved, I wasn't sure on the best way to do it.

OzzieOrca: I think you could do the `giftStartDate` request at the top of the `get` function in cart service and use something like:
```
Observable.forkJoin(cartRequestObservable, nextDrawDateObservable)
  .map(([cartResponse, nextDrawDateResponse]) => {
    console.log(cartResponse, nextDrawDateResponse);
    //do stuff
  });
```
to get both requests at once.

adammeyer: @OzzieOrca That worked great!  Thanks for the idea.

OzzieOrca: Cool. Code looks good. Tests are a little lacking but most of that is from the cart service which never really got tested.

adammeyer: Yup, I'll take care of those tomorrow.

src/app/cart/cart.tpl.html
```diff
@@ -32,7 +32,7 @@ <h3 class="panel-name" translate>Your Gift Cart</h3>
                         <span class="giftsum-frequency giftsum-title">{{i.frequency}}</span>
                         <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'">
                           <translate>Starts on:</translate>
-                          {{$ctrl.donationStartDate(i.config['start-month'], i.config['start-day']) | date:'MMM d'}}
+                          {{i.giftStartDate | date:'MMM d, yyyy'}}
```
OzzieOrca: This seems to be missing a space. The view looks like `Starts on:Oct 4, 2016`. If you move the translate up to be an attribute on the span that might work.

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -87,38 +87,23 @@ <h4 class="panel-title border-bottom-small mt0 pt-" translate>
...
+                              ng-change="$ctrl.changeStartDay($ctrl.itemConfig['recurring-day-of-month'])">
                       </select>
                     </div>
                   </div>
                 </div>
-                <div class="col-sm-2">
+                <div class="col-sm-4">
                   <div class="form-group">
-                    <label translate>Day</label>
+                    <label>Will start on</label>
```
OzzieOrca: Do you think this should be hidden if there is no day selected yet? It seems a little weird with just a floating heading. Or actually the day is supposed to be defaulted to the nextDrawDate so this won't be an issue.

It could use a `translate` attribute though...

src/app/thankYou/thankYou.tpl.html
```diff
@@ -125,7 +125,7 @@ <h4 class="panel-name" translate>Gift Summary</h4>
                               <td class="td-frequency">
                                 <span class="giftsum-about giftsum-title">{{item.rate.display}}</span>
                                 <span class="giftsum-frequency giftsum-title">
-                                  <span ng-if="item.rate.recurrence.interval !== 'NA'" translate>First Gift: XX/XX/XXXX</span>
+                                  <span ng-if="item.rate.recurrence.interval !== 'NA'" translate>First Gift: {{item.rate['start-date'].value | date:'MMM d, yyyy':'UTC'}}</span>
```
adammeyer: @OzzieOrca I'm a little confused on this?  The dates should all be in the local timezone, so won't this add or remove hours to the correct date?

src/app/productConfig/productConfigModal.tpl.html
adammeyer: Correct, so I'll leave it as is.  The set default day task will fix that.

adammeyer:
@Omicron7 @OzzieOrca I added some specs.  Please review as soon as you can so I can unblock Bob.  Thanks!

src/app/thankYou/thankYou.tpl.html
Omicron7: date:'mediumDate' might be a better choice here, and it's equivalent. If this ever gets used in other locales, it'll be localized to the language specific date format.

adammeyer: Good call.  Done.

Omicron7: In you-giving, I've been using
{{rate['start-date'].display | date:'mediumDate'}}

We may want to pick one and normalize all of these throughout the code at some point.

PR #146, opened 2016-09-28
[#EP-816] Validate gift amount is more than $1, less than $10 million…

Omicron7:
…, is a number and no more then 2 decimal places.

### DIFF
--- a/src/app/productConfig/productConfigModal.tpl.html
+++ b/src/app/productConfig/productConfigModal.tpl.html
@@ -48,14 +48,20 @@
               <span ng-repeat-end></span>
               <label class="btn btn-default-form u-textLeft custom-amount">
                 <div class="form-group form-group-default u-inline">
-                  <input class="form-control form-control-subtle number" type="text" name="amount"
+                  <input class="form-control form-control-subtle number"
+                         type="number"
+                         name="amount"
                          ng-model="$ctrl.customAmount"
                          ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
-                         ng-pattern="/^[1-9]\d*(\.\d+)?$/">
+                         ng-min="1"
+                         ng-max="10000000"
+                         ng-pattern="/^[0-9]+(\.[0-9]{1,2})?$/">
                 </div>
               </label>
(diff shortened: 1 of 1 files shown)


src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -48,14 +48,20 @@ <h4 class="panel-title border-bottom-small" translate>
...
-                         ng-pattern="/^[1-9]\d*(\.\d+)?$/">
+                         ng-min="1"
+                         ng-max="10000000"
+                         ng-pattern="/^[0-9]+(\.[0-9]{1,2})?$/">
                 </div>
               </label>
               <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error" ng-if="($ctrl.itemConfigForm.amount | showErrors)">
-                <div class="help-block" ng-message="pattern" translate>Amount must be a valid amount and at least $1</div>
+                <div class="help-block" ng-message="min" translate>Amount must be at least $1.00</div>
+                <div class="help-block" ng-message="max" translate>Amount must be less than $10,000,000.00</div>
```
OzzieOrca: `Amount cannot be more than $10,000,000.00`? The wording for less than equal is always wierd... ng-max should allow a million flat though right?

Omicron7: Yes, it'll allow the max, but nothing more.

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -48,14 +48,20 @@ <h4 class="panel-title border-bottom-small" translate>
               <span ng-repeat-end></span>
               <label class="btn btn-default-form u-textLeft custom-amount">
                 <div class="form-group form-group-default u-inline">
-                  <input class="form-control form-control-subtle number" type="text" name="amount"
+                  <input class="form-control form-control-subtle number"
```
adammeyer: input type number can't be used here.  Jason doesn't want the up/down arrows to show on the input, that's why it was using text.

A validation function may need to be written.

Omicron7: so we throw out all the html browsers validation and roll our own?

OzzieOrca: You could look into hiding the spinner with CSS...

PR #165, opened 2016-10-06
Profile Payment Methods

denlight:


denlight:
@adammeyer , so far EditPaymentMethod is a dummy modal. I will pick up this task today. Just wanted to deliver all of the above first before the end of the sprint.

src/common/components/paymentMethods/bankAccountForm/bankAccountForm.tpl.html
```diff
@@ -19,10 +19,10 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Bank Account Pay
         <div class="form-group form-group-default form-group-inline" ng-class="{'has-error': ($ctrl.bankPaymentForm.accountType | showErrors)}">
           <label class="label-block">Account Type</label>
           <label class="radio-inline">
-            <input type="radio" name="accountType" ng-model="$ctrl.bankPayment.accountType" value="checking" required> Checking <span class="hidden-md">Account</span>
+            <input type="radio" name="accountType" ng-model="$ctrl.bankPayment.accountType" value="Checking" required> Checking <span class="hidden-md">Account</span>
```
OzzieOrca: Thanks for updating these :)

src/common/components/paymentMethods/deletePaymentMethod/deletePaymentMethod.modal.component.js
```diff
@@ -0,0 +1,276 @@
...
+
+        for(let i=0;i<4;i++){
+          months.push(month);
+          month+=3;
+          month = month > 11 ? month - 12 : month;
+        }
+        months.sort((a,b) => {
+          return a-b;
+        });
+        text = `On the ${gift['recurring-day-of-month']}th day of each ${monthNames[months[0]]}, ${monthNames[months[1]]}, ${monthNames[months[2]]} and ${monthNames[months[3]]}`;
```
OzzieOrca: I might try to do more of this in the template... You could use ng-switch there. Idk. As long as you can make this work with angular-gettext, I don't care where you do this logic though... I'm just more familiar with doing it in the template. I think you've got a few places where you are constructing strings in JS. Just make sure they'll be translated correctly. Angular filters can handle some of the localization stuff which might be nice in the future.

package.json
```diff
@@ -76,6 +77,7 @@
       "jsencrypt": "npm:jsencrypt@^2.3.1",
       "jwt-decode": "github:auth0/jwt-decode@^2.1.0",
       "lodash": "npm:lodash@^4.14.2",
+      "querystring": "github:jspm/nodelibs-querystring@^0.1.0",
```
OzzieOrca: From Brian's comment, I guess if you can avoid using this that would be good. If you think it has better features or think `$location.search()` is insufficient you can push back.

src/app/main/main.component.js
```diff
@@ -44,6 +46,10 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
       url: "/thank-you.html",
       template: '<thank-you></thank-you>'
     })
+    .state('profile', {
+      url: "/profile.html",
```
OzzieOrca: I guess the default profile page doesn't exist yet. You can use query params for now but idk what the final version should use. @adammeyer What links do we want to use in AEM? Do we have a profiles directory like `give.cru.org/profile/payment-methods.html` or is everything based off the root folder as its own html file?

src/app/profile/payment-methods/recurring-gifts/recurring-gifts.component.js
```diff
@@ -0,0 +1,53 @@
...
+      forEach(gift['donation-lines'], (donationLine) => {
+        donationLine['rate'] = rate;
+        donationLine['recurring-day-of-month'] = day;
+        gifts = gifts.concat(donationLine);
+      });
+    });
+    return gifts;
+  }
+
+  getNextGiftDate(gift){
```
OzzieOrca: Could Adam's code https://github.com/CruGlobal/give-web/blob/master/src/common/services/api/cart.service.js#L153 work here? If you want to use it I guess it might be good to put it in a more general service that is not tied to the cart.

You can have angular format the date in the template: https://github.com/CruGlobal/give-web/blob/master/src/app/cart/cart.tpl.html#L35

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.spec.js
```diff
@@ -127,6 +129,7 @@ describe('credit card form', () => {
     });
     it('should send a request to save the credit card payment and billing address info', () => {
       self.controller.creditCardPayment = {
+        'address': formatAddressForCortex(self.controller.billingAddress),
```
OzzieOrca: Was this test not working without this? https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L112 should fill it in depending on if we are using mailing address or billing address.

denlight: that got leaked in. It works with or without =) I'll remove

src/app/main/main.component.js
```diff
@@ -1,6 +1,7 @@
 import 'babel/external-helpers';
 import angular from 'angular';
 import 'angular-ui-router';
+import 'angular-animate';
```
denlight: @OzzieOrca , angular-animate is required by bootstrap ui to make smooth transitions. Otherwise they are choppy.

src/app/main/main.component.js
denlight: yes, profile page just adds another link to the top 'Payment Methods'. That is about it =) But it serves as a parent component for 'payment methods' and I assume it will be for manageGiving etc

src/app/profile/payment-methods/payment-methods.tpl.html
```diff
@@ -0,0 +1,34 @@
+<loading-overlay ng-if="$ctrl.loading"></loading-overlay>
```
OzzieOrca: This seems to take over the entire page, including the menu so a user can't navigate away. Can you nest this farther in so it only covers the panel? Or it might be nice to keep the Add New Payment Method button accessible during loading... idk.

src/app/profile/payment-methods/bank-account/bank-account.component.js
```diff
@@ -0,0 +1,73 @@
...
+      resolve: {
+        model: () => this.model,
+        paymentType: () => 'bankAccount',
+        submissionError: () => this.submissionError
+      }
+    });
+  }
+
+  deletePaymentMethod(){
+    this.deletePaymentMethodModal = this.$uibModal.open({
```
OzzieOrca: This seems to be missing the alert as seen at http://crugive.site/payment-methods/?success when the modal is closed. I'm not sure if that alert should be `alert-danger` or `alert-success`...  Red makes it seem like it failed. All the dangerous stuff has already been confirmed.

src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -0,0 +1,106 @@
...
+    this.log = $log;
+    this.$uibModal = $uibModal;
+    this.paymentMethod = 'bankAccount';
+    this.profileService = profileService;
+    this.loading = false;
+    this.submissionError = {error: ''};
+  }
+
+  $onDestroy(){
+    this.addNewPaymentMethodModal.close();
```
OzzieOrca: What if `this. addNewPaymentMethodModal` is undefined?

src/app/profile/payment-methods/bank-account/bank-account.tpl.html
```diff
@@ -0,0 +1,60 @@
+  <div class="panel panel-default panel-with-gray-background">
+    <div ng-click="$ctrl.isCollapsed = !$ctrl.isCollapsed" class="panel-heading" role="tab">
+      <div ng-class="{'collapsed': $ctrl.isCollapsed}" class="payment-summary-link p- collapse-indicator-row" role="button">
+        <div class="row">
+          <div class="col-md-8">
+            <p>
+              <img class="hidden-xs payment-icon" ng-src="{{$ctrl.imgDomain}}/assets/img/icon-bank.png"/>
```
OzzieOrca: Same thing here. Can we use the `paymentMethodDisplay` component?

src/app/profile/payment-methods/credit-card/credit-card.component.js
```diff
@@ -0,0 +1,84 @@
...
+  constructor(envService,$uibModal){
+    this.isCollapsed = true;
+    this.$uibModal = $uibModal;
+    this.imgDomain = envService.read('imgDomain');
+    this.submissionError = {error: ''};
+  }
+
+  $onInit(){
+    let address = this.model.creditcard ? this.model.creditcard.address : false;
+    this.formattedAddress = address ? formatAddressForTemplate(address) : false;
```
OzzieOrca: Can you make the profile service do this? Just to be consistent. You can probably just copy what is here https://github.com/CruGlobal/give-web/blob/master/src/common/services/api/profile.service.js#L80-L88 or turn that into a function that can be called by both profile service methods.

system.config.js
```diff
@@ -37,10 +38,14 @@ System.config({
     "lodash": "npm:lodash@4.15.0",
     "plugin-babel": "npm:systemjs-plugin-babel@0.0.13",
     "plugin-babel-runtime": "npm:babel-runtime@5.8.38",
+    "querystring": "github:jspm/nodelibs-querystring@0.1.0",
```
OzzieOrca: jspm should have deleted this... I think running `jspm clean` should get rid of them if they are no longer in package.json

system.config.js
```diff
@@ -68,6 +73,9 @@ System.config({
     "github:jspm/nodelibs-process@0.1.2": {
       "process": "npm:process@0.11.8"
     },
+    "github:jspm/nodelibs-querystring@0.1.0": {
```
OzzieOrca: Same here with jspm

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.modal.tpl.html
```diff
@@ -0,0 +1,72 @@
...
+              </span>
+              <span ng-switch-default translate>
+                There was an unknown error saving your payment. Please verify all your info and try again. If you are still seeing this message, contact support.
+              </span>
+            </p>
+          </div>
+          <div class="panel panel-default panel-with-gray-background">
+            <div class="panel-body">
+              <p>
+                <i class="fa fa-cc-{{$ctrl.getIcon()}}" aria-hidden="true"></i>
```
OzzieOrca: Same here with `paymentMethodDisplay`

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.modal.tpl.html
```diff
@@ -0,0 +1,72 @@
...
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-body pt0">
+    <div class="container-fluid" tabindex="-1" role="dialog">
+      <div class="row">
+        <div class="col-md-12 col-xs-12">
+          <div class="alert alert-danger" role="alert" ng-if="$ctrl.resolve.submissionError.error">
+            <p ng-if="$ctrl.resolve.submissionError.error" ng-switch="$ctrl.resolve.submissionError.error">
```
OzzieOrca: I think we should make a component for this or include it in an existing one. I've copied and pasted it a couple of places already. You can leave it for now. I can make a PR for it or you can if you get to it. Maybe it should go in https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/addNewPaymentMethod/addNewPaymentMethod.tpl.html. I guess earlier I was hesitant to have to pass the errors further down the component tree.

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.modal.component.js
```diff
@@ -0,0 +1,57 @@
...
+import paymentMethodDisplay from 'common/components/paymentMethods/paymentMethodDisplay.component';
+
+import loadingOverlay from 'common/components/loadingOverlay/loadingOverlay.component.js';
+
+import bankAccountForm from '../bankAccountForm/bankAccountForm.component';
+import creditCardForm from '../creditCardForm/creditCardForm.component';
+
+let componentName = 'editPaymentMethodModal';
+
+class editPaymentMethodModalController{
```
OzzieOrca: How hard do you think it would be to use https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/addNewPaymentMethod/addNewPaymentMethod.component.js for this? We could add that alert at the top that shows the current card details. We could rename it to addUpdatePaymentMethod.component.js or something. That might be too long... You could leave it for now though and we can get to it later. I'm going to need this or the combined component for the edit recurring gifts modal.

src/app/profile/payment-methods/bank-account/bank-account.component.js
```diff
@@ -0,0 +1,73 @@
+import angular from 'angular';
+import template from './bank-account.tpl';
+import displayAddressComponent from 'common/components/display-address/display-address.component';
+import editPaymentMethodModal from 'common/components/paymentMethods/editPaymentMethod/editPaymentMethod.modal.component';
+import deletePaymentMethodModal from 'common/components/paymentMethods/deletePaymentMethod/deletePaymentMethod.modal.component.js';
+import giveModalWindowTemplate from 'common/templates/giveModalWindow.tpl';
+import recurringGiftsComponent from '../recurring-gifts/recurring-gifts.component';
+
+
+class BankAccountController{
```
OzzieOrca: Do you think this and the credit-card component could be combined? It looks like a lot of the same code. The template could show different content depending on the payment type.

src/common/components/paymentMethods/editPaymentMethod/editPaymentMethod.modal.component.js
denlight: @OzzieOrca, I was also thinking about reusing your modal, but was a little hesitant because I'm also a little afraid it can get a little too messy. I might be wrong. Let's keep it that way and after all is set and done with adding and editing we can take a look at the whole picture and make a decision about combining vs keeping separate.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -64,8 +65,9 @@ class ExistingPaymentMethodsController {
   openAddNewPaymentMethodModal(){
     this.addNewPaymentMethodModal = this.$uibModal.open({
       component: 'addNewPaymentMethodModal',
-      size: 'large new-payment-method-modal',
+      size: 'lg',
```
denlight: I totally was =) I'll fix it

src/common/components/paymentMethods/deletePaymentMethod/deletePaymentMethod.modal.tpl.html
```diff
@@ -0,0 +1,164 @@
...
+  <div class="modal-body pt0">
+    <div class="container-fluid" tabindex="-1" role="dialog">
+      <div class="row">
+        <div class="col-md-12 col-xs-12">
+          <form>
+
+            <div ng-if="$ctrl.view != 'addPaymentMethod'" class="panel panel-default bg-danger">
+              <div class="panel-body">
+                <p>
+                  <span class="payment-icon">
```
denlight: It ended up looking like this - 
screen shot 2016-10-10 at 11 05 03 am

denlight: =) I'll modify paymentMethodDisplay component to output fa icons as an option

Omicron7: Add <div class="radio-method"></div> around the payment-method-display to fix that large image issue. I ran into the same thing in the your-giving component.

src/app/profile/payment-methods/bank-account/bank-account.component.js
denlight: @OzzieOrca , I made a hardcoded solution for now but I think we can combine error handling and success messages into one component. BTW, according to the David Sudarma we will fade out success messages after 60 seconds.

src/common/components/paymentMethods/deletePaymentMethod/deletePaymentMethod.modal.tpl.html
denlight: @Omicron7 , I also found a solution by adding 'payment-icon' class to the icon image. So I updated 'paymentMethodDisplay' component's view and all img tags have that class now along with 'method-logo'. @OzzieOrca , @adammeyer , what do you guys think would be a better solution?

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  editPaymentMethod() {
+    this.editPaymentMethodModal = this.$uibModal.open({
+      component: 'paymentMethodFormModal',
+      windowTemplateUrl: giveModalWindowTemplate.name,
+      resolve: {
+        paymentMethod: () => this.model,
+        mailingAddress: () => {}, // TODO: load this from donor details
+        onSubmit: () => this.onSubmit,
```
OzzieOrca: You should be able to call https://github.com/CruGlobal/give-web/blob/update-payment-methods/src/common/services/api/profile.service.js#L162 to update it. An example is here https://github.com/CruGlobal/give-web/blob/update-payment-methods/src/app/profile/yourGiving/editRecurringGifts/step0/addUpdatePaymentMethod.component.js#L38

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,85 @@
...
+    return this.model['card-number'] ? true : false;
+  }
+
+  editPaymentMethod() {
+    this.editPaymentMethodModal = this.$uibModal.open({
+      component: 'paymentMethodFormModal',
+      windowTemplateUrl: giveModalWindowTemplate.name,
+      resolve: {
+        paymentMethod: () => this.model,
+        mailingAddress: () => {}, // TODO: load this from donor details
```
OzzieOrca: You'll need to implement this. It is needed so a use can chose to use their mailing address for their billing address. I moved it outside cuz the order and profile endpoints are different. You should be able to just use https://github.com/CruGlobal/give-web/blob/update-payment-methods/src/app/profile/yourGiving/editRecurringGifts/step0/addUpdatePaymentMethod.component.js#L25 to load it.

src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -0,0 +1,124 @@
...
+      },60000);
+    });
+  }
+
+  onSubmit(e) {
+    if(e.data) {
+      this.profileService.addPaymentMethod(e.data)
+        .subscribe((data) => {
+          this.submitted = false;
+          this.parentComponent.paymentMethodFormModal.close(data);
```
OzzieOrca: I don't think this needs `parentComponent` as it is available on `this`.

denlight: @OzzieOrca , I was trying to see what 'this' has and it seamed like there wasn't a way to reach modal's instance so I had to do this ugly hack. Are you saying I should be able to do this.close()?

OzzieOrca: I would think `this.paymentMethodFormModal.close(data);` would work as you set it above. Sometimes `this` is undefined in the inspector as it is actually `_this` or something. You could try disabling source maps (chrome dev tools setting) and see if you can view what's in the object.

denlight: @OzzieOrca , in this case PaymentMethodsController's function onSubmit has scope of PaymentMethodFormModalController which doesn't have access to the modal instance. But you have those bindings down at PaymentMethodFormModalController resolve and dismiss. I was trying to call 'this.dismiss' inside PaymentMethodsController's onSubmit function and still no luck. It is kind of tricky and I am out of options. Could you please take a look at it and see if you can figure out a better way to close the modal on successful save? Thanks, man!

src/app/profile/profile.component.js
```diff
@@ -0,0 +1,47 @@
...
+
+  /* @ngInject */
+  constructor(sessionEnforcerService, $window, $location) {
+    this.sessionEnforcerService = sessionEnforcerService;
+    this.$location = $location;
+    this.$window = $window;
+  }
+
+  $onInit() {
+    this.enforcerId = this.sessionEnforcerService([Roles.registered], {
```
denlight: sounds good, I'll move getNextGiftDate() to donationsService.

src/app/profile/payment-methods/payment-method/payment-method.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    },
+    uibModal = jasmine.createSpyObj('$uibModal', ['open','close']);
+
+    uibModal.open.and.callFake(fakeModal);
+
+  beforeEach(inject(function ($rootScope, $componentController) {
+    var $scope = $rootScope.$new();
+
+    self.controller = $componentController(module.name, {
+      $scope: $scope,
```
denlight: no worries. I'll do it a new way next time

src/common/services/api/donations.service.spec.js
```diff
@@ -65,4 +65,15 @@ describe( 'donations service', () => {
       $httpBackend.flush();
     } );
   } );
+
+  describe( 'getNextGiftDate( data )', () => {
+    it('should give next gift date', () => {
+      let data = {};
+      expect(donationsService.getNextGiftDate(data)).toBe(false);
+      data['recurring-day-of-month'] = '25';
+      // im not sure how to test this function without duplicating its functionality into here =))
```
OzzieOrca: I'd use jasmine to mock the date and then test a couple of inputs and outputs. If current date (mocked by jasmine) is 8/10/2015 and recurring day of month is 9, I expect the output to be 9/9/2015. I'd just do a normal case and then one where it has to jump to the next month. You can do more if you want.

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,118 @@
...
+      resolve: {
+        paymentMethod: () => this.model,
+        mailingAddress: () => this.mailingAddress,
+        onSubmit: () => this.onSubmit,
+        submitted: () => this.submitted,
+        submissionError: () => this.submissionError,
+        profileService: () => this.profileService
+      }
+    });
+    this.editPaymentMethodModal.result.then(() => {
```
OzzieOrca: I don't think you need these lines. If the modal has a result, that means it has already been closed.

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,118 @@
...
+
+  editPaymentMethod() {
+    this.editPaymentMethodModal = this.$uibModal.open({
+      component: 'paymentMethodFormModal',
+      windowTemplateUrl: giveModalWindowTemplate.name,
+      resolve: {
+        paymentMethod: () => this.model,
+        mailingAddress: () => this.mailingAddress,
+        onSubmit: () => this.onSubmit,
+        submitted: () => this.submitted,
```
OzzieOrca: This shouldn't be needed as it is handled in paymentMethodFormModal since that's where the buttons are.

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,118 @@
...
+        submissionError: () => this.submissionError,
+        profileService: () => this.profileService
```
OzzieOrca: I don't think it needs the profileService either.

denlight: @OzzieOrca , if I don't do that then `onSubmit()` function (which runs inside modal's instance) doesn't know what profileService is when it tries to run `this.profileService.updatePaymentMethod(this.paymentMethod, e.data)`

src/app/profile/payment-methods/payment-method/payment-method.component.js
denlight: @OzzieOrca , if I don't pass `submitted` then loading spinner won't show on submit

src/common/services/api/donations.service.spec.js
denlight: that's a good idea. thanks!

src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -0,0 +1,140 @@
...
+  loadPaymentMethods() {
+    this.profileService.getPaymentMethodsWithDonations()
+      .subscribe(
+        data => {
+          this.loading = false;
+          this.paymentMethods = data;
+        },
+        error => {
+          this.loading = false;
+          this.error = 'Failed retrieving payment methods.';
```
OzzieOrca: This doesn't go away if the session enforcer loads payment methods again. You need to clear it out on success.

Also I think it should go in a bootstrap alert.

src/common/components/paymentMethods/paymentMethodForm/paymentMethodForm.modal.tpl.html
```diff
@@ -1,19 +1,43 @@
-<div class="modal-body">
-  <div class="row">
-    <div class="col-sm-12">
-      <h3 class="text-center border-bottom-small">
-        <span ng-if="!$ctrl.resolve.paymentMethod" translate>Add New Payment Method</span>
-        <span ng-if="$ctrl.resolve.paymentMethod" translate>Edit Payment Method</span>
-      </h3>
-      <payment-method-form submitted="$ctrl.submitted" on-submit="$ctrl.onSubmit(success, data)" submission-error="$ctrl.resolve.submissionError" payment-method="$ctrl.resolve.paymentMethod" mailing-address="$ctrl.resolve.mailingAddress"></payment-method-form>
+<loading-overlay ng-if="$ctrl.submitted"></loading-overlay>
```
OzzieOrca: This somehow still isn't right. If I hit submit with an invalid form, the loading indicator takes over and submitted never gets set back to false. Let me know if you want me to take a look at it.

src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  $onDestroy(){
+    if(this.paymentMethodFormModal) {
+      this.paymentMethodFormModal.close();
+    }
+  }
+
+  $onInit(){
+    this.enforcerId = this.sessionEnforcerService([Roles.registered], {
```
Omicron7: sessionEnforcer has changed slightly with the latest PR #178. sessionEnforcer will now call the sign-in callback if the user is already signed in, meaning we don't need lines 47 - 49 anymore. sessionEnforcer will also enforce a donor session, which is required for profile pages. To enable this, add 'donor' as the third parameter to sessionEnforcerService.

See [Your Giving](https://github.com/CruGlobal/give-web/blob/master/src/app/profile/yourGiving/yourGiving.component.js#L45) for example.

denlight:
@OzzieOrca , I made it work. For some reason this.resolve.submitted is not working, but this.submissionError.loading works. So here that fix you suggested last Friday. Pushed, ready to review. @adammeyer

src/app/profile/payment-methods/payment-method/payment-method.component.js
```diff
@@ -0,0 +1,121 @@
...
+        .subscribe(() => {
+            let editedData = e.data.creditCard || e.data.bankAccount;
+            this.submissionError.loading = false;
+            this.editPaymentMethodModal.close();
+            for(let key in editedData){
+              this.model[key] = editedData[key];
+            }
+          },
+          error => {
+            this.submissionError.error = error.data;
```
OzzieOrca: I think setting `submissionError.loading` here too would be good.

PR #174, opened 2016-10-12
Desktop Nav

adammeyer:


src/common/components/nav/desktopNav.tpl.html
```diff
@@ -0,0 +1,124 @@
+<header id="desktop-navigation" class="gh-desk-nav">
+  <div class="container">
+
+    <a href="" title="Cru" class="gh-desk-logo"></a>
+
+    <ul class="gh-desk-nav-tree col-level-1">
+      <li ng-repeat="i in $ctrl.menuStructure.main" ng-class="{'has-children': i.children.length, 'open-menu': $ctrl.mobileMenuPath.main[0] === i.path.split('/').pop()}"
```
OzzieOrca: I think it is really hard to know what `i` is here. You've used it for 3 different things. The code works fine, it is just hard to read as you have to go find the nearest ng-repeat to see what it is but then `ng-repeat="i in i.children"` really doesn't tell you how deep you are. Maybe it could be `menuItemLevel1`, `itemLevel1`, `itemLvl1`, `item1` or something similar. I think sometimes I end up making things too verbose but I don't really like single character variable names. Up to you.

src/common/components/nav/nav.component.js
```diff
@@ -37,6 +45,34 @@ class NavController{
     this.getNav().subscribe((structure) => {
       this.menuStructure = structure;
     });
+
+    this.subscription = this.sessionService.sessionSubject.subscribe( () => this.sessionChanged() );
+
+    this.templateUrl = this.$window.screen && this.$window.screen.width < 991 ? mobileTemplate.name : desktopTemplate.name;
```
OzzieOrca: This doesn't seem to be working. It is always 1920 for me no matter what my chrome window size is or where I move the inspector. Also can this be done in CSS? I feel like you could put both on the template and use http://getbootstrap.com/css/#responsive-utilities-classes to show/hide them. That would also allow the page to switch menus responsively as the user resizes the page.

adammeyer: This was intended.  If you're on desktop, you get desktop, mobile you get mobile, regardless of window width.  That is why `screen` is used instead of document.

src/common/components/nav/desktopNav.tpl.html
adammeyer: I can make the variable name longer, but I kinda like having it just be re-used in the loop.  Maybe something like menuItem. Sound good?

OzzieOrca: Whatever you want. It doesn't matter much. That's fine.

src/common/components/nav/nav.component.js
OzzieOrca: Oh... Okay...

adammeyer: Yes, it's different, they liked what Amazon does.

OzzieOrca: I'd assume Amazon does it cuz they have a completely separate mobile site... and they have a min width. This desktop menu looks broken on small sizes (especially the search bar) but that's not an issue for this PR.

PR #175, opened 2016-10-12
Modal Fix or Long Names

riceguitar:
@OzzieOrca  @elikem

### DIFF
--- a/src/common/components/paymentMethods/bankAccountForm/bankAccountForm.tpl.html
+++ b/src/common/components/paymentMethods/bankAccountForm/bankAccountForm.tpl.html
@@ -19,10 +19,10 @@
         <div class="form-group form-group-default form-group-inline" ng-class="{'has-error': ($ctrl.bankPaymentForm.accountType | showErrors)}">
           <label class="label-block">Account Type</label>
           <label class="radio-inline">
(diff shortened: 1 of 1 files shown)


src/common/components/paymentMethods/bankAccountForm/bankAccountForm.tpl.html
```diff
@@ -19,10 +19,10 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Bank Account Pay
         <div class="form-group form-group-default form-group-inline" ng-class="{'has-error': ($ctrl.bankPaymentForm.accountType | showErrors)}">
           <label class="label-block">Account Type</label>
           <label class="radio-inline">
-            <input type="radio" name="accountType" ng-model="$ctrl.bankPayment.accountType" value="Checking" required> Checking <span class="hidden-md">Account</span>
+            <input type="radio" name="accountType" ng-model="$ctrl.bankPayment.accountType" value="Checking" checked required> Checking <span class="hidden-md hide-modal">Account</span>
```
OzzieOrca: Can you get rid of `checked`? Angular needs the value in its model.
You can change https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/bankAccountForm/bankAccountForm.component.js#L26 to `accountType: 'Checking'` if you want to make that change right now.

PR #176, opened 2016-10-13
Allow payment method components to edit payment methods

OzzieOrca:
- Rename addNewPaymentMethod components to paymentMethodForm
- Move mailingAddress loading outside of payment method components
- Move submission errors into paymentMethodForm
- Reduce card expiration years to only show 10 years
- Have editRecurringGiftsModal refresh payment methods after step 1
- Default bankAccountForm to Checking


src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -61,20 +61,21 @@ class ExistingPaymentMethodsController {
...
         submissionError: this.submissionError,
+        mailingAddress: this.mailingAddress,
         onSubmit: () => (params) => {
           params.stayOnStep = true;
           this.onSubmit(params);
         }
       }
     });
-    this.addNewPaymentMethodModal.result.then(null, () => {
+    this.paymentMethodFormModal.result.then(null, () => {
```
Omicron7: Instead on then(null, Function), you can use catch(Function) to drop the need for the initial null. Internally I think catch calls then with a null. Catch might be a little cleaner.

PR #177, opened 2016-10-13
ReceiptsComponent

denlight:
@adammeyer , @OzzieOrca , @Omicron7 here is the PR for the receipts. Feel free to review and provide feedback. Thanks, guys!


src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,99 @@
...
+
+    let data = {
+      'end-date': {
+
+      },
+      'start-date': {
+        'display-value': currentYear - 10 + '-' + ((today.getMonth()*1)+1) + '-' + today.getDate()
+      }
+    };
+    this.donationsService.getReceipts(data)
```
Omicron7: How long does fetching the last 10 years take? This may need to be refactored to fetch a year at a time, or as the user selects a year from the UI. Some users may have upwards of a few hundred receipts a year.

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,99 @@
...
+    this.years = [];
+    this.showYear = '';
+    this.maxShow = this.step = 25;
+    this.retrievingError = '';
+  }
+
+  $onInit() {
+    this.enforcerId = this.sessionEnforcerService([Roles.registered], {
+      'sign-in': () => {
+        this.$window.location = '/receipts.html';
```
OzzieOrca: Could we just call `getReceipts` to reload the receipts instead of reloading the whole page?

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,99 @@
...
+      },
+      cancel: () => {
+        this.$window.location = 'cart.html';
```
OzzieOrca: @Omicron7 Idk if you know but where should we be going back to? `/` or a sign in page or something?

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,99 @@
...
+    this.loading = true;
+    let today = new Date();
+    let currentYear = today.getFullYear()*1;
+
+    let data = {
+      'end-date': {
+
+      },
+      'start-date': {
+        'display-value': currentYear - 10 + '-' + ((today.getMonth()*1)+1) + '-' + today.getDate()
```
OzzieOrca: What's the `*1` for? Is that to convert it to a number? I think it is already an integer.

src/app/profile/receipts/receipts.tpl.html
```diff
@@ -0,0 +1,80 @@
...
+                <div class="form-group">
+                  <span class="pr-">View My Gifts:</span>
+                </div>
+                <div class="input-group">
+                  <div class="btn-group" uib-dropdown>
+                    <button ng-click="$ctrl.toggle = !$ctrl.toggle" class="btn btn-default dropdown-toggle" uib-dropdown-toggle>
+                      <span class="hidden-xs">{{$ctrl.showYear}} <span class="caret"></span></span>
+                      <span class="visible-xs"><span class="fa fa-filter"></span></span>
+                    </button>
+                    <ul ng-click="$ctrl.toggle = false" class="dropdown-menu" uib-dropdown-menu>
```
Omicron7: This should be a list of the last 10 years and have $ctrl.setYear() drive the loading of the receipts from the API for that year.

src/app/profile/receipts/receipts.filter.js
```diff
@@ -0,0 +1,20 @@
...
+
+let filterName = 'filterByYear';
+
+function Capitalize(){
+  return (receipts,showYear) => {
+    if(!showYear) return receipts;
+    let filteredReceipts = filter(receipts, (item) => {
+      return item['transaction-date']['display-value'].indexOf(showYear) != -1;
+    });
+    return(filteredReceipts);
```
OzzieOrca: This shouldn't need parenthesis

src/common/services/api/donations.service.js
```diff
@@ -44,10 +44,27 @@ function DonationsService( cortexApiService ) {
...
+  function getReceipts( data ) {
+    return cortexApiService
+      .post( {
+        path: '/receipts/items',
+        followLocation: true,
+        data: data
+      } )
+      .map( (response) => {
+        angular.forEach(response['receipt-summaries'], (item, index) => {
+          item['pdf-link'] = response['links'][index+1];
```
OzzieOrca: Relying on this index seems like it could be a little brittle... If another link gets added it'll be off by one. I guess the elements should be in the same order but idk if that is guaranteed.

It might be better to create that link with the `transaction-number`. Idk. What do you think? I know you asked about it at one point.

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+    this.donationsService.getReceipts(data)
+      .subscribe(
+        data => {
+          angular.forEach(data,(item) => {
+            let year = item['transaction-date']['display-value'].split('-').shift();
+            this.years.push(year*1);
+          });
+          this.receipts = data;
+          this.years = sortBy(uniq(this.years), (item) => {
```
OzzieOrca: I'm not sure how the receipt data is structured. I think uniq keeps only the first occurrence of each element. Is that what you want? How did we get duplicates? Seems like we might be throwing good data away. You could use groupBy if you want to keep them all.

src/app/profile/receipts/receipts.component.js
adammeyer: Yes, it should be the home page.  `/` should be fine.

adammeyer: It's fine that `/` redirects to the cart on local dev.  It won't on AEM.

denlight:
@adammeyer , @Omicron7 , @OzzieOrca The updated code is here. Please review and provide feedback if needed. Thanks!

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,106 @@
...
+    this.currentYear = this.today.getFullYear();
+    this.getReceipts(this.currentYear,true);
+  }
+
+  getReceipts(year, tryPreviousYear){
+    this.loading = true;
+
+    let data = {
+      'end-date': {
+        'display-value': year + '-' + (this.today.getMonth()+1) + '-' + this.today.getDate()
```
Omicron7: This should probably be today's date or {year}-12-31 depending if year is current year or previous.

denlight: good catch, @Omicron7 !

denlight:
@Omicron7 , @adammeyer last fix is done

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,109 @@
...
+        data => {
+          this.retrievingError = '';
+          this.receipts = data;
+          //if there are no receipts in the current year try previous year
+          if(this.receipts.length == 0 && tryPreviousYear) {
+            this.currentYear = this.currentYear - 1;
+            this.getReceipts(this.currentYear);
+            return;
+          }
+          // if no receipts were found for the past two years, set current year back to 2016
```
OzzieOrca: This comment is going to get outdated...

src/app/profile/receipts/receipts.component.js
```diff
@@ -0,0 +1,109 @@
...
+        this.$window.location = '/';
+      }
+    }, EnforcerModes.donor);
+    this.today = new Date();
+    this.currentYear = this.today.getFullYear();
+    this.getReceipts(this.currentYear,true);
+  }
+
+  getReceipts(year, tryPreviousYear){
+    this.loading = true;
```
OzzieOrca: I think you should put `this.retrievingError = '';` up here so the error will go away after signing in and not still be there while it is loading.

denlight:
@OzzieOrca , @Omicron7 , @adammeyer updated code was pushed.

PR #178, opened 2016-10-13
Pre checkout register account

Omicron7:
This PR adds the pre-checkout donor matching workflow. It also adds donor (registration-state) enforcement to the sessionEnforcer.

Profile pages, like your-giving, have a requirement that you must be signed in as well as have a completed donorDetails registration-state before you can access them. SessionEnforcer can now enforce this as well as drive up the required matching process if you are not completed.

I'm thinking through ways of refactoring the registerAccountService because of the complexity of the workflow. I believe it also needs at least one more POST to donorMatchesForm before it will handle all the possible match scenarios.


src/common/components/userMatchModal/userMatchModal.component.spec.js
```diff
@@ -25,7 +25,7 @@ describe( 'userMatchModal', function () {
 
   describe( '$onInit', () => {
     beforeEach( () => {
-      spyOn( $ctrl.verificationService, 'getContacts' ).and.returnValue( Observable.of( [] ) );
+      spyOn( $ctrl.verificationService, 'getContacts' ).and.callFake( () => Observable.of( [] ) );
```
OzzieOrca: What's the motivation for this switch to `callFake`? I've been slowly moving things to use `returnValue` cuz it doesn't need to be wrapped in a function and therefore is shorter. I didn't realize it would work when I first started writing these tests. It's trivial, I'm just curious.

src/common/services/registerAccount.service.js
```diff
@@ -0,0 +1,86 @@
...
+import sessionService from 'common/services/session/session.service';
+
+import {Roles} from 'common/services/session/session.service';
+
+let serviceName = 'registerAccountService';
+
+/*@ngInject*/
+function RegisterAccountService( $q, orderService, sessionService, sessionModalService ) {
+
+// angular.element(document.querySelector('body')).injector().get('registerAccountService')().then(function() {console.log('success');}, function() {console.log('failure')});
```
OzzieOrca: You want to add a comment to indicate this code is for manually opening the modal? (I assume that's what it is anyways.) Just so it doesn't look like commented out code that someone might remove and so people know how to use it in the future.

src/common/components/userMatchModal/userMatchModal.component.spec.js
Omicron7: I've used returnValue when returning a primitive or immutable object. returnValue always returns the same object on multiple calls, meaning if the spy was called twice, return values would be equal(===). callFake will execute the method each time, meaning it will return different objects each time or allow return value to react to parameters passed to the spy.

In this case, I think they are equivalent.

src/common/services/registerAccount.service.js
Omicron7: I just had it in there so I'd remember it, since chrome incognito doesn't keep it in history. I'm going to remove it before final PR.

src/common/components/userMatchModal/userMatchModal.component.spec.js
Omicron7: I know why I changed this. I was testing out passing back a variable, then changing it slightly in each it block. Didn't work though.

Omicron7:
I had an idea to simplify this which I tried this morning but ran into other problems. This looks like the best way to handle the registerAccount flow for right now.

PR #181, opened 2016-10-17
[#EP-938] Stop / Restart gift step 0

Omicron7:
This is the initial Step 0 for Stopping or Restarting a gift modal.

### DIFF
--- a/src/app/profile/yourGiving/stopStartRecurringGifts/step0/stopStartStep0.tpl.html
+++ b/src/app/profile/yourGiving/stopStartRecurringGifts/step0/stopStartStep0.tpl.html
@@ -1,80 +1,78 @@
-<div class="modal-content give-modal-content">
-  <div class="modal-header">
-    <div class="container-fluid">
-      <div class="row">
-        <div class="col-xs-12">
-          <div class="border-bottom-small">
-            <button type="button" class="close" data-dismiss="modal" aria-label="Close">
-              <span aria-hidden="true">×</span></button>
-            <h3 translate>What would you like to do?</h3>
+<div class="modal-header">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-xs-12">
+        <div class="border-bottom-small">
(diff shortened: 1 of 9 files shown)


src/app/profile/yourGiving/stopStartRecurringGifts/step0/stopStartStep0.tpl.html
```diff
@@ -0,0 +1,80 @@
+<div class="modal-content give-modal-content">
```
OzzieOrca: I think https://github.com/CruGlobal/give-web/blob/master/src/common/templates/giveModalWindow.tpl.html will provide this div. `<div class="modal-content give-modal-content" uib-modal-transclude></div>`

src/app/profile/yourGiving/stopStartRecurringGifts/step0/stopStartStep0.tpl.html
```diff
@@ -0,0 +1,80 @@
...
+              <span aria-hidden="true">×</span></button>
+            <h3 translate>What would you like to do?</h3>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-body">
+    <div class="container-fluid" tabindex="-1" role="dialog">
```
OzzieOrca: Do we need `tabindex="-1" role="dialog"`? I've been removing them. I think `$uibModal` sets these on a div that is higher up the DOM. Idk why the UX guys have it only on the body. I could be wrong...

Omicron7: I think UX team is pulling from core bootstrap for modal markup, which includes those. $uibModal does things slightly different, so you are correct, we don't need em.

src/app/profile/yourGiving/stopStartRecurringGifts/step0/stopStartStep0.tpl.html
Omicron7: You're correct, this was a copy/paste error.

PR #182, opened 2016-10-18
Refactor gift date operations out into giftHelpers/giftDates.service

OzzieOrca:
- Update productConfigModal and cart service to use these new functions

### DIFF
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -5,28 +5,26 @@ import 'rxjs/add/observable/forkJoin';
 import 'rxjs/add/operator/map';
 
 import cortexApiService from '../cortexApi.service';
+import giftDatesService from '../giftHelpers/giftDates.service';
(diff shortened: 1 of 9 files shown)


src/common/services/api/cart.service.js
```diff
@@ -117,54 +115,25 @@ function cart(cortexApiService){
...
 
-    return cortexApiService.post({
-      path: ['itemfieldslineitem', 'items', cortexApiService.scope, id],
+    return this.cortexApiService.post({
+      path: ['itemfieldslineitem', 'items', this.cortexApiService.scope, id],
       data: data
     });
   }
 
-  function updateItem(uri){
```
adammeyer: That sounds good.  Where you able to confirm that an update isn't possible?

OzzieOrca: I'll ask now. I forgot about that.

PR #183, opened 2016-10-18
Edit recurring gifts step 1

OzzieOrca:
@brian This isn't done but I wanted you to at least see it before you get too far on your modals as so many of the modal views are similar.

Mainly I wanted to get your thoughts on the `giftListItem` component and it's usage. I think it or some variant could be used for most, if not all of the modal views that display gifts. I think at the very least they all share the profile picture and name and then have some sort of body or select action. I might be trying to cram too much functionality in there but they all should take in a gift object and either just display it in full or abbreviated form, return a modified version, or a selected status. Maybe we should split it into a display, edit, and select component... Idk.

Let me know what you think and how I could clean it up or make it better. Feel free to suggest even small things like names. Maybe you've already got something started or some ideas... I'd love to hear them.

It would probably be easier to just look at the second commit as this is based off of https://github.com/CruGlobal/give-web/pull/182.


src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
```diff
@@ -0,0 +1,66 @@
...
+        this.loading = false;
+      });
+  }
+
+  processChanges() {
+    let recurringGiftChanges = filter(this.recurringGifts, { dirty: true });
+
+    map(recurringGiftChanges, (gift) => {
+      if(gift['updated-rate']['recurrence']['interval'] !== 'Monthly'){
+        //TODO: figure out start date to send. This could be done in giftListItem (and might need to be for displaying there) but that probably requires a watcher or ngChange.
```
wrandall22: Please let us know, or collaborate with @dsgoers about this field.

Omicron7:
@OzzieOrca I agree that we should have a common giftListItem component. Looking through the UX site, it looks like it can be configured in 4 different ways.

1. VIEW / CONFIRM - view only version, used to confirm changes
screen shot 2016-10-18 at 10 04 28 am
2. EDIT / CONFIGURE - Allows full editing of the gift
screen shot 2016-10-18 at 10 04 57 am
3. ONE-TIME - edit / configure one-time gift 
screen shot 2016-10-18 at 10 05 36 am
4. PAYMENT - Only allows changing the payment method (dropdown has option to add new payment method). 
screen shot 2016-10-18 at 10 06 00 am

It might be overkill to build all of this in a single component. My suggestion would be a component for the picture / name, and transcluded components for the different editing modes:

```javascript
angular.module('name').component({
    controller: Function,
    templateUrl: template.name,
    transclude: true
});
```
```html
<div>Template</div>
<div>
    <div ng-transclude />
</div>
```
This allows insertion of another component/directive into a specific place in the component. Allows for easy customization of sections of a larger component, determined by the parent, and not the child component.

```html
<gift-list-item gift="$ctrl.item">
    <gift-list-view gift="$ctrl.item" />
</gift-list-item>
````
```html
<gift-list-item gift="$ctrl.item">
    <gift-list-edit gift="$ctrl.item" />
</gift-list-item>
````
We could have a single component for the different edit modes if they share enough common code, or a component for each.
```html
<gift-list-item gift="$ctrl.item">
    <gift-list-one-time gift="$ctrl.item" />
    <!-- <gift-list-edit gift="$ctrl.item" mode="'one-time'" /> -->
</gift-list-item>
````

Thoughts? Did I miss any giftListItem configurations? Not sure if this is also used in the profile pages anywhere.

**Comment by OzzieOrca** [comment]
This has been updated to use Brian's giftListItem component.

@dsgoers Can you review all my functions that end in `*Model` starting at https://github.com/CruGlobal/give-web/pull/183/files#diff-7e168fa0e66ed494b94e15748d08a762R25?. You can also look through the corresponding tests https://github.com/CruGlobal/give-web/pull/183/files#diff-824983ecfbf592091178f493c3c083b7R37. Reading the test descriptions might help a little bit. Idk.

Basically those functions are called with no parameters when the template needs to read one of the values and it returns the value based on the `updated-*` value if it exists or uses the original, unmodified value.

When the template needs to update a value, it calls one of those functions with the `value` parameter set and it sets the `updated-*` value unless the new value is the same as the old value or special conditions are met.

Let me know if you see any issues with the `updated-start-month/year` and when I'm initializing them (defaulting them based off the user input and the nextDrawDate) or clearing them. Let me know if you have any questions or catch anything. I realized I still have this comment/question https://github.com/CruGlobal/give-web/pull/183/files#diff-824983ecfbf592091178f493c3c083b7R225. Idk if you have any thoughts about that. I could probably describe it better...

Thanks :)

File: src/common/components/giftViews/giftUpdate/giftUpdate.component.js [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+import giftDatesService from 'common/services/giftHelpers/giftDates.service';
+
+import template from './giftUpdate.tpl';
+
+let componentName = 'giftUpdate';
+
+class GiftUpdateController {
+
+  /* @ngInject */
+  constructor($scope, giftDatesService) {
```
Omicron7: It doesn't look like $scope is used here or in the template.

src/app/profile/yourGiving/editRecurringGifts/editRecurringGifts.modal.tpl.html
```diff
@@ -105,5 +105,6 @@
     dismiss="$ctrl.dismiss()"
     previous="$ctrl.previous()"
     next="$ctrl.next()">
+    <pre>{{$ctrl.recurringGiftChanges | json }}</pre>
```
Omicron7: Did you want to commit this?

PR #188, opened 2016-10-19
Make main nav responsive

adammeyer:


adammeyer:
Thanks @OzzieOrca, I'll add that those notes to the nav css bugs jira task.

PR #189, opened 2016-10-20
up Karma timeout

adammeyer:
So tests pass


adammeyer:
My other idea was to remove tests...thought this might be safer. 😄

OzzieOrca:
Hahaha

PR #190, opened 2016-10-20
[#EP-933] Stop recurring gifts step 1

Omicron7:
This includes step-1 of the stop gift action as well as the groundwork for the giftListItem component.

### DIFF
--- a/src/common/components/giftViews/giftListItem/giftListItem.component.js
+++ b/src/common/components/giftViews/giftListItem/giftListItem.component.js
@@ -1,0 +1,31 @@
+import angular from 'angular';
+import template from './giftListItem.tpl';
+import desigSrc from 'common/directives/desigSrc.directive';
+
+let componentName = 'giftListItem';
+
+class GiftListItemController {
+
+  /* @ngInject */
+  constructor() {
+  }
+}
+
+export default angular
+  .module( componentName, [
+    template.name,
+    desigSrc.name
+  ] )
+  .component( componentName, {
+    controller:  GiftListItemController,
+    templateUrl: template.name,
+    transclude:  {
+      'selectInput': '?label'
+    },
+    bindings:    {
+      gift:        '<',
+      selectable:  '@',
+      selectLabel: '@',
+      onSelected:  '&'
+    }
+  } );
--- a/src/common/components/giftViews/giftListItem/giftListItem.tpl.html
+++ b/src/common/components/giftViews/giftListItem/giftListItem.tpl.html
@@ -14,13 +14,13 @@
     <label ng-switch-when="checkbox">{{$ctrl.selectLabel}}
       <input type="checkbox"
              name="selectedGift"
-             ng-model="$ctrl.gift.$$selectedGift"
-             ng-change="$ctrl.onSelected({gift: $ctrl.gift, selected: $ctrl.gift.$$selectedGift})">
+             ng-model="$ctrl.gift._selectedGift"
+             ng-change="$ctrl.onSelected({gift: $ctrl.gift, selected: $ctrl.gift._selectedGift})">
     </label>
     <label ng-switch-when="radio">{{$ctrl.selectLabel}}
       <input type="radio"
              name="selectedGift"
-             ng-model="$ctrl.gift.$$selectedGift"
+             ng-model="$ctrl.gift._selectedGift"
              ng-value="true"
              ng-change="$ctrl.onSelected({gift: $ctrl.gift, selected: true})">
     </label>
--- a/src/common/components/giftViews/giftListItem/giftListItem.tpl.html
+++ b/src/common/components/giftViews/giftListItem/giftListItem.tpl.html
@@ -1,0 +1,28 @@
+<div class="col-xs-12 col-sm-3 mb">
+  <img desig-src="{{$ctrl.gift['designation-number']}}">
+</div>
+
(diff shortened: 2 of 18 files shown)


src/common/components/giftViews/giftSimpleView/giftSimpleView.tpl.html
```diff
@@ -0,0 +1,5 @@
+<h5 class="mt-- mb0">{{$ctrl.gift['amount'] | currency}} {{$ctrl.gift.rate.recurrence.interval}}</h5>
+<div ng-switch="$ctrl.gift.rate.recurrence.interval">
+  <small ng-switch-when="Monthly" translate>On the {{$ctrl.gift['recurring-day-of-month'] | ordinal}} day of each month</small>
+  <!-- TODO: Other intervals -->
```
Omicron7: I left these blank for now since I believe @OzzieOrca is working on helper services to generate these.

src/common/components/giftViews/giftListItem/giftListItem.component.js
```diff
@@ -0,0 +1,31 @@
...
+    desigSrc.name
+  ] )
+  .component( componentName, {
+    controller:  GiftListItemController,
+    templateUrl: template.name,
+    transclude:  {
+      'selectInput': '?label'
+    },
+    bindings:    {
+      gift:        '<',
```
OzzieOrca: How does selectedGift get out of this component if this is one-way binding? I guess it updating the object that is passed by reference right?

src/common/components/giftViews/giftListItem/giftListItem.tpl.html
```diff
@@ -0,0 +1,28 @@
...
+  <ng-transclude />
+</div>
+
+<div ng-if="$ctrl.selectable"
+     class="col-xs-5 col-sm-3">
+  <div class="checkbox-right" ng-transclude="selectInput" ng-switch="$ctrl.selectable">
+    <label ng-switch-when="checkbox">{{$ctrl.selectLabel}}
+      <input type="checkbox"
+             name="selectedGift"
+             ng-model="$ctrl.gift.$$selectedGift"
```
OzzieOrca: There are a few places in this file that still have `$$selectedGift`. Are these supposed to be `_selectedGift` now?

src/common/components/giftViews/giftListItem/giftListItem.tpl.html
```diff
@@ -0,0 +1,28 @@
...
+             ng-change="$ctrl.onSelected({gift: $ctrl.gift, selected: $ctrl.gift.$$selectedGift})">
```
OzzieOrca: What are these for? I guess I'm not immediately seeing how data is getting out of this component. Is it by updating the gift object or by calling onSelected? It doesn't look like the onSelected binding is hooked up here https://github.com/CruGlobal/give-web/pull/190/files#diff-119da8a34d16697efc4a9919ad4bc57eR24.

Omicron7: This is if the parent component wants to listen to the checkbox/radio changes. I'm not using it in stop step 1.

OzzieOrca: Gotcha. Ok

src/common/components/giftViews/giftListItem/giftListItem.component.js
Omicron7: [Right here](https://github.com/CruGlobal/give-web/blob/stop-gift-step-1/src/app/profile/yourGiving/stopStartRecurringGifts/stopGift/stopGiftStep1.component.js#L18). This function is called by the stopGiftStep1 "Continue" button and is passing back to the  parent stopGift component the selected gifts from the list. It's somewhat open how the developer or component wants to get the selected gifts back. The gift objects are modified in place, so a component could just call next() and the parent can find selected ones by filtering.

OzzieOrca: If angular was really only doing one-way binding, a gift that had _selected added to it in giftListItem should never make it back out into the array of gifts right? I'm guessing it just isn't watching on the scope of the parent but that variable is updated when you specifically read it in the filter in the file you linked. Just seems a little weird.

OzzieOrca: Is https://github.com/CruGlobal/give-web/pull/190/files#diff-119da8a34d16697efc4a9919ad4bc57eR49 updating correctly? Seems like it might not always run that when gift is updated since it isn't watching it. Idk

Omicron7: The object is bound by reference. If you set `this.gifts = ['a']` in the component, the parent wouldn't see this since it's a new object. The parent would see this `this.gifts.push('a')` though, since it's modifying the existing object and they are bound by reference. This is the same for object properties. Angular isn't copying the object when it's passed to the component.

The difference between one-way, and two-way, is that two-way, the parent would see `this.gifts = ['a']'.

See [component-based-application-architecture](https://docs.angularjs.org/guide/component#component-based-application-architecture), the section starting with "Inputs should be using "

Omicron7: As to the $ctrl.find, yes, it updates it correctly. Angular calls a bound function each digest cycle.

PR #192, opened 2016-10-20
move css to repo

adammeyer:


OzzieOrca:
@riceguitar @j2trumpet Can these be removed now? https://github.com/CruGlobal/give-web/blob/2d7a2060738f9e7fdf45d9305ccadfc16d001de7/src/assets/scss/styles.scss#L32-L38

PR #193, opened 2016-10-20
Add more date helper functions to giftDates and use moment.js

OzzieOrca:
- Add moment
- Use moment.months() for generating list of months
- Add quarterlyMonths function to giftDates for generating a list of months where a quarterly gift will be given
- Use quarterlyMonths in delete payment method modal
- Update startDate and add startMonth functions to giftDates

@Omicron7 I decided to split this out since it is ready and would be easier to review. I might need to make some changes to the functions eventually but I think they should be good for now.


src/app/cart/cart.tpl.html
```diff
@@ -32,7 +32,7 @@ <h3 class="panel-name" translate>Your Gift Cart</h3>
                         <span class="giftsum-frequency giftsum-title">{{i.frequency}}</span>
                         <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'">
                           <span translate>Starts on:</span>
-                          {{i.giftStartDate | date:'mediumDate'}}
+                          {{i.giftStartDate.format('LL')}}
```
Omicron7: I believe this should be 'll'.
From Angular Docs
> 'longDate': equivalent to 'MMMM d, y' for en_US locale (e.g. September 3, 2010)
> 'mediumDate': equivalent to 'MMM d, y' for en_US locale (e.g. Sep 3, 2010)

From Moment Docs
> 'LL' September 4 1986
> 'll' Sep 4 1986

At some point, possibly post code-complete, we should go back and normalize dates across the application.

src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -141,7 +141,7 @@
           <td class="td-frequency">
             <span class="giftsum-frequency giftsum-title">{{i.frequency}}</span>
             <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'" translate>
-              Starts on: {{i.giftStartDate | date:'MMM d, yyyy'}}
+              Starts on: {{i.giftStartDate.format('LL')}}
```
Omicron7: Same here 'll' for the month abbreviation

PR #194, opened 2016-10-20
On travis, skip gulp release on branches other than master or dev

OzzieOrca:
What do you guys think about this? The results of gulp release are thrown away unless a deploy happens. I think this would speed up building for PRs as gulp release takes over a minute on its own.

The only issue I can think of is if we want to make sure something doesn't break during release before building master but I don't think I've ever seen it break. If that did happen, the build on master would fail and it wouldn't deploy anyways.

Feel free to discuss. I'd be fine leaving it as is if there are concerns but it seemed like a decent idea.


OzzieOrca:
I've tested this using the `dev` branch. It runs release on `dev` but not on travis-ci/pr or travis-ci/push to a non `master` or `dev` branch.

PR #196, opened 2016-10-21
Add update recurring gift method

Omicron7:
This method is used to update recurring gifts. See https://wiki.cru.org/display/EP/Self-Service+API#Self-ServiceAPI-Response.6

### DIFF
--- a/src/common/services/api/donations.service.js
+++ b/src/common/services/api/donations.service.js
@@ -1,6 +1,7 @@
 import angular from 'angular';
 import map from 'lodash/map';
 import flatMap from 'lodash/flatMap';
+import omit from 'lodash/omit';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/forkJoin';
 import 'rxjs/add/operator/pluck';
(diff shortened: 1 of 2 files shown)


src/common/services/api/donations.service.js
```diff
@@ -54,54 +55,63 @@ function DonationsService( cortexApiService, profileService ) {
...
+          } );
+        } );
+      } );
+  }
+
+  function updateRecurringGifts( gifts ) {
+    gifts = angular.isArray( gifts ) ? gifts : [gifts];
+    return cortexApiService.put( {
+      path: ['donations', 'recurring', cortexApiService.scope, 'active'],
+      data: {donations: map( gifts, ( gift ) => omit( gift, ['_selectedGift', 'paymentMethod'] ) )}
```
OzzieOrca: Should we rename paymentMethod to have an underscore? Guess it doesn't matter too much. I also haven't figured out what I want to do when the user changes the payment method id on a gift. Do you think I should update the paymentMethod object when the user changes that or just leave it and let other views that know it could be changed fetch the new one? I guess that could become a method in our new model.

Omicron7: I'd leave it as is for now. When we convert the gifts to a model, it can handle omitting the properties. I'd expect paymentMethod to be a getter and work the same as the other getters. If payment-method-id is updated, paymentMethod should return the new updated payment information.

PR #214, opened 2016-10-25
Bugs EP-816 + EP-1425

denlight:
@adammeyer, @OzzieOrca Please review and feel free to provide feedback. Thanks!


src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -21,11 +21,19 @@ <h4 class="modal-title" translate>
...
+
 <form novalidate name="$ctrl.itemConfigForm" ng-submit="$ctrl.addToCart()" ng-if="!$ctrl.giftSubmitted">
   <div class="modal-body pt0 p">
     <div class="container-fluid">
       <div class="panel panel-default panel-plain mt">
         <div class="panel panel-default give-modal-panel">
+
+          <div ng-if="$ctrl.error" ng-switch="$ctrl.error.indexOf('already in the cart') !== -1" class="alert" role="alert"
+               ng-class="{'alert-warning': $ctrl.error.indexOf('already in the cart') !== -1, 'alert-danger': $ctrl.error.indexOf('already in the cart') === -1}">
+            <span ng-switch-when="true" translate>This gift has already been added. Please proceed to checkout.</span>
```
OzzieOrca: I might add something about you can edit it in the cart (maybe their intentions were to give a different amount/frequency than however it got added to the cart previously). Maybe someone gave you specific requirements, that was just my thought.

src/app/productConfig/productConfig.modal.js
```diff
@@ -49,13 +49,29 @@ class ModalInstanceCtrl {
...
+      this.customInputActive = true;
     }
 
     if(!this.itemConfig['recurring-day-of-month'] && nextDrawDate) {
       this.itemConfig['recurring-day-of-month'] = Number(nextDrawDate.split('-')[2]).toString();
     }
   }
 
+  $onInit() {
+    this.handlePatternValidation = (() => {
```
OzzieOrca: This seems like a little bit of a hack on `ng-pattern`. I guess it's fine as it's pretty clear what you are doing and I don't think this will break.

You could consider doing something like [this](https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L73-L102) but it is a bit of code as you have to make sure the form is defined before adding something to the `$validators` array of the input. https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#$validators

I'm fine with you leaving it though.

src/app/productConfig/productConfigModal.tpl.html
denlight: Just received some instructions from Jason based on your comment

src/app/productConfig/productConfigModal.tpl.html
```diff
@@ -42,26 +50,33 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                         required>
                 </div>
               </label>
               <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error" ng-if="($ctrl.itemConfigForm.amount | showErrors)">
-                <div class="help-block" ng-message="min" translate>Amount must be at least $1.00</div>
-                <div class="help-block" ng-message="max" translate>Amount must be less than $10,000,000.00</div>
-                <div class="help-block" ng-message="pattern" translate>Amount must be a valid number with no more than 2 decimal places</div>
+                <div class="help-block" ng-message="pattern" translate>Amount must be a valid number</div>
+                <div class="help-block" ng-message="required" translate>Amount must be not empty</div>
+                <div class="help-block" ng-message="minimum" translate>Amount must be more than $1.00</div>
```
OzzieOrca: `Amount must be at least $1.00`? The greater than equal cases are always weird to phrase. Also should translation handle the localization of this static number or should we use https://docs.angularjs.org/api/ng/filter/currency?

twinge: Is the backend enforcing > 1.0, or >= 1.0? It's definitely meant to be >= 1.0 with instructional text of `Amount must be at least $1.00`

denlight: @OzzieOrca , @twinge , good catch!

PR #215, opened 2016-10-25
[#EP-936] Redirect a gift Step 2

Omicron7:


src/app/profile/yourGiving/stopStartRecurringGifts/redirectGift/step2/redirectGiftStep2.component.js
```diff
@@ -0,0 +1,80 @@
...
+
+class RedirectGiftStep2Controller {
+
+  /* @ngInject */
+  constructor( designationsService ) {
+    this.designationsService = designationsService;
+    this.find = find;
+
+    // Create debounced search method
+    this.performSearch = debounce( this._performSearch, 750 );
```
OzzieOrca: What do you think about using Rxjs for this? I guess you've got it working with the unsubscribe. Up to you. I think it could be slightly condensed but there's not really a need to rewrite it and update tests since you've got it running.

Here's how I did something similar:
https://github.com/CruGlobal/mpdx/blob/master/app/assets/javascripts/angular/components/contacts/contactList/contactList.js#L109 Search for `filterChangeEventEmitter`. It's using a subject, debounceTime, and switchMap. Here you could probably pass the input value through to switchMap on change instead of just placeholder strings.

PR #218, opened 2016-10-26
Pads possibleTransactionDays to match recurringGift responses.

Omicron7:
I was noticing that transactionDay's from the API are padded with the leading zero for single digit days. This was causing `<select>` to not match. This change pads the single digits while converting it to a string.

### DIFF
--- a/src/common/services/giftHelpers/giftDates.service.js
+++ b/src/common/services/giftHelpers/giftDates.service.js
@@ -1,10 +1,9 @@
 import moment from 'moment';
 import range from 'lodash/range';
 import map from 'lodash/map';
-import toString from 'lodash/toString';
 
 export function possibleTransactionDays() {
-  return range( 1, 29 ).map( toString );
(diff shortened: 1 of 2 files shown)


Omicron7:
I also tested this with the ordinal filter, and it works correctly.
`{{'04' | ordinal}}  == '4th'`

src/common/services/giftHelpers/giftDates.service.js
```diff
@@ -1,10 +1,9 @@
 import moment from 'moment';
 import range from 'lodash/range';
 import map from 'lodash/map';
-import toString from 'lodash/toString';
 
 export function possibleTransactionDays() {
-  return range( 1, 29 ).map( toString );
+  return range( 1, 29 ).map((i) => (`0${i}`).slice(-2));
```
OzzieOrca: https://lodash.com/docs/4.16.4#padStart would work too. This is good too.

PR #222, opened 2016-10-27
Designation Page Editor

adammeyer:
Still to do:
- verify "Save" works. (endpoint still under development)
- Improve photo uploader (endpoint still under development)

@OzzieOrca @Omicron7 Let me know if you have any initial feedback. Thanks!


src/app/designationEditor/designationEditor.component.js
```diff
@@ -0,0 +1,230 @@
...
+class DesignationEditorController {
+
+  /* @ngInject */
+  constructor( $http, $q, $uibModal, $location, $window, envService, sessionService, sessionEnforcerService ) {
+    this.sessionService = sessionService;
+    this.sessionEnforcerService = sessionEnforcerService;
+    this.imgDomain = envService.read('imgDomain');
+    this.imgDomainDesignation = envService.read('imgDomainDesignation');
+    this.$location = $location;
+    this.$http = $http;
```
Omicron7: I'd suggest whatever is using $http get moved to an api service and out of the controller.

src/app/designationEditor/pageOptionsModal/pageOptionsModal.tpl.html
```diff
@@ -0,0 +1,81 @@
...
+          <tr>
+            <th>Amount</th>
+            <th>Title (Description)</th>
+            <th></th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr ng-repeat="amount in $ctrl.suggestedAmounts">
+            <td><input type="number" class="form-control" placeholder="50" min="0"></td>
+            <td><input type="text" class="form-control" placeholder="$50.53 to help 14 refugees"></td>
```
OzzieOrca: Is this supposed to be static?

src/app/designationEditor/designationEditor.component.js
```diff
@@ -0,0 +1,220 @@
...
+      })
+    ]).then(() => {
+      this.loadingOverlay = false;
+    });
+  }
+
+  editTitle() {
+    let modalOptions = {
+      templateUrl: titleModalTemplate.name,
+      controller: titleModalController.name,
```
OzzieOrca: I like the idea of using the component syntax for all of these.  I don't feel that you need to change it if you don't want to but I think in the future it would be nice to use components. You can look at [existingPaymentMethods.component.js#L67-L79](https://github.com/CruGlobal/give-web/blob/master/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js#L67-L79) and [paymentMethodForm.modal.component.js#L32-L33](https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/paymentMethodForm/paymentMethodForm.modal.component.js#L32-L33) for an example.

src/app/designationEditor/pageOptionsModal/pageOptionsModal.tpl.html
adammeyer: No, still waiting to see what that json will look like from AEM.

adammeyer:
@OzzieOrca @Omicron7 ready for your review again.

PR #224, opened 2016-10-28
Extend getRecurringGifts to handle multiple types

Omicron7:
Some of the recurring gift modals require showing more than one type (active, cancelled, suspended, ...) of gift. Instead of creating new methods to fetch each type, I've updated the getRecurringGifts method to take an array of types.


OzzieOrca:
I just went ahead and merged this since I'm going to make some changes to the same file. Hopefully that's ok.

PR #225, opened 2016-10-28
Set up global websites modal

OzzieOrca:
You can put `ng-click="$ctrl.dismiss()"` on any element you want to close the modal. You probably need to customize `globalWebsitesModalWindow.tpl.html` to get the modal styles you need. Let me know if you run into any issues.


adammeyer:
@OzzieOrca Who was this PR addressed to?

OzzieOrca:
@j2trumpet asked for it

adammeyer:
@j2trumpet Is this good to go?

OzzieOrca:
@adammeyer I don't think he's worked on it yet. I think we was working on the CSS somewhere but it's not in here yet. It needs content.

PR #226, opened 2016-10-29
Recurring gift model static payment methods next draw date

OzzieOrca:
It might be easier to look at the 2 commits separately as the first makes it possible for nextDrawDate and paymentMethods to be stored only on the RecurringGiftModel and the second removes the old loading methods and bindings in the app for that data.

@Omicron7 Can you make sure I modified your start/stop components correctly?


Omicron7:
I'm merging this. I've got restart gift tasks that require using this.

PR #229, opened 2016-10-31
Use same code for adding and editing a gift in productConfig

OzzieOrca:
- Add update gift method to cart service
- Remove btoa conversion of uri

https://jira.cru.org/browse/EP-1448
@adammeyer Can you verify that removing `btoa` is ok? I couldn't figure out why it was there as it seems to be converted right back with `atob` and it seems to work fine without it.


adammeyer:
@OzzieOrca Removing `btoa` is fine, thanks for checking.

PR #234, opened 2016-11-02
mobile detect nav

adammeyer:
Uses UserAgent instead of window width


adammeyer:
@Omicron7 haha, we'll have to wait and see!

PR #235, opened 2016-11-02
[#EP-930][#EP-931] Restart giving

Omicron7:


src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step1/suspendedGifts/suspendedGifts.tpl.html
```diff
@@ -0,0 +1,57 @@
...
+    <div class="row row-no-spacing">
+      <div class="col-xs-12">
+        <form name="$ctrl.restartForm">
+          <div class="repeating-row select-recipient-row" ng-repeat="gift in $ctrl.gifts">
+            <gift-list-item gift="gift"
+                            selectable="checkbox">
+              <label>{{'Restart' | translate}}
+                <input type="checkbox" disabled>
+              </label>
+              <p translate>This gift cannot be changed online. For assistance,
```
OzzieOrca: So does this component/modal step do anything other than show the user what they can't do online?

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step3/confirmGifts/confirmGifts.component.js
```diff
@@ -0,0 +1,75 @@
...
+    } );
+  }
+
+  processRestarts() {
+    let requests = [];
+    this.setLoading( {loading: true} );
+    if ( !isEmpty( this.adds ) ) {
+      requests.push( this.donationsService.addRecurringGifts( this.adds ).do( () => {
+        this.saved = concat( this.saved, this.adds );
+        this.adds.length = 0;
```
OzzieOrca: Can you make sure this works if the user saves, has one set of gift succeed and the other fail, and then goes back to some other step and then comes back to confirm?  Hopefully it is an edge case but I think it would be bad if a user ended up adding gifts twice. Maybe you've already tested it or maybe it already works. In my modals, in one of the cases, going back would forget the changes done here (because it was using data in the parent component to send to this confirm page) and have the gifts added back when they came back.

An easy way I found to test it is just to comment out the `this.next()` below.

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step1/suspendedGifts/suspendedGifts.tpl.html
Omicron7: It's complicated. Step 1 can show up to 3 different modals to allow the user to restart a gift. The initial modal shows suspended gifts, which you can't restart in give currently (backend post deploy feature). If the user has not suspended gifts, you skip this modal. We then show suggested recipients, You can pick 1 or more of these to start giving gifts to (From user perspective: "I sent in a check twice last year, and now I want to 'restart' my giving online). If you have no suggested recipients, you skip it. If you have no suspended or suggested, or the user didn't select anything here, we show the Search Box. User is allowed to pick 1 (radio input) here. It then proceeds to configuration and confirmation.

I wasn't able to find a user with suspended gifts, so in development, it was always skipping to suggested.

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step3/confirmGifts/confirmGifts.component.js
Omicron7: I tested this in the specs, if either one succeeds and the other fails, it moves saved items to the saved array and sets an error state. The UI changes the "Back" button to a "Cancel" in this case and doesn't allow the user to go back. If we need to allow it, I'll need to do some additional work to un-select saved gifts. I clone the gifts before sending them to this component so that changes made to them aren't changing the originals.

Omicron7:
@OzzieOrca What payment stuff are you talking about?

PR #236, opened 2016-11-03
Take current date and a gift's existing start date into account when …

OzzieOrca:
…determining when a gift should start

- Add _earliestValidDate function to determine the greatest of nextDrawDate, startDate, and the current date
- Add moment timezone
- Remove explicit systemjs dependency as jspm includes the version it wants
- Update some packages

@Omicron7 @adammeyer @wrandall22 This still needs some tests, but could you look over it? I added moment-timezone but it seems like a little overkill just to convert to Florida server time (or AWS East, wherever the server is). It does need to take into account DST though... Maybe I could figure out how to import just the America/New_York timezone.

I'm still a little unclear on how we want to handle dates. No one has expressed that having dates off by a day for users in other timezones matters (except for sometimes/somewhere using Pacific Time to handle end of the year gifts that are given before midnight on the west coast). So this converts the browser's current date to Eastern Time for comparison with the other values. Hopefully the values coming from the api are in Eastern Time (which is what I was told) but the `value`s (not the `display-value`s) seem to be at midnight UTC unless I'm reading them wrong. We've been using the `display-values` everywhere as if they were in the user's current timezone.

This is https://jira.cru.org/browse/EP-1412. Thanks :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "gulp-rename": "^1.2.2",
     "jasmine-core": "^2.5.2",
     "jasmine-promise-matchers": "^2.2.0",
-    "jspm": "^0.16.42",
+    "jspm": "^0.16.47",
     "karma": "^1.2.0",
     "karma-babel-preprocessor": "^6.0.1",
     "karma-chrome-launcher": "^2.0.0",
@@ -55,7 +55,6 @@
     "require-dir": "^0.3.0",
     "rsvp": "^3.2.1",
     "run-sequence": "^1.1.5",
-    "systemjs": "^0.19.27",
     "systemjs-plugin-babel": "^0.0.13",
     "vinyl-paths": "^2.1.0",
     "proxy-middleware": "^0.15.0"
--- a/src/app/productConfig/productConfig.modal.spec.js
+++ b/src/app/productConfig/productConfig.modal.spec.js
@@ -1,6 +1,7 @@
 import angular from 'angular';
 import 'angular-mocks';
 import module, {giftAddedEvent} from './productConfig.modal';
+import moment from 'moment';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/from';
 import 'rxjs/add/observable/of';
@@ -16,6 +17,7 @@ describe( 'product config modal', function () {
     uibModalInstance = jasmine.createSpyObj( 'uibModalInstance', ['close', 'dismiss'] );
     productData = {};
     nextDrawDate = '2016-10-01';
+    jasmine.clock().mockDate( moment.utc('2016-10-01').toDate() );
     itemConfig = {
       amount: 85
     };
--- a/src/common/models/recurringGift.model.js
+++ b/src/common/models/recurringGift.model.js
@@ -29,6 +29,9 @@ export default class RecurringGiftModel {
         },
         'next-draw-date': {
           'display-value': ''
+        },
+        'start-date': {
+          'display-value': ''
         }
       };
     }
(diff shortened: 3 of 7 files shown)


Omicron7:
I remember hearing or reading somewhere that `value` is always UTC. Maybe we should be using that instead and just letting moment/javascript Date convert it to user's local time. Either should be able to convert back to UTC easily enough. I also believe we can send `value` instead of `display-value` on POST/PUT's and it should work correctly.

OzzieOrca:
That sounds better. What do we do with the fact that all the dates seem to be at midnight UTC? Even converting that back to Florida time will result in showing the day before to the user...

Omicron7:
We probably need to make sure that EP/cortex is using the correct TimeZone before it generates the unix timestamp.

mattdrees:
Hey Scotty & Brian,

Hopefully I can help verify a few things.
The ECS servers run using Eastern time, but for some reason the docker containers use UTC.
When our EP code creates a DateEntity, it uses the server (docker container) time zone. For some reason, EP decided to represent dates with milliseconds-from-epoch instead of ISO-8601, and we've just followed suit. (Although, it's a little bit of an accident that we got the timezone right; we should be using UTC regardless of what the server time zone is. I'll get that fixed.)

None of the dates EP is sending are conceptually timestamps, right? I believe they're all dates, without time information. I'd recommend, if possible, to treat them that way on the client side. I'm not sure what javascript gives you as far as a 'date without time or timezone' concept.

We've been sending the `display-value` as ISO-8601. If you want, maybe we should just agree that you'll use this and ignore the `value`, since it's awkward to use for dates.

OzzieOrca:
@mattdrees Thanks for looking into this. We've been using `display-value`. Doing so does mean that at different times we are showing the wrong date (in their local time) to some users in the world. Maybe that is ok for this project. Idk.

Using `display-value` is fine (except for the above issue) but it gets interesting when you try to compare it to the browser's date which could be anywhere in the world. That's what this PR hopefully addresses. It converts the browser's date/time to America/New_York and then converts it to the YYYY-MM-DD format for comparison with all the other dates in that format.

adammeyer:
@OzzieOrca @mattdrees Can you clarify why using `value` isn't the best in this case?  That will be in UTC which will avoid the need for this conversion.

OzzieOrca:
To use `value` I think it will need to have some meaningful EST time attached to it. Like we allow donations to be setup for a given day until 5pm EST. Currently `value` is a date at midnight UTC which is intended to mean say Oct 15th but if you convert that to anywhere west of UTC it will be Oct 14th. Midnight is very arbitrary in this case and we sort of want timeless dates. But you can't really do timezone operations on timeless dates. Idk what the right answer is.

mattdrees:
Ok, thanks for the further explanation. I think I see now.

It is a little it arbitrary where we place the cutoff time for a given date, I agree.

What do you think of using UTC for the time, for simplicity's sake?

So, I think this would mean:
 -  if the next draw date is Nov 27, and it's currently Nov 26 at 6:59pm EST, then you *can* schedule a gift for Nov 27
 -  if the next draw date is Nov 27, and it's currently Nov 26 at 7:00pm EST, then you *can not* schedule a gift for Nov 27

Does that sound right?

OzzieOrca:
I think that sounds right. So we always show UTC dates to the user, right? @Omicron7 @mattdrees can you review this and make sure I'm doing it right as it affects most computed dates in the app?

For correctness maybe we should convert all instances of `moment()` to `moment.utc()` to actually have the UTC timezone set for all the dates (or use the `value` property returned from the api with `moment.utc()`). For now we are parsing the string display version (`2015-05-16`) in the browser's local timezone which ignores timezone anyways and accomplishes the same thing.

PR #237, opened 2016-11-03
Profile page

denlight:
I had to do some refactoring of existing endpoint requests due to updates on back end. Let me know about your concerns if you have any. Thanks


src/app/profile/profile.component.js
```diff
@@ -0,0 +1,385 @@
...
+      if(form)form.$setPristine();
+    });
+  }
+
+  loadMailingAddress() {
+    this.mailingAddressLoading = true;
+    this.profileService.getMailingAddress()
+      .subscribe(
+        data => {
+          data.address = formatAddressForTemplate(data.address);
```
OzzieOrca: Can you move this into the service? I think that's how the rest of the services have been implemented.

src/app/profile/profile.component.js
```diff
@@ -0,0 +1,385 @@
...
+          'primary': false,
+          'spouse': item.spouse
+        });
+      }
+      if(item.self && item.delete == undefined) { // update existing phone number
+        this.profileService.updatePhoneNumber(item)
+          .subscribe(
+            () => {
+              this.resetPhoneNumberForms();
+              this.success = true;
```
OzzieOrca: Seems like the first request that returns is going to set success to true and call reset `PhoneNumberForms`, regardless if the other requests have finished or failed. Idk if that is an issue.
You could build up an array of requests and send them off at once with RxJS's forkJoin. That way you'd have only one subscribe that would call the success function only once they have all succeeded or the error function if any of them failed. Handling some successes and but some failures is more complicated. Idk if this context warrants that. You could take a look at https://github.com/CruGlobal/give-web/blob/master/src/app/profile/yourGiving/editRecurringGifts/step4/confirmRecurringGifts.component.js#L37-L64

denlight:
@OzzieOrca , changes are pushed. Thanks!

src/app/profile/profile.component.js
```diff
@@ -169,51 +170,33 @@ class ProfileController {
...
-          );
+        requests.push(this.profileService.deletePhoneNumber(item)
+          .do(() => {
+            pull(this.phoneNumbers, item);
+          })
+        );
+      } else if(!item.self && !item.delete) { // add new phone number
+        requests.push(this.profileService.addPhoneNumber(item)
+          .do((data) => {
+            data.spouse = item.spouse;
```
OzzieOrca: Does this do anything? Seems like nothing is using `data.spouse` here. If you did want to use a modified data object later in a subscribe or something I think you'd need to use `map` here instead of `do` and return the modified object. Maybe it would work with JS references but it wouldn't be very easy to figure out what's going on in the code.

src/app/profile/profile.component.js
```diff
@@ -6,8 +6,8 @@ import profileService from 'common/services/api/profile.service';
 import loadingOverlay from 'common/components/loadingOverlay/loadingOverlay.component';
 import sessionEnforcerService, {EnforcerCallbacks, EnforcerModes} from 'common/services/session/sessionEnforcer.service';
 import {Roles} from 'common/services/session/session.service';
-import formatAddressForTemplate from 'common/services/addressHelpers/formatAddressForTemplate';
 import showErrors from 'common/filters/showErrors.filter';
+import {Observable} from 'rxjs/Observable';
```
OzzieOrca: Can you explicitly include `forkJoin` and `do` here?
```
import 'rxjs/add/observable/forkJoin';
import 'rxjs/add/operator/map';
```
I think it works cuz some other file includes them but they might not always be available in tests or if the files that include this component remove those import.

adammeyer:
@denlight Make sure to also add an entry in build/tasks/release.js.  So the component gets a JS file built.

PR #238, opened 2016-11-03
Bug 1354 - added back removed function

denlight:


OzzieOrca:
Can you bring back the test too?

PR #240, opened 2016-11-04
forced user to stay within a set range of numbers

denlight:
@adammeyer , please review and provide feedback. Thanks!

### DIFF
--- a/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.tpl.html
+++ b/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.tpl.html
@@ -25,9 +25,9 @@
         <div class="alert alert-warning" role="alert" ng-if="!$ctrl.loading && !$ctrl.loadingError && (!$ctrl.recurringGifts || $ctrl.recurringGifts.length === 0)">
           <p translate>You don't seem to have any recurring gifts to edit. You may click continue to set up a new recurring gift for a person or ministry you have given to recently.</p>
         </div>
-        <div ng-repeat="recurringGift in $ctrl.recurringGifts" class="row repeating-row user-profile">
-          <gift-list-item gift="recurringGift">
-            <gift-update-view gift="recurringGift.gift"></gift-update-view>
+        <div ng-repeat="gift in $ctrl.recurringGifts" class="row repeating-row user-profile">
+          <gift-list-item gift="gift">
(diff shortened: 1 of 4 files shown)


src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.tpl.html
```diff
@@ -25,9 +25,9 @@ <h3 translate>Make changes to your recurring gifts</h3>
         <div class="alert alert-warning" role="alert" ng-if="!$ctrl.loading && !$ctrl.loadingError && (!$ctrl.recurringGifts || $ctrl.recurringGifts.length === 0)">
           <p translate>You don't seem to have any recurring gifts to edit. You may click continue to set up a new recurring gift for a person or ministry you have given to recently.</p>
         </div>
-        <div ng-repeat="gift in $ctrl.recurringGifts" class="row repeating-row user-profile">
-          <gift-list-item gift="gift">
-            <gift-update-view gift="gift"></gift-update-view>
+        <div ng-repeat="recurringGift in $ctrl.recurringGifts" class="row repeating-row user-profile">
+          <gift-list-item gift="recurringGift">
+            <gift-update-view gift="recurringGift.gift"></gift-update-view>
```
OzzieOrca: This needs to be using the original object as we are using https://github.com/CruGlobal/give-web/blob/master/src/common/models/recurringGift.model.js which stores the updated value in a different location than the original value. I can look at this more later today. I think we're going to have to modify my detection of if the amount has changed in the recurringGift model or not save it until it is valid.

denlight:
@OzzieOrca , I reversed tempalte change. Thanks

OzzieOrca:
Hey @denlight can you look at https://github.com/CruGlobal/give-web/pull/241? I tried to solve this there. This is complicated because, for recurring gifts, the API sends us the data in `amount` but expects any changes to be made to `updated-amount`. recurringGiftsModel tries to solve/handle this issue but obviously in this case it added a bug.

My PR behaves slightly weirdly when the user inputs an invalid value. Because in recurringGiftsModel I made it show an empty string when ng-model sets it to undefined and it clears the input instantly. I suppose that could be solved by adding allowInvalid to ngModelOptions but that adds more complexity as we don't actually want that value saved.

I think this could be closed in favor of my PR but if you seen anything that is missing from this PR or is broken feel free to fix it.

PR #241, opened 2016-11-08
[EP-1464] In giftUpdateView, add validation to amount input and allow the input…

OzzieOrca:
… to be cleared

- In recurringGiftModel, if updated-amount is undefined, amount getter should return an empty string


denlight:
@OzzieOrca , looks good to me!

PR #243, opened 2016-11-08
Bug ep 1189

denlight:


src/common/components/paymentMethods/bankAccountForm/bankAccountForm.component.js
```diff
@@ -97,7 +97,8 @@ class BankAccountController{
             'bank-name': this.bankPayment.bankName,
             'encrypted-account-number': ccpAccountNumber,
             'routing-number': this.bankPayment.routingNumber
-          }
+          },
+          paymentMethod: this.bankPayment
```
OzzieOrca: What do you think about splitting the last 4 digits off right here and only returning the masked credit card/bank account? I don't think you are using the rest of the paymentMethod object. I just think it would be better to keep the unencrypted number in as few places as possible so future code changes are less likely to expose it somewhere.

You can also use the ccp library to get the last 4. https://github.com/CruGlobal/give-web/blob/master/src/common/lib/ccp.js#L277. The above `new (this.ccp.BankAccountNumber)(this.bankPayment.accountNumber)`  gives you a `SensitiveDataElement `. You can call encrypt and abbreviate on that.

PR #244, opened 2016-11-09
[EP-909] productConfig modal opens nav cart on successfully adding an…

Omicron7:
… item


src/common/components/nav/nav.component.js
```diff
@@ -71,6 +75,13 @@ class NavController{
     this.changeMetaTag('viewport', this.menuType === 'mobile' ? 'width=device-width, minimum-scale=1.0' : 'width=1024');
   }
 
+  // eslint-disable-next-line no-unused-vars
+  giftAddedToCart(event, item) {
```
adammeyer: Are event and item here for future use?

Omicron7: There are there because angular events pass them along. I'm not sure how it would work leaving them off.

PR #248, opened 2016-11-09
removed CVV field from edit payment method modal

denlight:


src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
```diff
@@ -68,7 +68,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Credit Card Paym
           </div>
         </div>
       </div>
-      <div class="col-sm-4">
+      <div class="col-sm-4" ng-hide="$ctrl.paymentMethod">
```
adammeyer: Can ng-if be used here?

denlight: @adammeyer  I did ng-if originally and it messed things up. With ng-if line #100 would throw an error on adding new credit card because this.creditCardPaymentForm.securityCode is undefined. So I decided to hide it instead of potentially adding another watcher to wait for this.creditCardPaymentForm.securityCode to become available. Let me know what you think

adammeyer: I think that's fine.  I was just curious. Thanks!

PR #255, opened 2016-11-10
Change payment method option

denlight:


OzzieOrca:
I don't think this is what was intended... I don't think your Payment Methods page has the ability to change the payment method of a single gift without completely removing a payment method. An example case would be I had a gift to Mike and Becky Crandall using a credit card and now I want that recurring donation to use a new bank account while keeping my other gifts using that old credit card. It also seems weird to jump out of the modal to a completely new page.

If you look at http://crugive.site/manage-giving/ > Stop or Restart Gift > Change my giving to a new payment method, there is UI for this. Choosing Add new payment method results in weird jump to a new modal state which idk is the best. Seems like there should have been a Jira task for implementing this modal flow.

If redirecting to Payment Methods is really what the product managers want, then your code looks fine to accomplish that.

wrandall22:
Hopefully this is helpful.  The current site has 2 places to modify payment methods for recurring donations. One place is while modifying the donations themselves, this allows one to change the payment method on a single donation without affecting the payment method itself.  The other place is in the payment methods, when deleting a payment method. This requires moving all of the associated donations to another payment method.

PR #257, opened 2016-11-11
Updated Spacing

riceguitar:
Dropdown location now aligned with parent
Removed Promo Bucket
Changed Menu Font Weight


riceguitar:
@j2trumpet

PR #258, opened 2016-11-11
fixed validation for editing payment method

denlight:


src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -96,11 +97,10 @@ class CreditCardController {
       this.creditCardPaymentForm.expiryMonth.$validate();
     });
 
-    if(!this.paymentMethod) {
```
OzzieOrca: I think you can keep this if statement since we will be removing the input from the DOM when we are editing. That'll prevent this extra validation from being run and the input will be undefined anyways.

src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
```diff
@@ -71,7 +71,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Credit Card Paym
       <div class="col-sm-4" ng-hide="$ctrl.paymentMethod">
         <div class="form-group is-required" ng-class="{'has-error': ($ctrl.creditCardPaymentForm.securityCode | showErrors)}">
           <label translate>CVV Code</label>
-          <input type="text" name="securityCode" class="form-control  form-control-subtle" ng-model="$ctrl.creditCardPayment.securityCode"ng-attr-placeholder="{{$ctrl.paymentMethod ? '****' : ''}}" ng-required="!$ctrl.paymentMethod || $ctrl.creditCardPayment.cardNumber">
+          <input type="text" name="securityCode" class="form-control form-control-subtle" ng-model="$ctrl.creditCardPayment.securityCode" ng-attr-placeholder="{{$ctrl.paymentMethod ? '****' : ''}}" ng-required="!$ctrl.paymentMethod && $ctrl.creditCardPayment.cardNumber">
```
OzzieOrca: I think you can change `ng-required="!$ctrl.paymentMethod && $ctrl.creditCardPayment.cardNumber"` to `required`. This will make it always required unless the `ng-if` that I suggested you add above hides it because we are editing a payment.

src/common/components/paymentMethods/creditCardForm/creditCardForm.tpl.html
```diff
@@ -71,7 +71,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Credit Card Paym
       <div class="col-sm-4" ng-hide="$ctrl.paymentMethod">
```
OzzieOrca: I think you can make this use `ng-if` to remove it from the DOM and prevent its validation from being run.

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -55,7 +55,8 @@ class CreditCardController {
         cardNumberPlaceholder: this.paymentMethod['card-number'],
         cardholderName: this.paymentMethod['cardholder-name'],
         expiryMonth: this.paymentMethod['expiry-month'],
-        expiryYear: parseInt(this.paymentMethod['expiry-year'])
+        expiryYear: parseInt(this.paymentMethod['expiry-year']),
+        securityCode: ''
```
OzzieOrca: I don't think this matters if we are removing it from the DOM. It won't be used anywhere so maybe toss it.

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -96,11 +97,10 @@ class CreditCardController {
       this.creditCardPaymentForm.expiryMonth.$validate();
     });
 
-    if(!this.paymentMethod) {
-      this.creditCardPaymentForm.securityCode.$parsers.push(this.paymentValidationService.stripNonDigits);
-      this.creditCardPaymentForm.securityCode.$validators.minlength = number => this.paymentMethod && !this.creditCardPayment.cardNumber || toString(number).length >= 3;
-      this.creditCardPaymentForm.securityCode.$validators.maxlength = number => toString(number).length <= 4;
-    }
+    this.creditCardPaymentForm.securityCode.$parsers.push(this.paymentValidationService.stripNonDigits);
+    this.creditCardPaymentForm.securityCode.$validators.minlength = number => this.paymentMethod && !this.creditCardPayment.cardNumber || toString(number).length >= 3;
```
OzzieOrca: I think you can toss `this.paymentMethod && !this.creditCardPayment.cardNumber || ` as that was only validating this input when the cardNumber was changed.

PR #259, opened 2016-11-11
Fixed form field spacing

j2trumpet:
@riceguitar @adammeyer Please review. The only way to consistently get rid of the white space was to put all the fields' code on one line.

### DIFF
--- a/src/app/profile/profile.tpl.html
+++ b/src/app/profile/profile.tpl.html
@@ -31,24 +31,7 @@
                         <li>
                           <form name="$ctrl.donorDetailsForm" novalidate>
                             <div class="form-group is-required person-row">
-                              <i class="fa fa-lock"></i>
-
-                              <select class="form-control form-control-subtle prefix" ng-model="$ctrl.donorDetails.name.title">
-                                <option value="Mr" translate>Mr.</option>
-                                <option value="Mrs" translate>Mrs.</option>
-                                <option value="Dr" translate>Dr.</option>
-                              </select>
-
-                              <input type="text" class="form-control  form-control-subtle first" disabled ng-model="$ctrl.donorDetails.name['given-name']">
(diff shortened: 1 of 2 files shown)


src/assets/scss/_form.scss
```diff
@@ -122,7 +123,7 @@
   }
   input.middle {
     width: 8%;
-    padding: 3px;
+    //padding: 3px;
```
riceguitar: Don't leave commented out code.

src/app/profile/profile.tpl.html
```diff
@@ -31,24 +31,7 @@ <h4 class="panel-title  border-bottom-small visible">Your Name</h4>
...
-                                <option></option>
-                                <option value="Jr." translate>Jr.</option>
-                                <option value="III" translate>III</option>
-                              </select>
-
+                              <i class="fa fa-lock"></i><select class="form-control form-control-subtle prefix" ng-model="$ctrl.donorDetails.name.title"><option value="Mr" translate>Mr.</option><option value="Mrs" translate>Mrs.</option><option value="Dr" translate>Dr.</option></select><input type="text" class="form-control  form-control-subtle first" disabled ng-model="$ctrl.donorDetails.name['given-name']"><input type="text" class="form-control  form-control-subtle middle" disabled ng-model="$ctrl.donorDetails.name['middle-initial']"><input type="text" class="form-control  form-control-subtle last" disabled ng-model="$ctrl.donorDetails.name['family-name']"><select class="form-control form-control-subtle suffix" ng-model="$ctrl.donorDetails.name.suffix"><option></option><option value="Jr." translate>Jr.</option><option value="III" translate>III</option></select>
```
riceguitar: Any reason there is no line breaks in this code? It's a little on the ugly side.

adammeyer: I agree, can you do something like this? http://stackoverflow.com/a/14574698/1099019
Is there a css option?

PR #265, opened 2016-11-14
added link on SignUp modal to switch to SignIn state

denlight:


src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -121,6 +121,9 @@
          ng-click="$ctrl.signUp()"
          translate>Sign Up</a>
     </div>
+    <p class="text-center">
+      <span>Already have an account?</span> <a href="" ng-click="$ctrl.onStateChange({state: 'sign-in'})" translate=""><span class="ng-scope">Sign In</span></a>
```
OzzieOrca: Looks like you copied the html from the browser which rendered the angular directives. Can you change `translate=""` to `translate` and toss the span with `ng-scope`? Thanks

denlight: doh

PR #268, opened 2016-11-15
Multiple stylesheets

riceguitar:
@adammeyer

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,6 +10,7 @@
       <title>Cru | Give</title>
       <base href="/">
       <!-- @if ENV='dev' -->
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -10,6 +10,7 @@
       <title>Cru | Give</title>
       <base href="/">
       <!-- @if ENV='dev' -->
+      <link rel="stylesheet" href="dist/nav.css" type="text/css">
```
adammeyer: Is the nav sheet still part of give.css?  Or did it get removed?

riceguitar: they're exclusive.

PR #269, opened 2016-11-15
[#EP-1466] Allows components to change redirect location on sign-out.

Omicron7:


Omicron7:
This should also fix:
https://rollbar.com/Cru/give-web/items/110/
https://rollbar.com/Cru/give-web/items/115/
https://rollbar.com/Cru/give-web/items/113/

src/app/main/main.tpl.html
```diff
@@ -1,13 +1,13 @@
...
-  <li><a href="cart.html">Cart</a></li>
-  <li><a href="sign-in.html">Sign-In</a></li>
-  <li><a href="profile/payment-methods.html">Payment Methods</a></li>
-  <li><a href="checkout.html">Checkout</a></li>
-  <li><a href="thank-you.html">Thank You</a></li>
-  <li><a href="your-giving.html">Your Giving</a></li>
-  <a href="profile.html">Profile</a>
+  <li><a href="/cart.html">Cart</a></li>
+  <li><a href="/sign-in.html">Sign-In</a></li>
+  <li><a href="/payment-methods.html">Payment Methods</a></li>
```
denlight: same here

src/app/main/main.component.js
```diff
@@ -53,7 +53,7 @@ function routingConfig($stateProvider, $locationProvider, $urlRouterProvider){
       template: '<your-giving></your-giving>'
     })
     .state('payment-methods', {
-      url: "/profile/payment-methods.html",
+      url: "/payment-methods.html",
```
Omicron7: It was updated in this PR, your link points at master.
https://github.com/CruGlobal/give-web/blob/71b5c417e5b0470dceab4e42179aa46828fd76e9/src/app/profile/yourGiving/stopStartRecurringGifts/stopStartRecurringGifts.modal.component.js#L23

OzzieOrca: I thought we were prefixing the profile pages with `/profile` too. Doesn't have to be done in this PR and maybe needs more discussion...

src/app/checkout/checkout.component.js
```diff
@@ -63,6 +64,13 @@ class CheckoutController{
     });
   }
 
+  signedOut( event ) {
```
OzzieOrca: This seems like a bit of code to have everywhere especially since there are tests that go along with it. Could we maybe make a function in the session service that looks something like this that each page could call:
`function redirectOnSignOut($rootScope, destination){...}`
Maybe the service can inject $rootScope directly... I'm not sure if it's available.

Omicron7: The signOut code is only necessary on root level components that care about doing something different from the default that the nav.component provides. I wanted to keep the separation between sessionService and and the components.

src/app/main/main.component.js
adammeyer: Jason doesn't want the /profile.  So we'll leave it without for now.

OzzieOrca: Alrighty. Good to know.

PR #271, opened 2016-11-15
Pdf link bug

denlight:


src/common/services/api/donations.service.js
```diff
@@ -74,12 +74,12 @@ function DonationsService( cortexApiService, profileService, commonService ) {
...
       .map( ( response ) => {
-        angular.forEach( response['receipt-summaries'], ( item ) => {
-          let link = find( response.links, ( r ) => {
-            return r.uri.indexOf( item['transaction-number'] ) != -1;
-          } );
-          item['pdf-link'] = link;
-        } );
+        let links = filter(response.links, (link) => {
+          return link.rel == 'element';
+        }) || [];
```
OzzieOrca: I think filter always returns an array. I tried `_.filter(undefined, () => false)` in the console on the lodash docs page and it returns `[]`.

src/common/services/api/donations.service.js
```diff
@@ -74,12 +74,12 @@ function DonationsService( cortexApiService, profileService, commonService ) {
         data:           data
       } )
       .map( ( response ) => {
```
OzzieOrca: This seems to be missing tests.

src/common/services/api/donations.service.js
```diff
@@ -74,12 +74,12 @@ function DonationsService( cortexApiService, profileService, commonService ) {
...
-          let link = find( response.links, ( r ) => {
-            return r.uri.indexOf( item['transaction-number'] ) != -1;
-          } );
-          item['pdf-link'] = link;
-        } );
+        let links = filter(response.links, (link) => {
+          return link.rel == 'element';
+        }) || [];
+        for(let i=0; i < links.length; i++) {
+          response['receipt-summaries'][i]['pdf-link'] = links[i];
```
OzzieOrca: Are we sure that the link elements are in the same order as the receipt-summaries? There's not a whole lot you can do on your end. Just want to make sure.

src/common/services/api/donations.service.js
denlight: Yeah, good catch. I just changed it

src/common/services/api/donations.service.js
denlight: Apparently yes. BE team promises that it will always be in a right order.

OzzieOrca: K cool.

src/common/services/api/donations.service.js
denlight: @OzzieOrca , tests added

PR #272, opened 2016-11-15
[#EP-1492] X should cancel restart giving

Omicron7:


Omicron7:
This fixes https://jira.cru.org/browse/EP-1492 and updates a couple missing translations.

PR #275, opened 2016-11-15
[#EP-1444] Clear Address 1&2 on country change

Omicron7:
This fixes https://jira.cru.org/browse/EP-1444

### DIFF
--- a/src/common/components/addressForm/addressForm.component.js
+++ b/src/common/components/addressForm/addressForm.component.js
@@ -23,12 +23,12 @@ class AddressFormController {
       .subscribe((data) => {
         this.countries = data;
         if(this.address){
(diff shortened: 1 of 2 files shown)


src/common/components/addressForm/addressForm.component.js
```diff
@@ -23,19 +23,24 @@ class AddressFormController {
...
-          this.refreshRegions(this.address.country);
+          this.refreshRegions(this.address.country, true);
         }
       });
   }
 
-  refreshRegions(country){
+  refreshRegions(country, initial = false){
     country = find(this.countries, {name: country});
     if(!country){ return; }
```
OzzieOrca: Would this stop your clearing code from running? I'm not really sure when it wouldn't find a country.

PR #277, opened 2016-11-16
added step-0 to add new payment method for restarting gifts

denlight:
for https://jira.cru.org/browse/EP-1491


src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/restartGift.component.js
```diff
@@ -51,6 +54,11 @@ class RestartGiftController {
         this.paymentMethods = paymentMethods;
         this.nextDrawDate = nextDrawDate;
         this.hasPaymentMethods = paymentMethods && paymentMethods.length > 0;
+        if(!this.hasPaymentMethods) {
```
Omicron7:
```javascript
         RecurringGiftModel.paymentMethods = this.validPaymentMethods;
         RecurringGiftModel.nextDrawDate = this.nextDrawDate;
```
This adds the list of valid (non expired) payment methods and nextDrawDate to the RecurringGiftModel. This is a class property that is used by each instance of RecurringGiftModel to either display payment information or allow editing payment information. If you add a new valid payment method, it should get added to this array or it won't show up in the list of available methods during the configureGift section.

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step0/addUpdatePaymentMethod.component.js
```diff
@@ -0,0 +1,86 @@
+import angular from 'angular';
```
OzzieOrca: Did you make any changes to this component? It looks mostly copied and pasted... I didn't look too closely. Idk how reusable I made it. It could be a bit of effort.

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/restartGift.component.js
```diff
@@ -166,6 +185,9 @@ class RestartGiftController {
       case 'confirm':
         this.step = 'configure';
         break;
+      case 'add-update-payment-method':
+        this.changeState( {state: 'step-0'} );
```
OzzieOrca: Should this go back to step0PaymentMethodList if `hasPaymentMethods`?  That should mean they came from the list to this state. If they don't have payment methods, going to step-0 should be good. (That's the list of Restart, Change, Redirect, Stop right?)

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step0/addUpdatePaymentMethod.component.js
denlight: @OzzieOrca , I wrote it from scratch at first. And then I found out you had something similar. I thought I could reuse it but then I saw that there is a difference in bindings. Sp I kept that component and reused payment method selector for updating.

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/restartGift.component.js
denlight: good catch!

src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/step0/addUpdatePaymentMethod.component.js
OzzieOrca: Alrighty. I think that's fine. If you have some time in the future maybe we could consolidate them.

PR #278, opened 2016-11-17
Create common giftSearchView component and use it in redirectGift modal

OzzieOrca:
@Omicron7 I needed this for my one time gift modal. Figured this would be easier to review outside of a large PR.

I don't think I actually needed to update RxJS for this but I figured it would be good to move it to a RC version.

### DIFF
--- a/src/common/components/giftViews/giftSearchView/giftSearchView.component.js
+++ b/src/common/components/giftViews/giftSearchView/giftSearchView.component.js
@@ -1,0 +1,104 @@
+import angular from 'angular';
+import map from 'lodash/map';
+import forEach from 'lodash/forEach';
+import find from 'lodash/find';
+import filter from 'lodash/filter';
+import {Observable, Subject} from 'rxjs';
+import 'rxjs/add/observable/empty';
+import 'rxjs/add/operator/debounceTime';
+import 'rxjs/add/operator/distinctUntilChanged';
+import 'rxjs/add/operator/do';
+import 'rxjs/add/operator/switchMap';
+
+import giftListItem from 'common/components/giftViews/giftListItem/giftListItem.component';
+import loadingComponent from 'common/components/loading/loading.component';
+
+import designationsService from 'common/services/api/designations.service';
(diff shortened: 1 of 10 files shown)


src/common/services/api/profile.service.js
```diff
@@ -176,7 +177,7 @@ class Profile {
 
 
   updateMailingAddress(mailingAddress){
-    let mailingAddressCopy = Object.assign({}, mailingAddress);
```
OzzieOrca: This change was because phantomjs seemed to not know what `Object.assign` was and a test was failing. Maybe it needs a polyfill but using lodash was a easy fix. Idk how that test passed before.

src/common/components/giftViews/giftSearchView/giftSearchView.component.js
```diff
@@ -0,0 +1,104 @@
...
+      .distinctUntilChanged()
+      .do(() => {
+        this.searchState = 'searching';
+      })
+      .switchMap(query => {
+        // switchMap cancels the existing observable when a new event arrives and uses the latest
+        if(query){
+          return this.designationsService.productSearch( {keyword: query} );
+        }else{
+          this.$scope.$apply(() => this.searchState = 'initial');
```
OzzieOrca: Changing the state here wasn't changing the template until something else caused a digest loop. If you have any ideas on how to avoid the apply that would be great. I guess this could be getting executed outside of what angular is watching but that doesn't quite make sense.

src/common/services/api/profile.service.js
Omicron7: I've run into pahntomjs missing some core js features as well.

Omicron7:
@OzzieOrca Same search component was also used in restartGift (I just pulled component in from redirectGift), it'll need to get updated as well.
https://github.com/CruGlobal/give-web/blob/giftSearchView/src/app/profile/yourGiving/stopStartRecurringGifts/restartGift/restartGift.component.js#L19

PR #284, opened 2016-11-18
use give subnav on give site

adammeyer:


adammeyer:
@OzzieOrca correct, it's going to be used for all of cru.org

PR #288, opened 2016-11-22
[#EP-1516] Added campaign suggested amounts to product modal.

Omicron7:


adammeyer:
@Omicron7 When editing a product from my cart, the focus goes directly to the "custom" amount box, regardless of what amount I had selected.

PR #290, opened 2016-11-22
[#EP-1525] Added skipEvent flag to downgradeToGuest

Omicron7:
This adds a skipEvent flag to the downgradeToGuest method. This method is used both in the global nav sign-out as well as the checkout as guest. Clicking Checkout as Guest was causing the nav to do a page reload when it saw the event.

### DIFF
--- a/src/common/services/session/session.service.js
+++ b/src/common/services/session/session.service.js
@@ -129,7 +129,7 @@ function session( $cookies, $rootScope, $http, $timeout, envService ) {
       .map( ( response ) => response.data );
   }
 
-  function downgradeToGuest() {
+  function downgradeToGuest( skipEvent = false ) {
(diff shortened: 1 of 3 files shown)


src/common/services/session/session.service.js
```diff
@@ -129,7 +129,7 @@ function session( $cookies, $rootScope, $http, $timeout, envService ) {
       .map( ( response ) => response.data );
   }
 
-  function downgradeToGuest() {
+  function downgradeToGuest( skipEvent = false ) {
```
adammeyer: sweet, didn't know you could set a default like that. 👍

Omicron7: only is babel/ES6 I think.

PR #306, opened 2016-12-05
Fixed sub-nav on desktop to keep from going off-screen

j2trumpet:
@riceguitar Please review

### DIFF
--- a/src/assets/scss/_desktop_navigation.scss
+++ b/src/assets/scss/_desktop_navigation.scss
@@ -88,6 +88,14 @@ $promo-width: 200px;
       background: $colorGrayscale-pebble;
       display: block;
     }
+    > li:nth-of-type(5) {
+      .col-level-2, .col-level-3 {
+        width: $col-width - 57px;
+      }
+      .col-level-3 {
+        left: $col-width - 57px;
+      }
+    }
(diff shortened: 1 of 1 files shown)


src/assets/scss/_desktop_navigation.scss
```diff
@@ -88,6 +88,14 @@ $promo-width: 200px;
       background: $colorGrayscale-pebble;
       display: block;
     }
+    > li:nth-of-type(5) {
```
adammeyer: @j2trumpet @riceguitar Should this be a fixed number 5?  Can you do last in list?

j2trumpet: As it is, yes, it should be a fixed 5 as the Give link will be the last list item. However, @riceguitar and I have talked about alternate solutions since then that I'm working on now.

adammeyer: "About" is the 5th link on the stage site. https://give-stage2.cru.org/  and that could change in the future.

j2trumpet: New solution in place. See https://github.com/CruGlobal/give-web/pull/311

PR #310, opened 2016-12-07
Create separate component for nav cart

OzzieOrca:
- Improve loading and error handling
- Add cartUpdatedEvent to reload nav cart without opening the dropdown
- Cleanup and use common HTML and CSS for desktop and mobile
- Remove pointer from dropdown and prevent dropdown from closing on click
- Use ng-show instead of ng-if for mobile nav cart so it stays up to date without having to load every time it is opened

I'm not sure which way to go with that last bullet point. Should the mobile nav cart preload the cart items and watch the events to stay up to date (provides better UX when the user opens their cart) or should it load the cart data every time it is opened (would save data and processing unless the user opens it repeatedly)? Now that I think about it, if this is going to be on all Cru pages maybe I should change it back. Maybe we could find a way to wait until the first load.

Should the mobile cart be opened when a gift is added to cart? Currently it doesn't seem to be.

@Omicron7 or @adammeyer Do you have any thoughts on the above questions?

@riceguitar Do you have any thoughts on the above questions? Can you also review my HTML/CSS changes described in the comments below?

I wanted to consolidate the nav cart into one set of HTML (and a separate Angular component) and it seems like the mobile cart might be slightly out of date (missing cart button) and is missing some data which may be purposeful on mobile. I don't want to take away any purposeful decisions to make it different on mobile but I think I like the desktop implementation better. What do you think of this screenshot? The first 2 are my changes and the last one is the current mobile cart from [give-stage2.cru.org](https://give-stage2.cru.org).

screen shot 2016-12-07 at 12 06 47 am

Specific differences that the old mobile styling had that you may want me to change back are:
1. More padding (didn't make sense to me on mobile where there is less room)
2. Yellow text (hard to read on grey)
3. No cart button (if this isn't desired, are there other ways for a user to get to the cart.html page?)
4. No frequency, item count, or subtotal
5. Smaller images (the bigger ones seem to fit fine on iPhone. Maybe you guys are testing on some smaller devices too)

### DIFF
--- a/src/app/cart/cart.component.js
+++ b/src/app/cart/cart.component.js
@@ -9,6 +9,7 @@ import productModalService from 'common/services/productModal.service';
 import desigSrcDirective from 'common/directives/desigSrc.directive';
 
 import displayRateTotals from 'common/components/displayRateTotals/displayRateTotals.component';
+import {cartUpdatedEvent} from 'common/components/nav/navCart/navCart.component';
 
 import template from './cart.tpl';
 
@@ -17,7 +18,8 @@ let componentName = 'cart';
 class CartController {
 
   /* @ngInject */
-  constructor( $window, $log, cartService, sessionService, productModalService ) {
+  constructor( $scope, $window, $log, cartService, sessionService, productModalService ) {
+    this.$scope = $scope;
     this.$window = $window;
     this.$log = $log;
     this.productModalService = productModalService;
@@ -60,6 +62,7 @@ class CartController {
       .subscribe( () => {
           pull(this.cartData.items, item);
           this.loadCart(true);
+          this.$scope.$emit( cartUpdatedEvent );
         },
         error => {
           this.$log.error('Error deleting item from cart', error);
--- a/src/app/cart/cart.component.spec.js
+++ b/src/app/cart/cart.component.spec.js
@@ -6,6 +6,8 @@ import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/of';
 import 'rxjs/add/observable/throw';
 
+import {cartUpdatedEvent} from 'common/components/nav/navCart/navCart.component';
+
 describe('cart', () => {
   beforeEach(angular.mock.module(module.name));
   let self = {};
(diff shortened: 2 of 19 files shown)


adammeyer:
@OzzieOrca Is there a JIRA task this work relates to?

adammeyer:
@OzzieOrca The cart should not load until opened.  This nav is also for cru.org, so gathering the cart of every page load seems unnecessary.

OzzieOrca:
@adammeyer Good point. What do you think about waiting to initialize it until it is first opened? Basically avoid [this line](https://github.com/CruGlobal/give-web/pull/310/files#diff-53b806acc67d836922a9d90bd1209a89R27) until it is opened on either mobile or desktop.

I was trying to think about how to implement that. I might be able to do it with a binding from the parent nav component or with an Angular event. Let me know if you have any suggestions.

The ticket I made for this is: https://jira.cru.org/browse/EP-1577

adammeyer:
@OzzieOrca yes, that sounds good.  Only trigger the cart load on open.

What are the new event listeners used for?  Is that if you have the cart menu open when you add a gift?  If the cart loads when it is opened, are those necessary?

OzzieOrca:
I was trying to prevent the nav cart from having to load every time it is opened and instead cache the data. When testing I was frustrated that it would clear all my items and hide the buttons while loading. Even if you had accidentally clicked away and want to open it again, it remembers nothing and reloads.

The events are used to clear the cache and open the dropdown if necessary
- **giftAddedEvent** opens the nav cart and reloads it
- **cartUpdatedEvent** reloads the nav cart without opening it. Used for edit, delete, and completion of order.
- **sessionService.sessionSubject** when signing out or in I think cortex may change the cart contents depending on the user or if they are a guest

I guess since we don't want it loading on page load, we should do a combination of those. Wait to load the cart and start caching the data until it has been interacted with (opened or one of the events fired) and then use the events to keep it up to date. I guess I should avoid watching the session changes until the cart has been initialized.

Maybe I'm making this too complex but I do think it adds value. I don't think it will take much more to hold off on loading until the nav cart is interacted with.

adammeyer:
@OzzieOrca I agree.  This adds value.  If it can wait until it has been interacted with once, I think it's good to go.

OzzieOrca:
@adammeyer I made those changes. I was hoping @riceguitar would take a look at this before it got merged. Is he the right guy to ask or is there someone else who can review the CSS changes?

riceguitar:
Only a couple comments... I think overall it's a good set of changes:

1. Font weight of Account ID reduce to 400
2. Font weight of Line Item Amount reduce to 400

Other than that, I think this should be fine to merge.

OzzieOrca:
@riceguitar Thanks. I'll make those changes.

OzzieOrca:
@adammeyer or @Omicron7 can you review? I made the changes David suggested.

PR #313, opened 2016-12-07
Reload cart after deleting item to refresh rate totals

OzzieOrca:
https://jira.cru.org/browse/EP-1561

We could avoid this request if we subtract the gift value from the rate totals. Seems like we are using the string version and not the number version for displaying the rate totals which makes it more complicated. We could switch that. What do you guys think?


adammeyer:
This is good.  I think changing totals manually would get messy.

PR #316, opened 2016-12-08
Remove extra requests in payment-methods component

OzzieOrca:
- Don’t load in $onInit since the EnforcerCallbacks call the load methods
- Load mailingAddress from donordetails once and pass it to subcomponents with a binding
- Don’t responsively wrap details button as soon
- Remove extra space from bottom of modals

### DIFF
--- a/src/app/profile/payment-methods/payment-methods.component.js
+++ b/src/app/profile/payment-methods/payment-methods.component.js
@@ -15,7 +15,7 @@ class PaymentMethodsController {
 
   /* @ngInject */
   constructor($rootScope, $uibModal, profileService, sessionEnforcerService, $log, $timeout, $window, $location) {
-    this.log = $log;
+    this.$log = $log;
     this.$rootScope = $rootScope;
     this.$uibModal = $uibModal;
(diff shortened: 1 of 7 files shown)


src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -54,14 +54,14 @@ class PaymentMethodsController {
     this.$rootScope.$on( SignOutEvent, ( event ) => this.signedOut( event ) );
 
     this.loading = true;
-    this.loadPaymentMethods();
-    this.loadDonorDetails();
```
OzzieOrca: It used to call the load methods twice, once in $onInit and once on the session enforcer signin callback. @Omicron7 this is the question I asked you on Hipchat: Is it safe to remove the $onInit ones and rely only on the session enforcer? What would you suggest/what have you been doing elsewhere?

Omicron7: Yes, the SessionEnforcer by default calls the sign-in callback on init if the session is signed in. It's safe to remove the calls from $onInit.

PR #323, opened 2016-12-09
handle error code from search results

adammeyer:


src/app/searchResults/searchResults.component.js
```diff
@@ -45,6 +45,9 @@ class SearchResultsController {
           this.searchResults = results;
         }
         this.loadingResults = false;
+      }, () => {
+        this.searchResults = [];
```
OzzieOrca: Should we show the user an error message that maybe includes a message to try again?

Also would it be useful to log this to rollbar?

OzzieOrca:
Is this https://jira.cru.org/browse/EP-1595? If so can I assign it to you?

src/app/searchResults/searchResults.component.js
adammeyer: Yes and yes.  I'll take care of that.

adammeyer:
@OzzieOrca Error message added

OzzieOrca:
Did we want to add a $log statement too?

adammeyer:
I can, do you recommend I do?  I was thinking that's a gateway error, not FE.

OzzieOrca:
Idk. I've been logging most API failures. Does the gateway log every error status code? Logging in give-web would let you know where that failed request originated from. (Although until we get sourcemaps working it's hard to tell exactly)

adammeyer:
@OzzieOrca Let me know if you have any other comments.

PR #331, opened 2016-12-13
[#EP-1612] Show different success message if questions skipped

Omicron7:


Omicron7:
Fixes https://jira.cru.org/browse/EP-1612

PR #335, opened 2016-12-13
Suppress gulp errors on watch only

OzzieOrca:
If you guys see a way to do this without using `globals` or creating another gulp task let me know. Both of those feel hacky but I wasn't sure how else to do it.

This should fix the errors we were seeing today when the Travis build wasn't failing on an error in `gulp release`.


adammeyer:
@OzzieOrca Why so we ever want to suppress gulp errors?

Omicron7:
@adammeyer When you do `gulp serve` it watches for changes. Without suppressing the errors, it would halt gulp and you need to re-run gulp serve again. So, we want to suppress errors when developing, you'll still see em in the console but it won't kill the watch. The issue was that it was also suppressing them during the travis build, so the build would succeed even though sass couldn't compile.

PR #336, opened 2016-12-13
[#EP-1616] Fixes year getting set to current when no receipts exist

Omicron7:
Fixes https://jira.cru.org/browse/EP-1616


Omicron7:
It only goes back a year when calling getReceipts from $onInit. This may drop the user into the UI with the previous year selected, which is fine. They can select current year and it'll say "no receipts".

PR #346, opened 2016-12-16
Ep-1579 - fix touch ui authoring interface issue

mgeatz:
Resolves authoring issue where page is most inaccessible due to conflicting styles in Touch UI.


mgeatz:
@adammeyer I believe you are correct. This is give-web though, so changes to scss here take effect, right?

adammeyer:
@mgeatz yes, you're correct.  For some reason I thought this was give-aem.  All good!

PR #349, opened 2016-12-16
[#EP-1558] Added scrollspy to global navigation.

Omicron7:


adammeyer:
@Omicron7 super! thanks

PR #350, opened 2016-12-19
switch designation editor button endpoint

adammeyer:


src/app/designationEditButton/designationEditButton.component.js
```diff
@@ -17,7 +17,7 @@ class DesignationEditButtonController {
 
   $onInit() {
     if(this.designationNumber && includes([Roles.identified, Roles.registered], this.sessionService.getRole())){
-      this.designationEditorService.getContent(this.designationNumber, this.campaignPage).then(() => {
+      this.designationEditorService.checkPermission(this.designationNumber, this.campaignPage).then(() => {
```
Omicron7: You should add an error callback here as well. Have it $log to rollbar if an error occurs. Not sure if it needs to let the user know if an error happened.

src/common/services/api/designationEditor.service.js
```diff
@@ -11,6 +11,16 @@ class designationEditorService {
     this.$http = $http;
   }
 
+  checkPermission(designationNumber, campaign) {
+    return this.$http.get(designationConstants.designationCanEdit, {
```
Omicron7: Coverage is showing this never gets called.

adammeyer: Coverage has been fixed.

src/app/designationEditButton/designationEditButton.component.js
adammeyer: Because a failed response code is normal behavior if the user doesn't have access to edit, I don't think that should be logged to rollbar.

Omicron7: I'd suggest still adding a
```javascript
() => {
    this.showEditButton = false;
}
```
just so it's clear.

PR #352, opened 2016-12-19
[#EP-1575] Added error checking to all Observables in /app

Omicron7:
This add error checking and messages to all Observables in src/app.
Still need to do src/common and well as add error checking to all promises.

### DIFF
--- a/src/common/components/retryModal/retryModal.tpl.html
+++ b/src/common/components/retryModal/retryModal.tpl.html
@@ -1,0 +1,46 @@
+<div class="modal-header">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-xs-12">
+        <div class="border-bottom-small">
+          <button type="button" class="close" ng-click="$ctrl.cancel()" aria-label="{{'Close' | translate}}">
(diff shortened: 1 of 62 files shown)


src/common/components/retryModal/retryModal.tpl.html
```diff
@@ -0,0 +1,46 @@
+<div class="modal-header">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-xs-12">
+        <div class="border-bottom-small">
+          <button type="button" class="close" ng-click="$ctrl.cancel()" aria-label="{{'Close' | translate}}">
+            <span aria-hidden="true">×</span>
+          </button>
+          <h3 translate>{{$ctrl.title}}</h3>
```
OzzieOrca: We have to translate sooner where we actually have the string right?

Omicron7: yeah, I'll fix that.

Omicron7:
@adammeyer @OzzieOrca This issue is ready for review.

PR #353, opened 2016-12-20
Fixed level-3 nav position in Firefox

j2trumpet:
@riceguitar @adammeyer Please review for completing https://jira.cru.org/browse/EP-1623

### DIFF
--- a/src/assets/scss/_desktop_navigation.scss
+++ b/src/assets/scss/_desktop_navigation.scss
@@ -131,7 +131,6 @@ $promo-width: 200px;
 
     .has-children {
       > a {
-        /* padding-right: 9px; */
         position: relative;
         > span {
           display: inline-block;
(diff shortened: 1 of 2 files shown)


src/assets/scss/_desktop_navigation.scss
```diff
@@ -130,10 +130,23 @@ $promo-width: 200px;
     }
 
     .has-children {
-      filter: grayscale(100%);
-      background: url("/assets/img/desktop-nav-arrow.png") no-repeat right 15px center;
-      background-size: 7px;
-      &:hover {
+      > a {
+        /* padding-right: 9px; */
```
adammeyer: If this line isn't needed, I would suggestion removing it.

j2trumpet: Removed

PR #354, opened 2016-12-22
[#EP-1572] Sign Out spinner

Omicron7:
This adds a simple modal to act as a spinner during the sign-out / downgrade process. This blocks interactions with the page while signing out and will be removed after sign-out, although most sign-out actions either reload or navigate to another page.

### DIFF
--- a/src/common/components/nav/desktopNav.tpl.html
+++ b/src/common/components/nav/desktopNav.tpl.html
@@ -54,7 +54,7 @@
           <ul class="pull-right" uib-dropdown-menu>
             <span class="triangle"></span>
             <li class="nav-edit-profile"><a href="profile.html" translate>Edit Profile</a></li>
-            <li class="nav-sign-out"><a href="" ng-click="$ctrl.signOut()" translate>Sign Out</a></li>
+            <li class="nav-sign-out"><a href ng-click="$ctrl.signOut()" translate>Sign Out</a></li>
           </ul>
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


src/common/components/nav/nav.component.js
```diff
@@ -127,7 +129,15 @@ class NavController{
...
-    this.sessionService.downgradeToGuest().subscribe(angular.noop);
+    let modal = this.$uibModal.open({
+      templateUrl: signOutTemplate.name,
+      backdrop: 'static',
+      keyboard: false,
+      size: 'sm'
+    });
+    this.sessionService.downgradeToGuest().subscribe(() => {
+      modal.close();
+    }, angular.noop);
```
OzzieOrca: Idk what could cause an error but should we close the modal on error too? I guess leaving it there would force the user to refresh the page as it stops them from doing anything... But then they still might not be signed out.

Omicron7: Technically, that close() code probably never happens, the page should reload or redirect before it happens. I just added it as a fallback, just in case.

src/common/components/nav/desktopNav.tpl.html
```diff
@@ -54,7 +54,7 @@
           <ul class="pull-right" uib-dropdown-menu>
             <span class="triangle"></span>
             <li class="nav-edit-profile"><a href="profile.html" translate>Edit Profile</a></li>
-            <li class="nav-sign-out"><a href="" ng-click="$ctrl.signOut()" translate>Sign Out</a></li>
+            <li class="nav-sign-out"><a href ng-click="$ctrl.signOut()" translate>Sign Out</a></li>
```
adammeyer: I believe this should be `href=""`.  I remember issues with IE 9 and 10 not recognizing the anchor tag as a link without it.

PR #379, opened 2017-01-09
Keep Mobile Scroll always on.

riceguitar:
Keeps Horizontal Mobile Scrolling always on. I know Jason already closed the issue, but this is still lingering and hanging out primarily when the Nav is being used on the rest of Cru.org with the long titles.


riceguitar:
@adammeyer

PR #384, opened 2017-01-10
revert photo upload timeout

adammeyer:


adammeyer:
@ndedios FYI

src/app/designationEditor/photoModal/photo.modal.js
```diff
@@ -1,28 +1,39 @@
...
     this.photos = photos;
   }
 
-  uploadComplete(response) {
-    this.photos = response.data;
-    this.uploading = false;
+  uploadComplete() {
+    //refresh photos (set timeout to give Adobe time to add to DAM)
+    this.$timeout(() => {
+      this.designationEditorService.getPhotos(this.designationNumber, this.campaignPage).then((response) => {
```
ndedios: @adammeyer is this method sending the campaign code to the endpoint? image upload and image listing (POST and GET) does not process the campaign parameter (as we previously check yesterday, images are shared between a designation and their campaigns)

adammeyer: correct, I can remove it

adammeyer: All images are under /content/dam/give/designations. Regardless if they're a designation or a campaign.  Is that correct?

ndedios: Yes, always are in the same location.

PR #403, opened 2017-01-20
Use cru-payments library to handle credit card and bank account transactions

OzzieOrca:
This still needs coverage. Matt may want to merge it in the morning to match the TSYS implementation currently deployed to UAT. Or he may roll back UAT and test this for another day.


mattdrees:
I'm going to merge. Thanks Scotty.

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -131,24 +119,42 @@ class CreditCardController {
...
-            paymentMethodNumber: this.creditCardPayment.cardNumber ? this.creditCardPayment.cardNumber.slice(-4) : false
-          }
-        }
-      });
+      this.tsysService.getManifest()
+        .mergeMap(data => {
+          cruPayments.creditCard.init(this.envService.get(), data.deviceId || '88812128320102', data.manifest); //TODO: remove hard coded deviceId when added to the manifest api endpoint
+          return this.paymentMethod && !this.creditCardPayment.cardNumber ?
+            Observable.of({ tsepToken: this.paymentMethod['card-number'], maskedCardNumber: this.paymentMethod['card-number'] }) : // Send masked card number when card number is not updated
+            cruPayments.creditCard.encrypt(this.creditCardPayment.cardNumber, this.creditCardPayment.securityCode, this.creditCardPayment.expiryMonth, this.creditCardPayment.expiryYear);
```
wrandall22: Will `this.creditCardPayment.securityCode` be null when it needs to be? The cases where we won't have one would be:
* Creating a new credit card via self-service
* Editing an existing credit card to a new card number via self-service
* Giving recurring donations with a credit card during checkout
* Giving a mix of recurring and one-time donations with a credit card during checkout

OzzieOrca: @wrandall22 I hadn't worked on that in this PR. @mattdrees thought it would be fine to require it on the front end and just ignore it in EP:

> Currently, we can't use a cvv anywhere except checkout with a brand new card. (And actually if the gift is recurring or if there's a comment-to-dsg, the cvv will be ignored because the CC charge will be performed by siebel (later on) instead of EP (right now). But I don't think you need to stop sending the cvv in this case; we'll just ignore it.) There is a decent chance that eventually we'll do a zero-dollar card verification call when the user adds a card via self-service, and that can use the cvv.

We can discuss this more.

If we want to ask for the CVV less, an easy solution for me would be to simplify the rules you wrote to hide the CVV field only when editing a credit card. There would still be the cases where we add a new card and ask for a CVV but don't use it. I could add logic to check if we are in self service when adding a card or if there are only recurring donations or only donations with comments but that logic could get complex, especially if I need to get some of that data from another api request.

Currently I'm collecting the CVV most places but omitting the CVV on any API requests except for the enhancedPurchaseForm. I'm also collecting the CVV when the user enters a new number while editing but hiding the CVV field until they've typed a valid number.

What direction do you guys think we should go? Everything should be doable for me but figuring out if a cart has gifts that qualify for realtime cc transactions could be more difficult, especially since the user can go back and add/remove anything from their cart.

wrandall22: Tagging @sixtrojans since this is kind of a UX decision.

sixtrojans:
If you all don't mind I'm going to move this discussion to the associated Jira task... EP-1637
@OzzieOrca @wrandall22

PR #407, opened 2017-01-25
Disable rollbar unhandled rejections

OzzieOrca:
I don't really know what code is triggering this but this should disable rollbar from reporting all these errors and hitting our rate limit. https://rollbar.com/Cru/give-web/items/211/.

They are getting fired on all kinds of cru.org pages. Idk if AEM uses promises anywhere.


OzzieOrca:
@adammeyer How do we get this into production? Do you have a deployment process setup?

adammeyer:
@OzzieOrca Done.  The "NG Prod Delpy" Jenkins job needs to be run manually to move the files to the prod bucket.

PR #416, opened 2017-02-03
cruNav bindings

adammeyer:
Add navigation options as component attributes so nav can be re-used on GSW.


src/common/components/nav/nav.component.js
```diff
@@ -39,10 +39,22 @@ class NavController{
     this.sessionModalService = sessionModalService;
 
     this.imgDomain = envService.read('imgDomain');
-    this.navFeed = '/bin/cru/site-nav.json';
   }
 
   $onInit() {
+    //init attributes
```
adammeyer: @OzzieOrca I want to do something like you have in the loading component (https://github.com/CruGlobal/give-web/blob/master/src/common/components/loading/loading.component.js#L36).  Have the defaults in the constructor.  But I wasn't able to get that to work, I always got undefined for attributes not defined in the html.  Am I misunderstanding the loading component?  Thanks

OzzieOrca: Huh that's interesting. The constructor should run and then once the bindings are ready, $onInit should run. You'd think a non-existent binding wouldn't override the value. I'm not sure... The type of binding is the same as the loading component... I suppose you could put a breakpoint at the end of the constructor and the beginning of $onInit to make sure the values are as you expect then to be but I imagine it gets set to undefined between there.

adammeyer: That's correct.  It is correct at the end of constructor, then gets set to undefined before $onInit.  I guess I'll leave it how I have it unless you have another suggestion.

OzzieOrca: Ok. Not really... That's weird.

PR #420, opened 2017-02-06
Fallback to main node if '/content/cru/us/en' doesn't exist

adammeyer:
Also remove manually Give addition, that can be activated via AEM json

### DIFF
--- a/src/common/components/nav/nav.component.js
+++ b/src/common/components/nav/nav.component.js
@@ -252,6 +252,6 @@ export default angular
       navFeed: '@navigationEndpoint',
       searchResultsPath: '@',
       editProfilePath: '@',
-      pullRightOptions: '@'
+      pullRightOptions: '<'
     }
   });
--- a/src/common/components/nav/nav.component.js
+++ b/src/common/components/nav/nav.component.js
@@ -169,20 +169,11 @@ class NavController{
     }))
       .map((response) => {
         let jsonStructure = response.data;
-        let menuStructure = {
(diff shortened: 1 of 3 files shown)


src/common/components/nav/nav.component.js
```diff
@@ -174,20 +174,11 @@ class NavController{
...
           global: jsonStructure['/content/cru/us/en/global'],
           give: jsonStructure['/content/give/us/en']
         };
-
-        //add give to main nav
-        menuStructure.main.push({
-          title: 'Give',
-          path: '/give',
-          children: jsonStructure['/content/give/us/en']
-        });
```
ndedios: I think this code section is still required if the jsonStructure var doesn't contains a "main" component. what do you think?

adammeyer: That part adds the Give link to the main menu.  That has to be removed otherwise "Give" will show on GSW.  We can enable a Give page in the AEM json to add that link back in for cru.org

ndedios: Ok, got it, but if this section code is removed, we must change the endpoint at cru.org and give sites to add give menu option to the menu. I'd like to leave that navigation generation endpoint at AEM side as it currently is to avoid risks.

adammeyer: It's already in the json, "Hide in Nav" just needs to be unchecked.  It won't require any code changes.  I'm not sure why that was hard coded.

ndedios: Ok, got it. Thank you Adam!

PR #421, opened 2017-02-06
Revert hard coded Give menu

adammeyer:


src/common/components/nav/nav.component.js
```diff
@@ -174,11 +174,22 @@ class NavController{
...
         let jsonStructure = response.data;
-        return {
+        let menuStructure = {
           main: jsonStructure['/content/cru/us/en'] || jsonStructure['main'],
           global: jsonStructure['/content/cru/us/en/global'],
           give: jsonStructure['/content/give/us/en']
         };
+
+        //add give to main nav
+        if(jsonStructure['/content/cru/us/en']){
```
adammeyer: @ndedios I added this check so the menu item won't be added on GSW.

ndedios: looks good

PR #422, opened 2017-02-06
add cru.org to production domains

adammeyer:


adammeyer:
@Omicron7 I believe this will solve the issue you saw on your mobile Chrome.

PR #426, opened 2017-02-07
On profile page, remove touched check preventing server errors from b…

OzzieOrca:
…eing shown

- Fix spelling error for donorDetailsLoading variable

This addresses part of https://jira.cru.org/browse/EP-1727 where the server email validation errors aren't even being shown.

### DIFF
--- a/src/app/profile/profile.tpl.html
+++ b/src/app/profile/profile.tpl.html
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-md-12">
 
-        <div ng-if="$ctrl.hasError() && !$ctrl.touched()" class="alert alert-danger" role="alert">
+        <div ng-if="$ctrl.hasError()" class="alert alert-danger" role="alert">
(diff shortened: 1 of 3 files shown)


src/app/profile/profile.tpl.html
```diff
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-md-12">
 
-        <div ng-if="$ctrl.hasError() && !$ctrl.touched()" class="alert alert-danger" role="alert">
```
OzzieOrca: I guess Denys was trying to hide the errors once the user made changes to the form again. To do this correctly I think we'd need to set all the forms back to their pristine state.

I'm not sure this feature is needed as the errors are cleared on submit and a user might refer to the error at the top of the page after touching an input.

If you guys think I should save that feature, let me know. The current logic is preventing errors from being shown. I think it would just require adding $setPristine to all the error handlers or moving that logic up to top of the update functions, before the requests.

PR #429, opened 2017-02-07
Gracefully handle missing next-draw-date

OzzieOrca:
It seems that errored donations don't include a next-draw-date. This makes the frontend handle that case without throwing an error. Should fix https://rollbar.com/Cru/give-web/items/177 which I think was causing us to hit our rate limit.

I'm still not sure why we are showing errored donations to the user as they prevent payment methods from being deleted. That discussion is here https://jira.cru.org/browse/EP-1723.

### DIFF
--- a/src/common/models/recurringGift.model.spec.js
+++ b/src/common/models/recurringGift.model.spec.js
@@ -381,6 +381,10 @@ describe('recurringGift model', () => {
     it('should return the old draw date when start month and transaction day are unchanged', () => {
       expect(giftModel.nextGiftDate.toString()).toEqual(moment('2015-05-06').toString());
     });
+    it('should handle the case when the parent donation doesn\'t have a next draw date', () => {
+      delete giftModel.parentDonation['next-draw-date'];
+      expect(giftModel.nextGiftDate).toBeUndefined();
+    });
   });
 
   describe('initStartMonth', () => {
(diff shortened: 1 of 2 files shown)


src/common/models/recurringGift.model.spec.js
```diff
@@ -381,6 +381,10 @@ describe('recurringGift model', () => {
     it('should return the old draw date when start month and transaction day are unchanged', () => {
       expect(giftModel.nextGiftDate.toString()).toEqual(moment('2015-05-06').toString());
     });
+    it('should handle the case when the parent donation doesn\'t have a next draw date', () => {
+      delete giftModel.parentDonation['next-draw-date'];
+      expect(giftModel.nextGiftDate).toBeUndefined();
```
Omicron7: How is an `undefined` nextGiftDate handled in the UI? Is it calling any moment methods on it that might cause additional errors?

OzzieOrca: We could find a better value. If an error is thrown here, the view never finishes compiling and `{{ gift.nextGiftDate.format('ll') }}` gets spit out on the screen. When it is undefined I think the angular template handles it gracefully and it leaves a blank here
screen shot 2017-02-07 at 12 19 53 pm
I've only tested this on the payment-methods.html page. If moment functions are called in the template we should be fine. If we do something in js, we're going to see an error.

Hopefully the errored gifts get removed from the views and this won't be hit much.

PR #433, opened 2017-02-08
reduce checkout analytics calls

adammeyer:


adammeyer:
Thanks @OzzieOrca for the quick review.  There will more of these in the next few days.  I'm working with the analytics guy to correct/modify things to match what he's looking for.  This PR is reducing the 3 analytics calls down to one when a user lands on the checkout page.

PR #437, opened 2017-02-09
Show generic sign in error if returned error value is falsy

OzzieOrca:
I'm trying to fix https://jira.cru.org/browse/EP-1700. It seems that no error was ever displayed to the user. My thought is that `error.data.error` was something falsy like `''`. I'm not sure the old condition covered all the cases. Feel free to suggest other ways to do this. Idk what the server 500 error was in the jira ticket but the user should be shown an error regardless. I'm also not sure how to reproduce a 500 to test this.


Omicron7:
cortex-gateway just passes along the error from either CAS (key/relay) or cortex (fetching token). CAS ones should have an error. Cortex usually is the one giving the 500 with no reason.

PR #439, opened 2017-02-10
Test rspec selenium tests

Omicron7:


Omicron7:
This adds selenium tests run in rspec on browsertack. These will allow the QA team to create selenium tests using selenium-ide, which will eventually be run before and deploy to production vi jenkins (http://devtools.aws.cru.org:8080/view/Givesite/job/Givesite%20UAT%20Browserstack%20tests/).

Omicron7:
It's very similar to protractor (both use selenium webdriver), except in protractor, the tests are written by the developers . This aims to have the tests written by the QA team using a Drag/Drop IDE (Selenium IDE) which has a save file format of html. The ruby aspect of this reads the html files and runs them as selenium tests inside browserstack.

I've setup the ruby side of this using the normal ruby .gitignore, which doesn't include vendor/cache. It caches gems required to run the tests which speeds up the test setup, as well as provides gems if rubygems is down.

OzzieOrca:
Makes sense. Thanks for the explanation.

PR #440, opened 2017-02-14
analytics additions

adammeyer:


src/app/analytics/analytics.factory.js
```diff
@@ -344,6 +345,26 @@ function analyticsFactory($window, $timeout, sessionService) {
         $window.digitalData.page.pageInfo.onsiteSearchResults = 0;
       }
     },
+    setLoggedInStatus: function(){
+      let ssoGuid = '';
+      if (typeof sessionService !== 'undefined') {
+        if (typeof sessionService.session['sub'] !== 'undefined') {
+          ssoGuid = sessionService.session['sub'].split('|').pop();
```
OzzieOrca: It might be cleaner to make the session service do this and provide a funtion/property to access the guid.

adammeyer: I agree, let me first verify this is what the analytics are looking for.

PR #452, opened 2017-02-16
Display semi annual as an option if already set as semi annual

adammeyer:


adammeyer:
Perfect. Thanks @OzzieOrca!

PR #453, opened 2017-02-16
search results page changes

adammeyer:


adammeyer:
@OzzieOrca correct. The change request(s) are in EP-1223.

PR #454, opened 2017-02-16
update card type on number change

adammeyer:


adammeyer:
@OzzieOrca Can you point me in the right direction on why this test is failing?

OzzieOrca:
The profile service is mocked https://github.com/CruGlobal/give-web/pull/454/files#diff-20938a29b750eae3dc700ac7d5f2e22fR66

OzzieOrca:
Also CCP probably can give you this value. It might take modifying the creditCardForm https://github.com/CruGlobal/give-web/blob/b9725c428ff77a1c41d9a213339054c9301ea3ee/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L140-L140

adammeyer:
Thanks. Which way do you recommend?  If I use JS to get the type, I'll have to create a table to match what that function returns with what cortex will.

OzzieOrca:
Idk. It's frustrating that this is going to change with TSYS and would be cleaner with that implementation. I guess avoiding the network request would be nice but not a big deal.

We could modify https://github.com/CruGlobal/give-web/blob/4ddb73ac6b7d2c72731d89bebc213753a33a8728/src/common/lib/ccp.js#L503-L503 to use `displayName` instead. I'm still not sure those match what the server has exactly but I think they are closer.

adammeyer:
I'm thinking the extra network request is okay.  Especially because it will only run if the card number changed.  Most edits will probably be expiration date.  Then we could just remove that request once TYSY is merged in. sound good?

OzzieOrca:
Yep sounds good

PR #456, opened 2017-02-17
Add other amount input to suggested amounts in productConfigModal

OzzieOrca:
- Default gift amounts inside product modal and remove explicit calls to pass in a default amount of $50
- Default amount to first suggested amount or first selectable amount
- Call changeCustomAmount if loaded amount is not a suggested amount or selectable amount
- Set useSuggestedAmounts during initialization to pick between suggested amounts or selectable amounts view and remove showDefaultAmounts
- Cleanup $onInit and refactor into multiple functions
- Refactor query param updates into updateQueryParam to reduce duplicated calls to isEdit
- Refactor isEdit tests to reduce duplicate code

https://jira.cru.org/browse/EP-1754

### DIFF
--- a/src/app/productConfig/productConfigModal/productConfig.modal.tpl.html
+++ b/src/app/productConfig/productConfigModal/productConfig.modal.tpl.html
@@ -38,24 +38,53 @@
             Gift Amount
           </h4>
           <div class="panel-body pt0">
-            <div ng-if="$ctrl.suggestedAmounts">
-              <div class="radio radio-custom-amount" ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
+            <div ng-if="$ctrl.useSuggestedAmounts">
+              <div class="radio radio-custom-amount"
+                   ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
                 <label>
-                  <input name="amount"
-                         ng-model="$ctrl.customAmount"
-                         ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
+                  <input name="suggestedAmount"
                          type="radio"
-                         ng-value="suggested.amount"
-                         required>{{suggested.amount | currency:'$':2}} {{suggested.label}}
+                         ng-click="$ctrl.changeAmount(suggested.amount)"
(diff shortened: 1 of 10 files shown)


src/app/productConfig/productConfigModal/productConfig.modal.tpl.html
```diff
@@ -38,24 +38,53 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                           ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
+                           ng-focus="$ctrl.customInputActive = true;"
+                           ng-required="$ctrl.customInputActive"
+                           placeholder="{{'Other' | translate}}">
+                  </label>
+                  <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error"
+                       ng-if="($ctrl.itemConfigForm.amount | showErrors)">
+                    <div class="help-block" ng-message="pattern" translate>Your gift must be a valid dollar amount</div>
+                    <div class="help-block" ng-message="required" translate>Amount must be not empty</div>
+                    <div class="help-block" ng-message="minimum" translate>Amount must be at least {{ 1 | currency }}</div>
```
adammeyer: Do you need to specify '$' on these currencies? (line 75 too)

OzzieOrca: I copied that from other parts of the page. https://docs.angularjs.org/api/ng/filter/currency says `When no currency symbol is provided, default symbol for current locale is used.`

adammeyer: Is current locale set by us or the user's browser?

OzzieOrca: `All localizable AngularJS components depend on locale-specific rule sets managed by the $locale service.` Seems like you have to call that explicitly and have loaded other locales. Also see https://docs.angularjs.org/guide/i18n#currency-symbol. Is that what you get from there? I could be wrong. The docs aren't the clearest.

adammeyer: Okay, I think it's good.  Just wanted to double check.

PR #460, opened 2017-02-17
reset state after payment save

adammeyer:
https://jira.cru.org/browse/EP-1766


src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -114,6 +114,7 @@ class PaymentMethodsController {
         .subscribe(data => {
             data.address = data.address && formatAddressForTemplate(data.address);
             this.paymentMethodFormModal.close(data);
+            this.paymentFormResolve.state = null;
```
OzzieOrca: I think I was making the default state `'unsubmitted'`. I think that would be better and make it consistent. It doesn't matter functionally though.

PR #464, opened 2017-02-20
[#EP-1769] Reset session when items added to a public, empty cart.

Omicron7:
This will reset a session anytime an item is added to a public, empty cart. This prevents saved address/payments from being used across guest checkout.

Fixes https://jira.cru.org/browse/EP-1769

### DIFF
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -105,18 +105,19 @@ class Cart {
           this.$cookies.remove(Sessions.give, {path: '/', domain: '.cru.org'});
           // Defer til next digest so $cookie.remove propagates.
           return Observable.from(this.$timeout(angular.noop, 10)).mergeMap(() => {
-            return this.cortexApiService.post({
-              path: ['itemfieldslineitem', uri],
-              data: data
-            });
+            return this._addItem(uri, data);
           });
         }
-        return this.cortexApiService.post({
-          path: ['itemfieldslineitem', uri],
-          data: data
-        });
+        return this._addItem(uri, data);
       });
     }
+    return this._addItem(uri, data);
+  }
+
+  /**
+   * @private
+   */
+  _addItem(uri, data) {
     return this.cortexApiService.post({
       path: ['itemfieldslineitem', uri],
       data: data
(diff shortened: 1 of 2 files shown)


src/common/services/api/cart.service.js
```diff
@@ -82,9 +87,36 @@ class Cart {
...
     data.quantity = 1;
 
+    if(this.sessionService.getRole() == Roles.public) {
+      return this.getTotalQuantity().mergeMap((total) => {
+        if(total <= 0) {
+          this.$cookies.remove(Sessions.cortex, {path: '/', domain: '.cru.org'});
+          this.$cookies.remove(Sessions.give, {path: '/', domain: '.cru.org'});
+          // Defer til next digest so $cookie.remove propagates.
+          return Observable.from(this.$timeout(angular.noop, 10)).mergeMap(() => {
+            return this.cortexApiService.post({
```
OzzieOrca: This code is repeated 3 times. Is that needed?

src/common/services/api/cart.service.js
```diff
@@ -82,9 +87,36 @@ class Cart {
...
+      .get({path: ['carts', this.cortexApiService.scope, 'default']})
+      .map((cart) => {
+        return cart['total-quantity'];
+      });
+  }
+
+  addItem(uri, data) {
     data.quantity = 1;
 
+    if(this.sessionService.getRole() == Roles.public) {
```
OzzieOrca: This might be cleaner in a separate function. That would clarify what it does and might help with the duplicate calls.

src/common/services/api/cart.service.js
Omicron7: I could DRY it up by moving it to another function. I tried assigning the Observable to a variable under the assumption that observables include setup/teardown, but since we are wrapping a promise, the request happens immediately when the observable is created, instead of when it is subscribed too.

OzzieOrca: http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-defer might help something. My impression is you provide it with a function that doesn't get called until it is subscribed too. I see now why you have 3 cases. For some reason I thought the last one was always called but the others get returned before it. Do what you think is best. I won't be too picky.

PR #469, opened 2017-02-21
Disable middle initial and suffix in contactInfo and profile when nam…

OzzieOrca:
…e fields are disabled

https://jira.cru.org/browse/EP-1799

### DIFF
--- a/src/app/profile/profile.tpl.html
+++ b/src/app/profile/profile.tpl.html
@@ -46,7 +46,7 @@
                               <input name="nameMiddleInitial" type="text" class="form-control  form-control-subtle middle" disabled ng-model="$ctrl.donorDetails.name['middle-initial']">
                               <input name="nameFamilyName" type="text" class="form-control  form-control-subtle last" disabled ng-model="$ctrl.donorDetails.name['family-name']">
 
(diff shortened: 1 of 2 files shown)


src/app/profile/profile.tpl.html
```diff
@@ -46,7 +46,7 @@ <h4 class="panel-title  border-bottom-small visible">Your Name</h4>
                               <input name="nameMiddleInitial" type="text" class="form-control  form-control-subtle middle" disabled ng-model="$ctrl.donorDetails.name['middle-initial']">
                               <input name="nameFamilyName" type="text" class="form-control  form-control-subtle last" disabled ng-model="$ctrl.donorDetails.name['family-name']">
 
-                              <select name="nameSuffix" class="form-control form-control-subtle suffix" ng-model="$ctrl.donorDetails.name.suffix">
+                              <select name="nameSuffix" class="form-control form-control-subtle suffix" disabled ng-model="$ctrl.donorDetails.name.suffix">
```
adammeyer: Should this be conditionally disabled?

OzzieOrca: The others aren't... I think the intention is that they will never be edited by our app. But they are all still wired up with ng-model and I think sent to the api. I guess this code could be used if that requirement changes in the future but right now I guess the code isn't needed. I'm not sure what requirements Denys got when he initially implemented this.

PR #486, opened 2017-02-28
autofocus fix for mobile

adammeyer:
https://jira.cru.org/browse/EP-1788

### DIFF
--- a/src/common/directives/autoFocus.directive.js
+++ b/src/common/directives/autoFocus.directive.js
@@ -3,11 +3,13 @@ import angular from 'angular';
 let directiveName = 'autoFocus';
 
 /* @ngInject */
-function desigSrc() {
+function desigSrc($timeout) {
   return {
     restrict: 'A',
     link:     function ( scope, element ) {
-      element[0].focus();
+      $timeout(function () {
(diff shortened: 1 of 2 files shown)


src/common/directives/autoFocus.directive.js
```diff
@@ -3,11 +3,13 @@ import angular from 'angular';
 let directiveName = 'autoFocus';
 
 /* @ngInject */
-function desigSrc() {
+function desigSrc($timeout) {
   return {
     restrict: 'A',
     link:     function ( scope, element ) {
-      element[0].focus();
+      $timeout(function () {
```
OzzieOrca: I guess this is waiting for something else to load. Timeouts aren't ideal but as long as this consistently works it should be fine. Would $scope.$apply or $digest work here? I'm not entirely sure what we're waiting for.

adammeyer: Without the timeout, it worked as intended on desktop, but on mobile it did a crazy scroll thing (see jira task).  Adding a timeout fixed it (I remembered that from ERT).  I will try a $digest and make another PR if that works.

PR #488, opened 2017-03-01
remove password from error before sending to log

adammeyer:


src/common/components/signInForm/signInForm.component.spec.js
```diff
@@ -95,5 +95,17 @@ describe( 'signInForm', function () {
       expect( $ctrl.errorMessage ).toEqual( 'An error has occurred signing in. Please try again.' );
       expect( $ctrl.isSigningIn ).toEqual( false );
     });
+
+    it( 'removes password from error log', () => {
```
adammeyer: @OzzieOrca Can you help me finish this spec?  I'm not sure how to get the response from the rejected deferred.

OzzieOrca: I'd just check what log was called with. https://github.com/CruGlobal/give-web/blob/master/src/app/profile/yourGiving/editRecurringGifts/editRecurringGifts.modal.component.spec.js#L128

PR #493, opened 2017-03-06
[#EP-1845] Add change btn and missing translate tags.

Omicron7:


Omicron7:
Fixes https://jira.cru.org/browse/EP-1845

PR #503, opened 2017-03-10
show payment type if no payment method

adammeyer:
https://jira.cru.org/browse/EP-1855

### DIFF
--- a/src/app/profile/yourGiving/recipientView/recipientGift/recipientGift.component.js
+++ b/src/app/profile/yourGiving/recipientView/recipientGift/recipientGift.component.js
@@ -1,5 +1,7 @@
 import angular from 'angular';
-import {Observable} from 'rxjs/Rx';
+import keyBy from 'lodash/keyBy';
+import {Observable} from 'rxjs/Observable';
+import 'rxjs/add/observable/forkJoin';
 
 import RecurringGiftModel from 'common/models/recurringGift.model';
 import desigSrc from 'common/directives/desigSrc.directive';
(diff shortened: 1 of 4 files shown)


wrandall22:
So, I don't think this is quite what we want.  See comments on [EP-1827](https://jira.cru.org/browse/EP-1827). Summary: if `payment-method-link` is not there, then show `payment-type`.

adammeyer:
@wrandall22 That still leaves some cells blank, is that okay?

wrandall22:
Not really, no. There needs to be a task (or rolled into this one) that uses payment-method-link to get the inactive payment methods.

adammeyer:
What was wrong with what I had?

wrandall22:
What you currently have (if I'm reading it right) will show only the payment type for inactive payment methods instead of the `Card Type - Last Four` for CC and `Bank Name - Last Four` for EFT.

adammeyer:
@OzzieOrca @Omicron7 could you provide your feedback on my latest commit before I write the specs?  Thanks

src/app/profile/yourGiving/recipientView/recipientGift/recipientGift.component.js
```diff
@@ -31,22 +31,38 @@ class RecipientGift {
...
+        if(uri && paymentMethodUris.indexOf(uri) < 0){
+          paymentMethodUris.push(uri);
+          paymentMethodRequests.push(this.profileService.getPaymentMethod( uri, true ));
+        }
+      });
+
+      if(paymentMethodRequests.length){
+        Observable.forkJoin(paymentMethodRequests)
+          .subscribe((response) => {
+            angular.forEach(response, (paymentMethod) => {
```
OzzieOrca: You could avoid a nested loop if you use [keyBy](https://lodash.com/docs/4.17.4#keyBy) on the payment methods first and then for each donation, just look that value up in the object keyBy created.

adammeyer:
@OzzieOrca Ready for your review again.

PR #506, opened 2017-03-14
add periods to Jr/Sr

adammeyer:
https://jira.cru.org/browse/EP-1880


OzzieOrca:
Guess I hadn't loaded the most recent version... You still missed https://github.com/CruGlobal/give-web/blob/master/src/common/components/contactInfo/contactInfo.tpl.html#L217

PR #513, opened 2017-03-15
Fixing button placement in payment methods

j2trumpet:
@adammeyer Please review. I had to restructure some of the HTML including some of the NG if statements. I think my logic/operators are correct. From what I could tell, the display-address.tpl.html file is only being used on this payment methods page, so I thought it was safe to remove the extra div wrapper and not affect anywhere else. This is for https://jira.cru.org/browse/EP-1876

### DIFF
--- a/src/app/profile/payment-methods/payment-method/payment-method.tpl.html
+++ b/src/app/profile/payment-methods/payment-method/payment-method.tpl.html
@@ -15,52 +15,47 @@
     <div uib-collapse="$ctrl.isCollapsed" role="tabpanel" class="panel-collapse">
       <div class="panel-body">
         <div class="container-fluid">
-          <div class="row">
+          <div class="row panel--payment-info">
 
             <div ng-if="$ctrl.isCard()" class="col-sm-6 col-md-4">
               <p class="section-title mb0">Name on card</p>
               <p>{{$ctrl.model['cardholder-name']}}</p>
             </div>
-
-            <div ng-if="!$ctrl.isCard()">
-              <div class="col-sm-6 col-md-4">
-                <p class="section-title mb0">Bank Name</p>
-                <p>{{$ctrl.model['bank-name']}}</p>
-              </div>
-
-              <div class="col-sm-6 col-md-4">
-                <p class="section-title mb0">Account Type</p>
-                <p>{{$ctrl.model['account-type']}}</p>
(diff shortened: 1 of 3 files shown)


OzzieOrca:
Display address is used everywhere in the app where there is an address https://github.com/CruGlobal/give-web/search?utf8=%E2%9C%93&q=display-address&type=Code.

That change looks fine though unless the missing form-group changes spacing somewhere. I'm not sure why that is there to begin with. Maybe because checkout step 3 had it...

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
```diff
@@ -15,28 +15,41 @@
...
             <div ng-if="$ctrl.isCard()" class="col-sm-6 col-md-4">
               <p class="section-title mb0">Name on card</p>
               <p>{{$ctrl.model['cardholder-name']}}</p>
             </div>
-
-            <div ng-if="!$ctrl.isCard()">
-              <div class="col-sm-6 col-md-4">
-                <p class="section-title mb0">Bank Name</p>
-                <p>{{$ctrl.model['bank-name']}}</p>
+            <div ng-if="$ctrl.isCard()" class="col-sm-6 col-md-4">
```
OzzieOrca: Could this go in the same div as above? I haven't tested your new html so maybe not...

j2trumpet: The two ng-if="$ctrl.isCard()" divs right next to each other? They're two different columns so there isn't a big gap of empty space between the payment info and the edit/delete buttons.

OzzieOrca: Oh gotcha. I guess they are stacked on the current version

j2trumpet:
@OzzieOrca Re: Display address - I did a global search and tried to locate every place `<display-address>` was used, but I probably didn't find it all. The default Bootstrap `.form-group` styling was throwing my new styles off and I thought it safe to remove. It would be easy to add back in if we find it's necessary.

src/app/profile/payment-methods/payment-method/payment-method.tpl.html
j2trumpet: They are. @adammeyer approved me moving the lower piece to a new column. That stacking is also part of what was causing the issue the Jira task needed me to fix.

PR #522, opened 2017-03-17
Add Google Custom Search to search results page for cru.org

adammeyer:
https://jira.cru.org/browse/EP-1872

### DIFF
--- a/src/app/searchResults/searchResults.tpl.html
+++ b/src/app/searchResults/searchResults.tpl.html
@@ -14,7 +14,7 @@
 
       <div class="row">
         <div class="col-xs-12 col-sm-10 col-sm-offset-1">
-          <form method="get" action="/search-results.html">
+          <form method="get" action="">
             <div class="input-group is-search-main">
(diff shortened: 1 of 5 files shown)


src/app/searchResults/searchResults.tpl.html
```diff
@@ -14,7 +14,7 @@ <h1 ng-if="!$ctrl.searchParams.type || $ctrl.searchParams.type === 'people'">
 
       <div class="row">
         <div class="col-xs-12 col-sm-10 col-sm-offset-1">
-          <form method="get" action="/search-results.html">
+          <form method="get" action="">
```
OzzieOrca: What does a blank `action` do? Reload the page?

adammeyer: Yes, keeps the current url.  This way it will work if it's search-results.html or search.html.  I could remove it, empty is the same as nothing.

OzzieOrca: That's fine. I didn't really look to see if the html form submit was still being used.

PR #545, opened 2017-03-27
Disable address, email, and phone number fields for staff in profile

OzzieOrca:
- Also disable address in checkout

https://jira.cru.org/browse/EP-1856

### DIFF
--- a/src/common/components/contactInfo/contactInfo.tpl.html
+++ b/src/common/components/contactInfo/contactInfo.tpl.html
@@ -192,7 +192,7 @@
     <div class="mb">
       <h4 class="panel-title  border-bottom-small  visible" translate>Mailing Address</h4>
 
-      <address-form address="$ctrl.donorDetails.mailingAddress" parent-form="$ctrl.detailsForm"></address-form>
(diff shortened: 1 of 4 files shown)


src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -192,7 +192,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>Your Information
     <div class="mb">
       <h4 class="panel-title  border-bottom-small  visible" translate>Mailing Address</h4>
 
-      <address-form address="$ctrl.donorDetails.mailingAddress" parent-form="$ctrl.detailsForm"></address-form>
+      <address-form address="$ctrl.donorDetails.mailingAddress" parent-form="$ctrl.detailsForm" disabled="$ctrl.donorDetails.staff"></address-form>
```
adammeyer: Does this need to be ng-disabled?

OzzieOrca: No it's a component. I added the ng-disabled attributes inside. I suppose I could use the ng-disabled name but it's not native angular functionality...

adammeyer: O, missed that.  Looks good.

PR #549, opened 2017-03-29
Run karma test as signle run so it exits when done.

Omicron7:
This might speed up travis as well as allowing `yarn run test` to exit when finished.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -18,6 +18,8 @@ webpackConfig.module.rules.push({
 module.exports = function(config) {
   config.set({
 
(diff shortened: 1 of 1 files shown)


OzzieOrca:
It is configured to do this for travis https://github.com/CruGlobal/give-web/blob/master/.travis.yml#L4.

If you'd prefer this to be the default for development, go for it. I think it might be faster to not enable single run locally as it should skip the long initial compile after the first run. I'm not sure. I usually run  karma with Intellij's run configurations so I'm fine having it either way.

PR #557, opened 2017-04-04
Update sign up error message to include eGift@cru.org

OzzieOrca:
@Omicron7 Should this also be updated? https://github.com/CruGlobal/give-web/blob/master/src/common/components/signUpModal/signUpModal.tpl.html#L7


Omicron7:
Unless there was an issue to change the contact-us link, I'd leave it as is. We could change it to not be domain specific though.

PR #564, opened 2017-04-06
Allow updating credit cards during checkout

OzzieOrca:
- Store all CVVs entered during current order in session storage and use them when credit cards are selected and edited
- Hide CVV field when editing cards where card number changes are allowed (self service) since CVV isn’t used there

https://jira.cru.org/browse/EP-1925

### DIFF
--- a/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html
+++ b/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html
@@ -5,7 +5,7 @@
       <label>
         <input type="radio" name="paymentMethod" ng-model="$ctrl.selectedPaymentMethod" ng-value="paymentMethod" ng-disabled="expired" required>
         <payment-method-display payment-method="paymentMethod" expired="expired"></payment-method-display>
-        <button class="btn btn-xs btn-link" ng-click="$ctrl.openPaymentMethodFormModal(paymentMethod)" ng-if="paymentMethod.self.type === 'cru.creditcards.named-credit-card'" translate>update</button>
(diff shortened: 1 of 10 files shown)


src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html
```diff
@@ -3,9 +3,9 @@
   <div class="panel-body">
     <div class="radio radio-method" ng-repeat="paymentMethod in $ctrl.paymentMethods" ng-init="expired = !$ctrl.validPaymentMethod(paymentMethod)">
       <label>
-        <input ng-if="!expired" type="radio" name="paymentMethod" ng-model="$ctrl.selectedPaymentMethod" ng-value="paymentMethod" required>
-        <button ng-if="expired" class="btn btn-xs btn-default" ng-click="$ctrl.openPaymentMethodFormModal(paymentMethod)">Update</button>
+        <input type="radio" name="paymentMethod" ng-model="$ctrl.selectedPaymentMethod" ng-value="paymentMethod" ng-disabled="expired" required>
         <payment-method-display payment-method="paymentMethod" expired="expired"></payment-method-display>
+        <button class="btn btn-xs btn-link" ng-click="$ctrl.openPaymentMethodFormModal(paymentMethod)" translate>update</button>
```
Omicron7: I had this link as `update` initially and Jason requested it be changed to `edit`.

Maybe switch label depending on on if it's expired?

OzzieOrca: Idk that users should have to think about what the difference between edit and update are. I changed them all to edit. While I personally think update sounds better, ("update an expired card", "update a mistyped CVV", "update my card's billing address after I moved") we are using the edit terminology everywhere else.

PR #575, opened 2017-04-11
Force Angular 1.6.3 for now due to showErrors ng-class issue with 1.6.4

OzzieOrca:
Looks like this might be a bug. I'm going to see if we get any more feedback on this thread before making bigger changes: https://github.com/angular/angular.js/issues/15905.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "give-web",
   "version": "1.0.0",
   "dependencies": {
-    "angular": "^1.6.3",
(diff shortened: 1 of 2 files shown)


jbedard:
The change will most likely be reverted in 1.6 (https://github.com/angular/angular.js/pull/15958) but I'm hoping to keep it in 1.7. Comments welcome: https://github.com/angular/angular.js/issues/15960

PR #579, opened 2017-04-17
[#EP-1971] Include jcr:title in campaign-page modals.

Omicron7:
https://jira.cru.org/browse/EP-1971

### DIFF
--- a/src/app/productConfig/productConfigModal/productConfig.modal.component.js
+++ b/src/app/productConfig/productConfigModal/productConfig.modal.component.js
@@ -236,14 +236,9 @@ class ProductConfigModalController {
     } );
   }
 
-  displayName() {
-    return this.productData.displayName === this.itemConfig['jcr-title'] ?
-      this.productData.displayName : this.itemConfig['jcr-title'];
-  }
-
   displayId() {
     let value = `#${this.productData.designationNumber}`;
-    if(this.itemConfig['jcr-title'] && this.productData.displayName !== this.itemConfig['jcr-title']) {
+    if(this.productData.displayName !== this.itemConfig['jcr-title'] && this.itemConfig['campaign-page']) {
       value += ` - ${this.productData.displayName}`;
     }
     return value;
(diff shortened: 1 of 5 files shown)


src/app/productConfig/productConfigModal/productConfig.modal.component.js
```diff
@@ -235,6 +235,19 @@ class ProductConfigModalController {
       this.submittingGift = false;
     } );
   }
+
+  displayName() {
+    return this.productData.displayName === this.itemConfig['jcr-title'] ?
+      this.productData.displayName : this.itemConfig['jcr-title'];
```
OzzieOrca: Isn't this the same as just returning `this.itemConfig['jcr-title']`?

src/app/productConfig/productConfigModal/productConfig.modal.component.js
```diff
@@ -235,6 +235,19 @@ class ProductConfigModalController {
...
+  }
+
+  displayId() {
+    let value = `#${this.productData.designationNumber}`;
+    if(this.itemConfig['jcr-title'] && this.productData.displayName !== this.itemConfig['jcr-title']) {
```
OzzieOrca: This condition doesn't make sense to me. Maybe `this.productData.displayName && this.productData.displayName !== this.itemConfig['jcr-title']`? If we have a display name and it isn't duplicate then we show it? The current condition seems like if the `jcr-title` is missing then we don't show this secondary title. Maybe that's what's intended...

src/app/productConfig/productConfigModal/productConfig.modal.component.js
Omicron7: I believe you are correct, jcr-title will always be set and should be used for the displayName.

src/app/productConfig/productConfigModal/productConfig.modal.component.js
Omicron7: In this scenario, if we used jcr-title instead of displayName, then we want to add displayName to the designationNumber line. We use jcr-title in place of displayName when they are different. I added jcr-title check since I didn't know if it would always be set, but I think it should be.

PR #582, opened 2017-04-17
[EP-1976] In text editor modal, include textangular css and add margi…

OzzieOrca:
…n below button groups

https://jira.cru.org/browse/EP-1976


OzzieOrca:
Ya. Not certain exactly how it works but it'll be in the designationEditor and I think it adds a style tag to the `<head>` when run that contains that minified css.

OzzieOrca:
The style-loader ([webpack.config.js#L120](https://github.com/CruGlobal/give-web/blob/master/webpack.config.js#L120)) is what sticks the compiled css into the head.

adammeyer:
👍

PR #591, opened 2017-04-20
Send card number when editing a credit card during checkout

OzzieOrca:
I missed this when making changes for https://jira.cru.org/browse/EP-1962. Previously the card number couldn't be modified.


wrandall22:
Will this also send card type and last four?

OzzieOrca:
Ya. I was picking specific fields before. Now I'm only excluding the CVV. These values come from https://github.com/CruGlobal/give-web/blob/master/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js#L142-L152

OzzieOrca:
I made it more explicit in my test.

OzzieOrca:
I made it more explicit in my test.

PR #595, opened 2017-04-22
Use fakeLocalStorage to keep localStorage and sessionStorage in memor…

OzzieOrca:
…y when the browser doesn’t support them.

https://jira.cru.org/browse/EP-1989

Since this overwrites the implementation of `window.Storage`, this seems to work for sessionStorage also.

I don't think we have any components other than checkout that are using Storage to communicate between different states of the same component so I just included it in that bundle.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,3 +2,4 @@ comment: off
 coverage:
   ignore:
     - "src/app/analytics/*"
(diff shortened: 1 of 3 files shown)


OzzieOrca:
Feel free to merge this if you approve. It would be nice to have it in before the code freeze.

PR #596, opened 2017-04-26
Update give production url

adammeyer:


adammeyer:
Good idea, I'll do that.

OzzieOrca:
I think a search for `.is('production')`, `envService.get`, and `envServiceProvider.get` will find them all. I'll let you evaluate how easy it is to change all of these to make your `production-test` (or whatever you want to call it equal `production` in all of these cases... It might just be easier to change it back later if this is all we are changing.

adammeyer:
Okay, I'll go ahead and merge this for now.  It's an easy change to switch it back on Monday.

PR #599, opened 2017-04-27
Don’t require CVV when not showing CVV input

OzzieOrca:
- Pass null CVV to cru-payments to skip CVV validation in this case

Currently saving payment methods in the profile is broken because it is requiring the CVV. Checkout has no issues.


BobBuford:
Reviewed code, with @OzzieOrca  explaining why some things are the way they are.  Inspected syntax.  All looks good.   👍

PR #603, opened 2017-05-03
sign out on cru.org instead of downgrade

adammeyer:


src/common/components/nav/nav.component.js
```diff
@@ -157,7 +157,7 @@ class NavController{
       keyboard: false,
       size: 'sm'
     });
-    this.sessionService.downgradeToGuest().subscribe(() => {
+    this.sessionService[this.customProfile ? 'signOut' : 'downgradeToGuest']().subscribe(() => {
```
OzzieOrca: What is `customProfile`?

adammeyer: It's just for the nav component.  Used to for cru.org and staff web.  It won't verify cortex role before displaying name etc.

OzzieOrca: Gotcha. The name wasn't immediately clear.

PR #605, opened 2017-05-03
EP-2016 - handle InsufficientFundException error type

mattdrees:
I haven't tested this, but I think this is what is required to nicely handle CruGlobal/give-ep-extensions/pull/475.


mattdrees:
@OzzieOrca can you review? I think this is right, but I can't run the unit tests (without figuring out how to get set up ).

mattdrees:
rebased on master

mattdrees:
Yeah, I figured that is what you are doing. This shouldn't be any different from the other error messages.

Thanks for the review!

PR #606, opened 2017-05-03
Scrub password reset k query param from rollbar

OzzieOrca:
I don't think we want this key in Rollbar. https://github.com/CruGlobal/give-web/blob/master/src/common/components/resetPasswordModal/resetPasswordModal.component.js#L29-L29 It would allow anyone with access to rollbar to potentially reset a user's password. There are a few request urls in Rollbar that contain this.


OzzieOrca:
I suppose analytics would probably also receive these request urls... Idk how big of an issue that is as they should be one time, expiring keys but there still could be a window of time where someone could look up and use that link. I'm not sure how to do that with Adobe Analytics...

PR #607, opened 2017-05-03
Cleanup reset password modal rollbar logs

OzzieOrca:
- Suppress “Attempted to open more than 1 modal” log when reset-password modal is open
- Handle case in session enforcer where modal is not defined because another modal was already open

### DIFF
--- a/src/common/services/session/sessionEnforcer.service.js
+++ b/src/common/services/session/sessionEnforcer.service.js
@@ -102,7 +102,7 @@ function SessionEnforcerService( orderService, sessionService, sessionModalServi
           backdrop: 'static',
           keyboard: false
         } ).result;
-        modal.then( () => {
+        modal && modal.then( () => {
           angular.forEach( enforced, ( enforcer ) => {
             if ( angular.isFunction( enforcer[EnforcerCallbacks.signIn] ) ) enforcer[EnforcerCallbacks.signIn]();
           } );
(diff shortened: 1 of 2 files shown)


src/common/services/session/sessionModal.service.js
```diff
@@ -18,7 +18,9 @@ function SessionModalService( $uibModal, $log, modalStateService, analyticsFacto
         currentModal.dismiss( 'replaced' );
       }
       else {
-        $log.error( 'Attempted to open more than 1 modal' );
+        if(currentModal.type !== 'reset-password'){
+          $log.error( 'Attempted to open more than 1 modal' );
```
OzzieOrca: https://rollbar.com/Cru/give-web/items/833

src/common/services/session/sessionModal.service.js
```diff
@@ -35,6 +37,7 @@ function SessionModalService( $uibModal, $log, modalStateService, analyticsFacto
       }
     }, options );
     currentModal = $uibModal.open( modalOptions );
+    currentModal.type = type;
```
OzzieOrca: Maybe I shouldn't store this on the uibModal object. I suppose I could add another variable.

src/common/services/session/sessionEnforcer.service.js
```diff
@@ -102,7 +102,7 @@ function SessionEnforcerService( orderService, sessionService, sessionModalServi
           backdrop: 'static',
           keyboard: false
         } ).result;
-        modal.then( () => {
+        modal && modal.then( () => {
```
OzzieOrca: https://rollbar.com/Cru/give-web/items/894

src/common/services/session/sessionModal.service.js
Omicron7: Should be ok.

Omicron7:
We may have another underlying issue here where the reset password modal is opened on a page that requires a registered donor. When reset is complete, it may leave them on a page they don't have access to. Maybe reload current page to fire sessionEnforcer again after forgot password is successful?

OzzieOrca:
Merging this and will address that issue in another PR since these changes don't affect that.

PR #608, opened 2017-05-04
Improve session handling with reset password modal

OzzieOrca:
- Refresh page when modal dismissed without updating password
- Sign in immediately after changing password

Based off [this comment](https://github.com/CruGlobal/give-web/pull/607#issuecomment-299034401).

@Omicron7 can you review this functionality? I'll add/fix tests after. Does signing in immediately after a password reset sound like it will work? One issue still is if the user clicks the `Back to Sign In` link and then dismisses the sign in modal, they won't be redirected away from the page. I thought about just tossing that link...

### DIFF
--- a/src/common/components/resetPasswordModal/resetPasswordModal.component.js
+++ b/src/common/components/resetPasswordModal/resetPasswordModal.component.js
@@ -12,8 +12,11 @@ let componentName = 'resetPasswordModal';
 class ResetPasswordModalController {
 
   /* @ngInject */
-  constructor( $location, gettext, sessionService, modalStateService ) {
+  constructor( $window, $location, $log, $timeout, gettext, sessionService, modalStateService ) {
+    this.$window = $window;
     this.$location = $location;
+    this.$log = $log;
+    this.$timeout = $timeout;
     this.gettext = gettext;
     this.sessionService = sessionService;
     this.modalState = modalStateService;
@@ -38,9 +41,11 @@ class ResetPasswordModalController {
   }
 
   $onDestroy() {
-    angular.forEach( ['e', 'k'], ( key ) => {
-      this.$location.search( key, null );
-    } );
+    this.removeQueryParams();
+    if(!this.exitWithoutRefresh){
+      // use $timeout here as workaround to Firefox bug
+      this.$timeout(() => this.$window.location.reload());
+    }
   }
 
   resetPassword() {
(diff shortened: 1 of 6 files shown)


src/common/components/resetPasswordModal/resetPasswordModal.component.js
```diff
@@ -38,9 +40,10 @@ class ResetPasswordModalController {
   }
 
   $onDestroy() {
-    angular.forEach( ['e', 'k'], ( key ) => {
-      this.$location.search( key, null );
-    } );
+    this.removeQueryParams();
+    if(!this.passwordChanged && !this.exitWithoutRefresh){
+      this.$window.location.reload();
```
Omicron7: This will probably need a $timeout added, firefox has a bug whit this.
See https://github.com/CruGlobal/give-web/blob/89489245caf086945ebe2409032db5df4e4ac408/src/common/components/nav/nav.component.js#L149

Omicron7:
I don't think sending the user directly to sign-in will work. The idea with page reload was that it would let the session enforcer fire whatever the page needed, which is usually the user-match workflow, and not specifically just sign in.

Another idea could be to remove the Back To Sign-In from the password reset screen, and add a queue to the session modal service. When a modal closes, success or failure, check the queue and if another modal is queued up, run it. This would allow callbacks to still be fired for each modal in succession.

Third option, probably not the best but it solves the issues, would be to always send the user to the give homepage (in the redirect link sent to cas). Then add a message in the reset success to close return to your previous window to continue.

src/common/components/resetPasswordModal/resetPasswordModal.component.js
```diff
@@ -38,9 +41,11 @@ class ResetPasswordModalController {
   }
 
   $onDestroy() {
-    angular.forEach( ['e', 'k'], ( key ) => {
-      this.$location.search( key, null );
-    } );
+    this.removeQueryParams();
+    if(!this.exitWithoutRefresh){
+      // use $timeout here as workaround to Firefox bug
+      this.$timeout(() => this.$window.location.reload());
```
OzzieOrca: This now refreshes the page when the reset modal is closed except when `Back to Sign In` is clicked. Back to sign won't refresh the page if the user signs in there. Maybe that should just be a cancel link that closes the modal and refreshes the page too...

src/common/services/session/session.service.js
```diff
@@ -135,7 +135,7 @@ function session( $cookies, $rootScope, $http, $timeout, envService ) {
           resetKey: resetKey
         }
       } ) )
-      .map( ( response ) => response.data );
+      .mergeMap( () => signIn(email, password) );
```
OzzieOrca: Since we have all the data needed when the password is reset, why not just sign the user in immediately? After the reset modal is closed which performs a page refresh, the user should be good to go. If this isn't desired, this could be removed and we could rely on the page refresh and then the session enforcer for sign in.

Omicron7:
👍 Looks good.

PR #611, opened 2017-05-04
In product config, handle attempted opening of second modal more grac…

OzzieOrca:
…efully

This sort of fixes some of the subsequent errors sent to rollbar related to https://jira.cru.org/browse/EP-2021 ([can't call rendered of undefined](https://rollbar.com/Cru/give-web/items/999/)) but doesn't address the main issues. My assumption with that issue was that some users were unable to open the product config modal but that could be wrong. If multiple buttons are clicked before the first one finishes loading, this code will get run and prevent JS errors.

### DIFF
--- a/src/app/cart/cart.component.js
+++ b/src/app/cart/cart.component.js
@@ -78,9 +78,9 @@ class CartController {
   }
 
   editItem( item ) {
-    this.productModalService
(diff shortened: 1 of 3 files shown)


OzzieOrca:
Found another case and added a commit for it https://rollbar.com/Cru/give-web/items/1011/

PR #614, opened 2017-05-04
Show duplicate recurring gift error to user in step 4 of editRecurrin…

OzzieOrca:
…gGiftsModal

I'm not sure what is causing this but we should give the user a better error message. I suppose they could just be adding a gift twice with the suggested gift or search step. Are there any other ways to receive this error? @mattdrees @wrandall22 do you have any thoughts? We've got about 8 people getting this.

https://rollbar.com/Cru/give-web/items/?query=%24log%3A%20%22Error%20updating%2Fadding%20recurring%20gifts


mattdrees:
That error was added by a [PR](https://github.com/CruGlobal/give-ep-extensions/pull/466) a few weeks ago.

It's hard to guess why this is happening so often. I think we need to search through the log files to see if the recurring gift in question is in the list that is being sent to the front end, or if it's being filtered out of the list but still preventing a duplicate gift from being added.

wrandall22:
I may have figured out the problem, I think the filtering is a little off.

OzzieOrca:
https://github.com/CruGlobal/give-web/pull/619 is fixing another case of this but for the restart modal.

PR #617, opened 2017-05-05
Fix IE disabled naming issue with address form disabling

OzzieOrca:


OzzieOrca:
@Omicron7 or @adammeyer could you deploy? I still don't think I have access but I'll check again.

PR #631, opened 2017-05-24
Hide Give a New Gift btn for designations that are no longer active

OzzieOrca:
- Handle undefined changes in historicalView component

@wrandall22 I had forgotten about our [helpscout conversation](https://secure.helpscout.net/conversation/360290266/131713/?folderId=610613). Thanks for following up.

Hopefully this should prevent donors from giving to designations that are no longer active.

### DIFF
--- a/src/app/profile/yourGiving/historicalView/historicalView.component.js
+++ b/src/app/profile/yourGiving/historicalView/historicalView.component.js
@@ -14,7 +14,7 @@ class HistoricalView {
   }
 
   $onChanges(changes) {
-    if ( changes.year || changes.month || changes.reload && changes.reload.currentValue === true) {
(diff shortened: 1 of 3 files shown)


src/app/profile/yourGiving/historicalView/historicalView.component.js
```diff
@@ -14,7 +14,7 @@ class HistoricalView {
   }
 
   $onChanges(changes) {
-    if ( changes.year || changes.month || changes.reload && changes.reload.currentValue === true) {
+    if ( changes && (changes.year || changes.month || changes.reload && changes.reload.currentValue === true)) {
```
OzzieOrca: This should fix these: https://rollbar.com/Cru/give-web/items/?query=year

PR #635, opened 2017-06-02
Upgrade webpack-dev-server and fix public path which is now validated…

OzzieOrca:
… in 2.4.3

Bill ran into this I'm assuming because he used npm and npm ignores the yarn lock file and grabbed the newest version of `webpack-dev-server` for him. `localhost.cru.org` was giving a `Invalid Host header` error.

https://github.com/webpack/webpack-dev-server/releases/tag/v2.4.3

https://github.com/webpack/webpack-dev-server/issues/882

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -137,7 +137,8 @@ module.exports = env => {
         '/etc': aemDomain,
         '/designations': aemDomain
       },
-      port: 9000
+      port: 9000,
+      public: "localhost.cru.org:9000"
     }
   };
 };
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -137,7 +137,8 @@ module.exports = env => {
         '/etc': aemDomain,
         '/designations': aemDomain
       },
-      port: 9000
+      port: 9000,
+      public: "localhost.cru.org:9000"
```
wrandall22: This is the value for localhost, but wouldn't be for other environments. Not sure if that matters.

OzzieOrca: This dev server is only for development. S3 serves the js everywhere else. This should be good.

PR #637, opened 2017-06-06
limit cart hash to 20 characters

adammeyer:


src/app/analytics/analytics.factory.js
```diff
@@ -14,7 +14,7 @@ function analyticsFactory($window, $timeout, sessionService) {
         // Instantiate cart data layer
         $window.digitalData.cart = {
           id: cartData.id,
-          hash: cartData.id ? md5(cartData.id).toString() : null,
+          hash: cartData.id ? md5(cartData.id).toString().substring(0, 20) : null,
```
OzzieOrca: How about `sha3(cartData.id, { outputLength: 80 }).toString()`? I feel like you might have more hash collisions if you truncate md5. I was looking at https://crypto.stackexchange.com/questions/3558/are-there-hash-algorithms-with-variable-length-output which suggested SHA3. `outputLength` specifies bits so 80 bits * 1 hex char / 4 bits = 20 hex chars.

https://www.webpackbin.com/bins/-KlykkvDQVIMFh_3DQgS

adammeyer:
@OzzieOrca Thanks, take a look at latest commit.

src/app/analytics/analytics.factory.js
```diff
@@ -12,9 +12,10 @@ function analyticsFactory($window, $timeout, sessionService) {
         var item, donationType;
 
         // Instantiate cart data layer
+        const hash = sha3(cartData.id, { outputLength: 20*4 });
```
OzzieOrca: It might be better to hard code `80` and add a comment as to why we need this to output a string with length of 20.

PR #641, opened 2017-06-07
Refactor productConfig form into separate component for use outside t…

OzzieOrca:
…he productConfigModal

I'm starting to work with Josh on using components from give-web to create a new branded checkout flow for other ministries to use. I needed a productConfig form that wasn't tied to a modal and could be dropped on a page somewhere.

The biggest changes are:
1. Creating a `productConfigForm`component that `productConfigModal` uses.
2. Moving the data loading logic out of the `$uibModal`'s resolve and into the `productConfigForm` and moving the loading/error logic with it.
3. Pulling anything relating to query params out of the `productConfigForm` component and into the `productConfigModal` component.
4. Moving the buttons out of the form and adding bindings to make the 2 components interact.


src/app/productConfig/productConfigModal/productConfig.modal.component.js
```diff
@@ -35,69 +20,57 @@ export const giveGiftParams = {
...
       return;
     }
 
-    let params = this.$location.search();
-
+    // Add query params for current modal
     this.modalStateService.name( 'give-gift' );
-    this.$location.search( giveGiftParams.designation, this.productData.code );
-    this.$location.search( giveGiftParams.campaignPage, this.itemConfig['campaign-page'] );
+    this.updateQueryParam( giveGiftParams.designation, this.code ); // TODO: can this.code change? Do we need to set this again if productData comes back with a different code?
```
OzzieOrca: This used to rely on a request for `productData`. Is there any reason the productData request would return a different code than the one it was called with? Is this how redirecting designations works? Do we care if the query param stays out of date? Even if it is redirecting, it seems like if the old query param was kept, it would just perform the redirect again to get new data on every page load.

adammeyer: I believe redirecting designations happen on the AEM side.  So I don't think the code will change.  @Omicron7

OzzieOrca: K. I was just making sure there wasn't some specific reason it wasn't using `this.code` in the first place.

OzzieOrca:
@Omicron7 Could you look through this?

I forgot about getting this merged. @adammeyer you fine with me merging it? I need to rebase it.

adammeyer:
I'm fine with you merging it, but I think it's a good idea for @Omicron7 to take a look first.

PR #649, opened 2017-06-20
Fire cart open event only on first item

adammeyer:


src/common/components/nav/navCart/navCart.component.js
```diff
@@ -43,7 +43,9 @@ class NavCartController{
           this.loading = false;
           this.hasItems = !isEmpty(this.cartData.items);
           this.analyticsFactory.buildProductVar(data);
-          this.analyticsFactory.setEvent('cart open');
+          if(this.cartData.items.length === 1){
+            this.analyticsFactory.setEvent('cart open');
```
OzzieOrca: This will get fired if the cart has 2 items and one is deleted and if the cart has one item and it is edited. Also it will get triggered on a page refresh/navigation.

Is the intent of this event to only capture when a user first starts using their cart/starts an order?

There's a possibility adding this event here will work: https://github.com/CruGlobal/give-web/blob/master/src/common/services/api/cart.service.js#L106-L106. We're starting a new session for guest users. We'd have to figure out how to handle non guest users if we want to fire the event here.

adammeyer: Yes, that's the intent.  I checked there, but it will require an extra request for non public users.  I don't want to add that.  I'll check with Brenton.

adammeyer:
@OzzieOrca Check out the latest commit.  Not fantastic, but I think it will do the trick.

PR #650, opened 2017-06-22
Update CCP Backup Keys

OzzieOrca:
Used for bank accounts

### DIFF
--- a/src/common/app.constants.js
+++ b/src/common/app.constants.js
@@ -1,24 +1,24 @@
 // These CCP keys are used as backup keys if a network request for one fails.
 // They should be updated annually (around mid-April) as CCP rotates encryption keys.
-// Last updated 08/11/2016
+// Last updated 06/22/2017
 const ccpKey = '-----BEGIN PUBLIC KEY-----' +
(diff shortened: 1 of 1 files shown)


mattdrees:
is that what caused https://rollbar.com/Cru/give-ep/items/3972/?

OzzieOrca:
Yes. I was testing branded checkout on http://cru-givedev.s3-website-us-east-1.amazonaws.com/ which should be hitting the staging env. I should have updated them earlier but thought I didn't have access to the keys, which was dumb since the app can request them at will.

wrandall22:
I'm curious why we are hard-coding these keys instead of looking them up from the CCP endpoint. We'll be doing a PR like this every time they change.

mattdrees:
They're only hardcoded as fallbacks, for when a runtime request to CCP fails. 
So when the keys changed in April, the give site worked just fine, except when requests to CCP failed. I'm guessing that is rare, though I guess not so rare that Scotty didn't notice it.

OzzieOrca:
Ya. I'm totally sure having a backup is necessary but that's what I did. I only noticed it when I accidentally was pointed at the production api and had a CORS issue fetching a key since I was on another domain.

PR #652, opened 2017-06-23
allow for custom signOutPath url

adammeyer:


adammeyer:
@OzzieOrca StaffWeb https://jira.cru.org/browse/SWM-29

PR #655, opened 2017-06-26
Mask CVV in checkout step 3 error log

OzzieOrca:


Omicron7:
@OzzieOrca Can we replace it with XXX, might be nice in the future to know it was in the request, and not trying to debug why it wasn't passed along in the request.

OzzieOrca:
I made that change. Good idea.

PR #656, opened 2017-06-26
Sandboxed Cru Nav

adammeyer:


adammeyer:
@OzzieOrca Let me know your thoughts.  I tried many things and this was the only way I could get it to produce the desired output.

PR #658, opened 2017-06-27
disable re-submission of gift if request was dropped

adammeyer:
Regarding: https://secure.helpscout.net/conversation/385447846/143173


src/app/checkout/step-3/step-3.component.js
```diff
@@ -99,6 +99,7 @@ class Step3Controller{
 
   submitOrder(){
     delete this.submissionError;
+    delete this.submissionErrorCode;
```
OzzieOrca: Looks like this has an old/inconsistent variable name

PR #660, opened 2017-06-28
Branded checkout

OzzieOrca:
http://cru-givedev.s3-website-us-east-1.amazonaws.com/


OzzieOrca:
Can I merge this to get it on staging? I suppose there is potential I broke a component in the existing app so we should test it some. I don't want to have it blocking the path to production though if something else needs to be deployed.

Omicron7:
I'm fine merging to staging. I agree we will want a bunch of testing before prod.

PR #663, opened 2017-07-10
Move loading css into component

OzzieOrca:
I did this cleanup as I was playing with CSS Modules in Angular 1.X and figured it would be good to pull it out and get it merged regardless of how we do CSS encapsulation.

### DIFF
--- a/src/common/components/loading/loading.tpl.html
+++ b/src/common/components/loading/loading.tpl.html
@@ -1,16 +1,16 @@
 <div ng-class="{
-        'loading-centered loading-overlay' : $ctrl.type === 'overlay',
-        'loading-centered' : $ctrl.type === 'centered'
+        'loading-centered': $ctrl.type === 'centered' || $ctrl.type === 'overlay',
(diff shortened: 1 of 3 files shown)


src/common/components/loading/loading.tpl.html
```diff
@@ -1,16 +1,16 @@
...
   <div>
     <span ng-class="{ 'mr-' : $ctrl.inline === 'true', 'center-block' : $ctrl.inline !== 'true'  }">
       <ng-transclude></ng-transclude>
     </span>
-    <span class="loading-wrap">
-      <span class="loading loading1"></span>
-      <span class="loading loading2"></span>
-      <span class="loading loading3"></span>
-      <span class="loading loading4"></span>
+    <span ng-class="loading-wrap">
```
adammeyer: Why did this change to ng-class?

OzzieOrca: Good catch. I guess I left that from my experiment with CSS Modules where you had to import the renamed class name.

PR #667, opened 2017-07-18
Branded checkout callback

OzzieOrca:
If you can think of any cleaner ways to call a user defined function from an angular output binding, let me know. The best I got was `on-order-completed="$event.$window.onOrderCompleted($event.purchase)"`.

The legacy branded checkout supported this and Josh said it was probably still desired.


Omicron7:
Does it need to be a callback, or could it just fire a javascript event that other js code could listen for?

OzzieOrca:
I suppose either. I feel like this is more explicit and doesn't add stuff to the window if it isn't used. This would let it be used inside an Angular app cleanly if that ever happens.

OzzieOrca:
I'm not even sure if this is being used currently. Josh was pretty sure someone would want it since it was added in the old app. He was waiting on someone else for direction but told me to work on it since they most likely wanted it.

PR #668, opened 2017-07-18
Convert suggestedAmounts ordering to a number.

Omicron7:
This fixes helpscout issue where suggestedAmounts with more than 10 amounts were getting sorted by string instead of numerically.
( ex: integer - 1, 2, 3, 7, 10, 11, 12, 20 ... vs string - 1, 10, 11, 12, 2, 20, 3, 7 ...)


OzzieOrca:
https://github.com/CruGlobal/give-web/blob/bfa7907e99c6b94da5ccf1d4426dbc2b754be96a/src/common/services/api/designations.service.spec.js#L137-L138 needs to be updated right?

adammeyer:
👍 Thanks Brian

PR #673, opened 2017-07-31
fix webpack proxy origin

adammeyer:
Without `changeOrigin`, it was setting localhost.cru.org as the host header causing the AEM server to not find the site.


adammeyer:
@OzzieOrca Makes sense, the other routes are not site specific, so they most likely work with any host header, but just to be safe I changed them all.

PR #676, opened 2017-08-07
Add angularLoaded and checkout step analytics events to branded checkout

OzzieOrca:
There isn't a thank you event right? I think that's supposed to be a new page view. We do [set the purchase number](https://github.com/CruGlobal/give-web/blob/master/src/app/thankYou/summary/thankYouSummary.component.js#L84) so maybe that's enough.

### DIFF
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -17,13 +17,16 @@ let componentName = 'brandedCheckout';
 class BrandedCheckoutController {
 
   /* @ngInject */
-  constructor($window) {
+  constructor($window, analyticsFactory) {
     this.$window = $window;
+    this.analyticsFactory = analyticsFactory;
   }
 
   $onInit() {
     this.checkoutStep = 'giftContactPayment';
     this.formatDonorDetails();
+    this.analyticsFactory.pageLoaded(true);
+    this.fireAnalyticsEvents('contact', 'payment');
   }
 
   formatDonorDetails(){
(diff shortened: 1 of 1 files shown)


src/app/branded/branded-checkout.component.js
```diff
@@ -59,6 +64,13 @@ class BrandedCheckoutController {
   onThankYouPurchaseLoaded(purchase) {
     this.onOrderCompleted({$event: {$window: this.$window, purchase: changeCaseObject.camelCase(pick(purchase, ['donorDetails', 'lineItems']))}});
   }
+
+  fireAnalyticsEvents(...checkoutSteps){
+    checkoutSteps.forEach(checkoutStep => {
+      this.analyticsFactory.setEvent('checkout step ' + checkoutStep);
+      this.analyticsFactory.track('aa-checkout-step-' + checkoutStep);
```
OzzieOrca: Do we need both of these still? I know Brenton was switching between them for some reason.

adammeyer: @brenton05 still needed?

adammeyer:
I'm not sure if we should include analytics on this.  The adobe dtm script likely won't be included on third party sites.  Thoughts?

OzzieOrca:
Josh seemed to imply that we would want it to be. Idk if/where @brenton05 wants to receive analytics for branded checkout. I guess that could use some more discussion.

adammeyer:
@OzzieOrca Talked with Brenton today.  We want that in.  He will put DTM scripts on the sites individually.

OzzieOrca:
So good to merge?

adammeyer:
yes

PR #678, opened 2017-08-11
Add analytics to cruNav bundle and call pageLoaded when nav is runnin…

OzzieOrca:
…g on its own

https://jira.cru.org/browse/CRUORG-115

Let me know your thoughts on making sure this works safely on cru.org.

Do we need to think about any non cru.org sites when doing this? I'm not sure where else this is running.

### DIFF
--- a/src/common/components/nav/nav.bundle.js
+++ b/src/common/components/nav/nav.bundle.js
@@ -7,11 +7,11 @@
   // angular.js to load itself into
   const angular = (window.angular = {});
 
-  require('./nav.component');
+  const nav = require('./nav.component').default;
 
   //Manually bootstrap cruNav
   angular.element(document).ready(function () {
-    angular.bootstrap(document.getElementsByTagName("cru-nav")[0], ["cruNav"]);
+    angular.bootstrap(document.getElementsByTagName("cru-nav")[0], [nav.name]);
     // restore the old angular version
     window.angular = existingWindowDotAngular;
   });
--- a/src/common/components/nav/nav.component.js
+++ b/src/common/components/nav/nav.component.js
@@ -11,6 +11,7 @@ import 'rxjs/add/operator/retry';
 import loading from 'common/components/loading/loading.component';
 import sessionService, {SignOutEvent} from 'common/services/session/session.service';
 import sessionModalService from 'common/services/session/sessionModal.service';
+import analyticsRun from 'app/analytics/analytics.run';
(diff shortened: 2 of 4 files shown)


src/common/components/nav/nav.component.js
```diff
@@ -88,6 +90,12 @@ class NavController{
     // Register signedOut event on child scope
     // this basically sets the listener at a lower priority, allowing $rootScope listeners first chance to respond
     this.$rootScope.$new(true, this.$rootScope).$on(SignOutEvent, (event) => this.signedOut(event) );
+
+    try {
+      angular.module('common');
+    } catch(err) {
+      this.analyticsFactory.pageLoaded(true); // Only call pageLoaded when common module doesn't exist, indicating that only the nav bundle is running and we aren't on the give site
```
OzzieOrca: I could do your `hostname.indexOf('give')` check here if that sounds safer/better.

adammeyer: I agree, looks good.

src/common/components/nav/nav.bundle.js
```diff
@@ -7,11 +7,11 @@
   // angular.js to load itself into
   const angular = (window.angular = {});
 
-  require('./nav.component');
+  const nav = require('./nav.component');
```
adammeyer: Does this work?  I believe the nav.component source needs to be put into the bundle.

OzzieOrca: Thanks for questioning this. It needed a `.default` on the end of it. `require` returns an object with all the named exports, including `default` as a key. I guess I'm used to working with `import`.

adammeyer: Will it include the nav component in the bundle?  Or just assigned it to the constant?

OzzieOrca: Ya it works. I tested it after your comment. When requiring the component it runs all the nav code and all the calls to `angular.xxx` add modules/components/whatever to the global angular window object which gets bootstrapped with the code here. This is the same as the `import`s in the rest of the app. This is no different than before other than I am now also using the default export to get the module name.

I guess I really didn't need to touch this file but at first I thought I might need to import the analytics module here. I guess this just makes it more consistent with the other angular module imports.

adammeyer: okay, thanks for the info.

PR #681, opened 2017-08-15
recurring start month selection

adammeyer:


adammeyer:
@wrandall22 FYI, you can use this brand for testing if you would like.

wrandall22:
Okay, thanks!

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -147,18 +147,30 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
         </h4>
         <div class="panel-body pt0">
           <div class="row">
+            <div class="col-sm-3">
+              <div class="form-group">
+                <label translate>Month</label>
+                <div class="form-group">
+                  <select class="form-control form-control-subtle"
+                          ng-model="$ctrl.itemConfig['recurring-start-month']"
+                          ng-options="m.value as m.text for m in $ctrl.possibleTransactionMonths($ctrl.nextDrawDate)"
```
adammeyer: @OzzieOrca I'm getting a digest loop on this line, any suggestions?  I tried track by, but that changes the value of the options.

OzzieOrca: I saw this but didn't reply immediately and forgot... I think this is different than the one below because you are returning an object for each item.

Doing `m.value as m.text for m in $ctrl.possibleTransactionMonths($ctrl.nextDrawDate) track by m.value` seems to work although the initial value seems messed up. Idk if the value is right (outputs `value="08"` instead of `value="string:08"`).

I think the other option would be to precompute it and save the array so you aren't calling a function that returns different object references on every loop.

Or maybe use https://lodash.com/docs/4.17.4#memoize on the function to cache the values.

PR #683, opened 2017-08-28
use start date from BE for cart

adammeyer:
if next-draw-date is in past, use current date (mainly for use on stage)

### DIFF
--- a/src/app/cart/cart.tpl.html
+++ b/src/app/cart/cart.tpl.html
@@ -47,7 +47,7 @@
                       <span class="giftsum-frequency giftsum-title">{{i.frequency}}</span>
                       <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'">
                           <span translate>Starts on:</span>
-                          {{i.giftStartDate.format('ll')}}
+                          {{i.giftStartDate | date}}
                         </span>
                     </td>
                     <td class="td-amount">
--- a/src/app/cart/cart.tpl.html
+++ b/src/app/cart/cart.tpl.html
@@ -47,7 +47,7 @@
                       <span class="giftsum-frequency giftsum-title">{{i.frequency}}</span>
                       <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'">
                           <span translate>Starts on:</span>
-                          {{i.giftStartDate | date}}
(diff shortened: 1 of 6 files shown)


adammeyer:
@OzzieOrca Wasn't sure why the BE start date wasn't being used.  Is it alright to use it?

adammeyer:
@wrandall22 @mattdrees Can you answer that question?

wrandall22:
It looks like it gets calculated each time the line item is requested. However, it does not take next draw date into account until the asynchronous submission.

OzzieOrca:
Does it take the current date into account? Just curious how it's calculated. Does it use a cached next draw date?

wrandall22:
It looks at current date, doesn't think about next draw date at all. Basically:
````
LocalDate tentativeStartDate = today.withMonth(chosenMonth).withDay(chosenDay);
if (tentativeStartDate.before(today)) {
    return tenativeStartDate.plusYears(1);
}
return tentativeStartDate;
````

adammeyer:
@wrandall22 So when the gift is submitted (completed checkout), that's when next draw date is checked?

wrandall22:
Right.

adammeyer:
@wrandall22 @OzzieOrca I'm fine with using the EP date.  Any objections?

wrandall22:
It will be inaccurate at times; could confuse the donor if they bother to remember what the page said.

wrandall22:
We could try to use the next draw date, but it will make the checkout process a tiny bit slower.

OzzieOrca:
I think that's fine. Maybe I'm thinking to much about corner cases but my last question is how does the next draw date handle weekends? Does it jump forward a couple days?

wrandall22:
Next draw date being not what the donor might expect is not really an edge case, it's probably very frequent.  However, we just use what Siebel gives us for the next draw date. Not sure how Siebel handles weekends, but it's a pretty sturdy process.

adammeyer:
@wrandall22 If a gift is added with a start date of August 28 and the user comes back to their cart tomorrow, would it show August 28, 2018?

wrandall22:
@adammeyer Yes, it looks like it would.

adammeyer:
okay, I'll go ahead and have the FE calculate the date

PR #684, opened 2017-08-29
warn for far future gifts

adammeyer:


src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -71,13 +71,13 @@ class ProductConfigFormController {
...
 
-    if(inRange(parseInt(this.itemConfig['recurring-day-of-month'], 10), 1, 28)){
+    if(inRange(parseInt(this.itemConfig['recurring-day-of-month'], 10), 1, 29)){
       this.itemConfig['recurring-day-of-month'] = padStart(this.itemConfig['recurring-day-of-month'], 2, '0');
     }else{
       delete this.itemConfig['recurring-day-of-month'];
     }
 
-    if(inRange(parseInt(this.itemConfig['recurring-start-month'], 10), 1, 12)){
+    if(inRange(parseInt(this.itemConfig['recurring-start-month'], 10), 1, 13)){
```
OzzieOrca: Nice catch.

src/app/cart/cart.tpl.html
```diff
@@ -48,6 +48,8 @@ <h3 class="panel-name" translate>Your Gift Cart</h3>
                       <span class="giftsum-date giftsum-detail" ng-if="i.frequency !== 'Single'">
                           <span translate>Starts on:</span>
                           {{i.giftStartDate.format('ll')}}
+                          <i class="fa fa-warning" ng-if="i.giftStartDateDaysFromNow > 300"
```
wrandall22: This is 300, but the step-3.tpl is 330.

src/app/checkout/step-3/step-3.component.js
```diff
@@ -86,6 +93,22 @@ class Step3Controller{
...
+      this.nextDrawDate = nextDrawDate;
+    });
+  }
+
+  updateGiftStartMonth(item, month){
+    item.config['recurring-start-month'] = month;
+
+    this.cartData = null;
+    this.cartService.editItem( item.uri, item.productUri, item.config ).subscribe( () => {
+      this.$scope.$emit( cartUpdatedEvent );
```
OzzieOrca: Do we want the nav cart to open? Is that all this does? If you keep this I think you need to create a spy. `spyOn(self.controller.$scope, 'emit');`

adammeyer: No, I'll remove it.

PR #691, opened 2017-10-11
Drop rollbar errors on network or authorization failures

OzzieOrca:
- Make JSON sent to rollbar prettier

I should have done this a long time ago... `args[1]` is kinda a magic number but it seems like our logging usually has the response in the second position.

### DIFF
--- a/src/common/rollbar.config.js
+++ b/src/common/rollbar.config.js
@@ -50,10 +50,14 @@ function rollbarConfig(envServiceProvider, $provide) {
           // Parse the exception to get the stack
           stackFramesPromise = stacktrace.fromError(args[0], {offline: true});
         }else{
+          if (args[1] && (args[1].status === -1 || args[1].status === 401)) {
+            return; // Drop browser network errors and unauthorized api errors due to expired tokens
+          }
+
           // Join $log arguments
           message = args
-            .map(arg => angular.toJson(arg))
-            .join(', ');
+            .map(arg => angular.toJson(arg, true))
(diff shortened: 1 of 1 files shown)


src/common/rollbar.config.js
```diff
@@ -50,10 +50,14 @@ function rollbarConfig(envServiceProvider, $provide) {
           // Parse the exception to get the stack
           stackFramesPromise = stacktrace.fromError(args[0], {offline: true});
         }else{
+          if (args[1] && (args[1].status === -1 || args[1].status === 401)) {
```
adammeyer: Do we want to drop browser network errors?  Sometimes that is helpful when responding to tickets, just to know it was a connection issue for sure.  I haven't gotten those tickets in a while though, so I'm open to removing them.

OzzieOrca: If you've found them useful, we can keep them. Spencer was complaining that there were too many errors in the missionhub-web project and a bunch were network errors. I figured it would be good to do the same thing here. There are a lot of them. I know the rollbar pricing plans are based on how many errors rollbar receives.

How have you used them? Looking up a specific person when they report an issue and see if some network issues are there? I could see that being useful.

adammeyer: Yes, that's what I have used them for in the past, but haven't in a while.

Omicron7:
I think we are safe not logging these. It adds a lot of white noise. We can enable it in the future again if we find specific instances where it would of been beneficial for debugging.

PR #697, opened 2017-10-17
double nav cart fixes

adammeyer:
Looking for some input on this.  I discovered the auto close on "outsideClick" does not work if both of the nav cart dropdowns are using the same `is-open` variable.  Giving them each a different variable solved that issue.  But now only the correct (visible) dropdown should open when a new gift is added.

What I have works, but am looking for some input.

Another idea I had was to emit an event when the sub navigation is locked (via subNav.directive) and use that to remove the cart div from the DOM when it's not visible.  I believe in that case I could keep using the same variable for `is-open` on both.

### DIFF
--- a/src/common/components/nav/nav.component.js
+++ b/src/common/components/nav/nav.component.js
@@ -23,17 +23,21 @@ import mobileTemplate from './mobileNav.tpl.html';
 import desktopTemplate from './desktopNav.tpl.html';
 import signOutTemplate from './signOut.modal.tpl.html';
 
+export let subNavLockEvent = 'subNavLock';
+export let subNavUnlockEvent = 'subNavUnlock';
+
 let componentName = 'cruNav';
 
 class NavController{
 
   /* @ngInject */
-  constructor($log, $rootScope, $http, $document, $window, $uibModal, $timeout, envService, sessionService, sessionModalService){
+  constructor($log, $rootScope, $scope, $http, $document, $window, $uibModal, $timeout, envService, sessionService, sessionModalService){
     this.$log = $log;
     this.$http = $http;
     this.$document = $document;
     this.$uibModal = $uibModal;
     this.$window = $window;
+    this.$scope = $scope;
     this.$rootScope = $rootScope;
     this.$timeout = $timeout;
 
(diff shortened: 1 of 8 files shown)


src/common/components/nav/subNav.tpl.html
```diff
@@ -30,7 +30,7 @@
     </ul>
 
     <div class="cart">
-      <div uib-dropdown on-toggle="$ctrl.cartOpened()" is-open="$ctrl.cartOpen" auto-close="outsideClick">
+      <div uib-dropdown on-toggle="$ctrl.cartOpened()" is-open="$ctrl.cartOpenSubNav" auto-close="outsideClick">
```
OzzieOrca: Could you still open both cart dropdowns and use `on-toggle` to control the `is-open` value of the second one? Basically have the secondary one mirror the primary one. Idk if that would solve all the problems but it seems like that would keep you from having to figure out which one is currently opened.

I was looking at the code and it seems like having 2 `is-open`s probably toggles it twice. At least it seems like something is calling a close function and it deregisters the outsideClick handler.

adammeyer: I tried that, but `$ctrl.cartOpened()` is fired causing a loop.

OzzieOrca: Could you bypass the emit event call? Or only call that on the second one? Idk...

adammeyer:
@OzzieOrca I updated the PR to use $emit events instead.  I like this better.  What are your thoughts?

OzzieOrca:
I added a commit to fix the tests. Hopefully it does what you want. I changed the `scrollY` value from 1 to 0 for the first test to get it to work. Idk if that is the right behavior.

I spied on `$window.addEventListener` to test that it was called and to get a reference to the inner anonymous function that I could call on demand. I didn't try to actually simulate a browser scroll.

src/common/components/nav/nav.component.js
```diff
@@ -85,6 +89,8 @@ class NavController{
     this.subscription = this.sessionService.sessionSubject.subscribe( () => this.sessionChanged() );
 
     this.$rootScope.$on(giftAddedEvent, () => this.giftAddedToCart() );
+    this.$rootScope.$on(subNavLockEvent, () => this.$scope.$apply(() => this.subNavLocked = true ));
+    this.$rootScope.$on(subNavUnlockEvent, () => this.$scope.$apply(() => this.subNavLocked = false ));
```
OzzieOrca: Is this `$apply` needed?

adammeyer: Yes

OzzieOrca: K that's fine

PR #698, opened 2017-10-18
Use CSS to fix position desktop icons to sub nav instead of seperate …

adammeyer:
…DOM elements

Add "Sign In" to sub nav

@OzzieOrca @Omicron7 Because "Sign In" needed to be added to sub nav as well, I thought using CSS to add icons to the sub nav was a better solution.  So two dropdown nav carts are no longer needed. :)

### DIFF
--- a/src/assets/scss/_desktop-sub-navigation.scss
+++ b/src/assets/scss/_desktop-sub-navigation.scss
@@ -78,49 +78,6 @@
     }
   }
 
-  .cart {
-    position: absolute;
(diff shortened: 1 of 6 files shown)


OzzieOrca:
Seems like a good idea. How should we review/test it? I was going to run it locally but I'm not sure how to make the subnav show up without digging into the code.

adammeyer:
In nav.component.js, change live 70 from 'give' to 'localhost'.

PR #701, opened 2017-10-31
Add TSYS env attribute to branded checkout to change the manifest url

OzzieOrca:
From our discussions about StoryRunners on hipchat and a seeing the [helpscout ticket](https://secure.helpscout.net/conversation/454114224/172768/?folderId=320673) that is 10 days old, I figured I would just get this done. We needed a way of letting EP know which device id to use for the manifest generation so we can whitelist other domains with TSYS.

Let me know if you have any naming suggestions or ways to improve the docs. I made a `default` string to try and make it clearer in the basic example that this needs to be changed.

@mattdrees is https://give-stage2.cru.org/cortex/tsys/manifest/test-env the right url to hit? I was getting a 404 presumably because that `accountIdentifier` (from your `/tsys/manifest/{accountIdentifier}` endpoint in hipchat) doesn't exist or this hasn't been fully implemented on the EP side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,17 +30,14 @@ This currently includes all cru.org styling from cru.scss.
 
 #### Branded checkout basic example
 
-Add the following code to your page where appropriate. You must change the value of the `code` and `tsys-env` attributes:
-- `code` must be changed to the designation number you want users to give to.
-- `tsys-env` must be changed to match your environment. It is provided by DPS when adding your domain to the TSYS whitelist.
+Add the following code to your page where appropriate. You must change the value of `designation-number` to the designation number you want users to give to.
 ```html
 <link rel="stylesheet" href="https://give-static.cru.org/branded-checkout.min.css">
 <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css">
 
 <branded-checkout
     ng-app="brandedCheckout"
-    code="0763355"
-    tsys-env="cru">
+    designation-number="0763355">
 </branded-checkout>
 
 <script src="https://give-static.cru.org/branded-checkout.js"></script>
@@ -55,13 +52,13 @@ Add the following code to your page where appropriate:
 
 <branded-checkout
     ng-app="brandedCheckout"
-    code="<designation number>"
-    tsys-env="<tsys-env>"
+    designation-number="<designation number>"
     campaign-code="<campaign code>"
     amount="<amount>"
     frequency="<frequency>"
     day="<day>"
     donor-details="<donor details object>"
+    tsys-device="<tsys-device>"
     on-order-completed="$event.$window.onOrderCompleted($event.purchase)">
 </branded-checkout>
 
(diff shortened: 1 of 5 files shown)


mattdrees:
Try cru instead of test-env

On Oct 30, 2017 6:28 PM, "Scotty Waggoner" <notifications@github.com> wrote:

@OzzieOrca <https://github.com/ozzieorca> requested your review on:
CruGlobal/give-web#701 <https://github.com/CruGlobal/give-web/pull/701> Add
TSYS env attribute to branded checkout to change the manifest url.

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/CruGlobal/give-web/pull/701#event-1317832211>, or mute
the thread

OzzieOrca:
Ya that works. Should I suggest that as the default in the readme? Either way the ministry will need to change it.

mattdrees:
The cru manifest will only work on give.cru.org. the readme should say the
ministry will need to get a device set up for each domain.

On Oct 30, 2017 7:13 PM, "Scotty Waggoner" <notifications@github.com> wrote:

Ya that works. Should I suggest that as the default in the readme? Either
way the ministry will need to change it.

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/CruGlobal/give-web/pull/701#issuecomment-340630569>,
or mute
the thread

OzzieOrca:
Ya. I was just trying to make my example explicit since the JS errors they were experiencing were that they had copied and pasted my code verbatim, not changed any of the placeholder values, and expected it to work. I guess I'll add an inline comment.

README.md
```diff
@@ -30,14 +30,17 @@ This currently includes all cru.org styling from cru.scss.
...
+- `code` must be changed to the designation number you want users to give to.
+- `tsys-env` must be changed to match your environment. It is provided by DPS when adding your domain to the TSYS whitelist.
 ```html
 <link rel="stylesheet" href="https://give-static.cru.org/branded-checkout.min.css">
 <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.6.3/css/font-awesome.min.css">
 
 <branded-checkout
     ng-app="brandedCheckout"
-    code="0763355">
+    code="0763355"
```
mattdrees: out of curiosity, why'd you name this 'code' instead of 'designationNumber'? Maybe we can change this?

README.md
```diff
@@ -30,14 +30,17 @@ This currently includes all cru.org styling from cru.scss.
 
 #### Branded checkout basic example
 
-Add the following code to your page where appropriate:
+Add the following code to your page where appropriate. You must change the value of the `code` and `tsys-env` attributes:
+- `code` must be changed to the designation number you want users to give to.
+- `tsys-env` must be changed to match your environment. It is provided by DPS when adding your domain to the TSYS whitelist.
```
mattdrees: I was actually wrong about this value being specific to the environment.
The current values are 'cru' and 'familyLife'.

Perhaps call this 'tsys-device'? Or 'tsys-device-id'?

README.md
OzzieOrca: Probably because other frontend code or EP called it that. Now's the time to change it before it starts getting implemented.

mattdrees: EP might probably does call it that. But EP's terminology isn't as important as Cru's terminology, which is going to be more stable long-term. (There's no guarantee we'll stick with EP for more than a few years.) Think of EP as an implementation detail.

So yeah, let's change this now. Thanks!

OzzieOrca:
I made those changes. I got Josh's email saying TSYS now supports multiple whitelist urls per merchant id so I marked `tsys-device` as optional in the readme and put it at the bottom with this comment https://github.com/CruGlobal/give-web/pull/701/files#diff-04c6e90faac2675aa89e2176d2eec7d8R123. We could remove it if it will never be used.

PR #703, opened 2017-11-07
Add language picker to nav, remove global modal

adammeyer:


src/app/main/main.tpl.html
```diff
@@ -1,4 +1,4 @@
-<cru-nav></cru-nav>
+<cru-nav language="United States / English"></cru-nav>
```
Omicron7: Not sure where the list comes from, but it would seem better to use a code here instead of the translated (or maybe always english) text. Ex: en-US

adammeyer: Then we would need a database of lang codes on the JS side.  I'll search around, maybe there is a library.

adammeyer: Maybe something like this code work.  https://github.com/bdswiss/country-language

@OzzieOrca What are you thoughts?

adammeyer:
Renamed to "language-picker-label"

src/app/main/main.tpl.html
OzzieOrca: Something like that could work. I know for the EP stuff we're doing API requests to get countries but that probably doesn't help for languages. It kinda looks like you are getting languages from AEM https://github.com/CruGlobal/give-web/pull/703/files#diff-6d445de8f81bf08615bab285832fae9aR20. I'm not sure I fully understand what you are trying to do.

adammeyer:
Good call.  css removed.

src/app/main/main.tpl.html
adammeyer: screen shot 2017-11-07 at 2 35 54 pm

The language-picker-lable is used for the link to expand the picker above.  The list in the picker does come from AEM, just like the global overlay did.

OzzieOrca: Ok. What would the language library be for then? Just prefilling the US/English text? Or for translating some of this text?

adammeyer: That attribute will be set by AEM, so I think the text label is fine.

OzzieOrca: k

PR #713, opened 2018-01-10
Scrub rollbar telemetry

OzzieOrca:
I messed up. I should have know rollbar would start logging input fields, including credit card and bank account fields. I sort of knew it had that capability but wasn't thinking about it when upgrading. I guess I've mostly looked at telemetry for network requests. This scrubs the values from those fields.

https://github.com/CruGlobal/give-web/pull/712 added telemetry.

### DIFF
--- a/src/common/rollbar.config.js
+++ b/src/common/rollbar.config.js
@@ -4,6 +4,8 @@ import stacktrace from 'stacktrace-js';
 import map from 'lodash/map';
 import defaults from 'lodash/defaults';
 import get from 'lodash/get';
+import find from 'lodash/find';
+import includes from 'lodash/includes';
 
 let Rollbar;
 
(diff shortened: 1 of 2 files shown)


OzzieOrca:
Merging to stop this logging. I'll try to figure out the best way to purge stuff from rollbar and talk to Lee about it. Idk if this has PCI implications.

src/common/rollbar.config.js
```diff
@@ -129,11 +132,19 @@ function updateRollbarPerson(session){
   });
 }
 
+function scrubDomNodes(scrubNames){
+  return node => {
+    const nameAttr = find(node.attributes, attr => attr.key === 'name');
+    return !!nameAttr && includes(scrubNames, nameAttr.value);
```
wrandall22: What does `!!` do?

OzzieOrca: Same as `Boolean(nameAttr)`. A cast essentially. Idk what is most readable. If nameAttr is falsy it will return false instead of undefined or whatever. Negating it turns it into a Boolean and doing it again unnegates it.

OzzieOrca: Same as `Boolean(nameAttr)`. A cast essentially. Idk what is most readable. If nameAttr is falsy it will return false instead of undefined or whatever. Negating it turns it into a Boolean and doing it again unnegates it.

wrandall22: Ah, interesting.

PR #720, opened 2018-02-16
Allow multiple formats for suggested amounts

wrandall22:
The out of the box Touch UI composite multifield editor creates nodes in the form of `item0`, `item1`, etc. We want to switch to using that instead of the broken Classic UI editor that was created for this purpose. In order to support the transition, this PR is here.


wrandall22:
Do you know why Line 179 uses `apiUrl` instead of `publicGive`? The cortex API doesn't expose the infinity json, its AEM that does that.

adammeyer:
Because it all goes through give.cru.org, the domain is the same.

PR #723, opened 2018-02-21
remove icon-search, conflicts with new nav search

adammeyer:


src/assets/scss/styles.scss
```diff
@@ -26,7 +26,3 @@
 @import 'homepage';
 @import 'search';
 @import 'mobile';
-
-// ********** This fonts scss file is only needed for gh-pages to correct relative links to be able to use icons both remote and local. ********** //
```
adammeyer: I'm assuming this is okay to remove because we no longer use gh-pages.  Will test on stage.

PR #727, opened 2018-03-02
sign in error msg improvement

adammeyer:
Requested here: https://jira.cru.org/browse/EP-2104


src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -1,4 +1,11 @@
-<div class="alert alert-danger" role="alert" ng-if="$ctrl.errorMessage">{{$ctrl.errorMessage | translate}}</div>
+<div class="alert alert-danger" role="alert" ng-if="$ctrl.errorMessage" ng-switch="$ctrl.errorMessage">
+  <span ng-switch-when="generic" translate>
+    An error has occurred. Please try again or contact <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.
+  </span>
+  <span ng-switch-default>
+    {{$ctrl.errorMessage | translate}}
```
Omicron7: You removed 'gettext', is that needed for translate filter?

adammeyer: O, okay.  I thought that was only for the controller.  I will add it back in.

PR #728, opened 2018-03-02
add ngAnimate

adammeyer:
Didn't mean to remove this.

What is ngAnimate used for?


adammeyer:
Looks like the login modals don't animate down when that's removed.

OzzieOrca:
I guess that's a good reason to keep it. Didn't know ng bootstrap used it... but makes sense

PR #729, opened 2018-03-07
add campaign page to branded checkout (for suggest amounts)

adammeyer:


README.md
```diff
@@ -76,6 +76,7 @@ Add the following code to your page where appropriate. See the [Branded checkout
 The `<branded-checkout>` element is where the branded checkout Angular app will be loaded. It is configured by providing HTML attributes that will be loaded by Angular. Attributes with values containing angle brackets (such as `<designation number>`) are placeholders and should be replaced with real values or, if not needed, the whole attribute should be omitted. The `<branded-checkout>` element accepts the following attributes:
 - `ng-app="brandedCheckout"` - tells Angular which module to load - **Required** - you could bootstrap Angular manually or include this `brandedCheckout` module in your own custom Angular module instead if desired
 - `designation-number` - the designation number you would like donors to give to - **Required**
+- `campaign-page` - the campaign page you would like to use - *Optional*
```
OzzieOrca: Could you maybe elaborate on what the difference between these 2 is? I don't know.  Idk if that will be common knowledge for people setting up branded checkout but they may be more familiar with what a designation looks like...

adammeyer: I added that it's used for suggested amounts.  Campaign code is something for email marketing and such.  I don't think it would be used for a branded checkout.

wrandall22: The legacy branded checkout uses campaign codes. Not sure if that will be true moving forward.

PR #734, opened 2018-03-12
logout before branded checkout

adammeyer:
This performs a logout before beginning a branded checkout session to ensure clean gift.

@OzzieOrca I added the "downgrade" call after the logout to ensure that new cookies are set.  I was noticing that if that's not done, asynchronous calls where setting different session cookies on step one load.  That may have actually been the root cause of the original issue.

### DIFF
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -30,13 +30,11 @@ class BrandedCheckoutController {
     this.code = this.designationNumber;
     this.tsysService.setDevice(this.tsysDevice);
     this.analyticsFactory.pageLoaded(true);
+    this.formatDonorDetails();
 
     this.sessionService.signOut().subscribe(() => {
-      this.sessionService.downgradeToGuest(true, false).subscribe(() => {
-        this.checkoutStep = 'giftContactPayment';
-        this.formatDonorDetails();
-        this.fireAnalyticsEvents('contact', 'payment');
-      }, angular.noop);
+      this.checkoutStep = 'giftContactPayment';
+      this.fireAnalyticsEvents('contact', 'payment');
     }, angular.noop);
   }
 
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -8,6 +8,8 @@ import step1 from './step-1/branded-checkout-step-1.component';
 import step2 from './step-2/branded-checkout-step-2.component';
 import thankYouSummary from 'app/thankYou/summary/thankYouSummary.component';
 
+import sessionService from 'common/services/session/session.service';
+
 import 'common/lib/fakeLocalStorage';
 
 import template from './branded-checkout.tpl.html';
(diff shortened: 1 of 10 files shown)


src/app/branded/branded-checkout.component.js
```diff
@@ -17,19 +19,25 @@ let componentName = 'brandedCheckout';
...
     this.tsysService.setDevice(this.tsysDevice);
-    this.checkoutStep = 'giftContactPayment';
-    this.formatDonorDetails();
     this.analyticsFactory.pageLoaded(true);
-    this.fireAnalyticsEvents('contact', 'payment');
+
+    this.sessionService.signOut().subscribe(() => {
+      this.sessionService.downgradeToGuest(true, false).subscribe(() => {
+        this.checkoutStep = 'giftContactPayment';
+        this.formatDonorDetails();
```
OzzieOrca: Did this need to be moved to after the signout? It probably doesn't matter when it happens since it is pulling from bindings unless I'm missing something or the session change overwrites them.

src/app/branded/branded-checkout.component.js
```diff
@@ -17,19 +19,25 @@ let componentName = 'brandedCheckout';
...
   $onInit() {
     this.code = this.designationNumber;
     this.tsysService.setDevice(this.tsysDevice);
-    this.checkoutStep = 'giftContactPayment';
-    this.formatDonorDetails();
     this.analyticsFactory.pageLoaded(true);
-    this.fireAnalyticsEvents('contact', 'payment');
+
+    this.sessionService.signOut().subscribe(() => {
+      this.sessionService.downgradeToGuest(true, false).subscribe(() => {
```
OzzieOrca: I think this just needs to be any, single, request. I talked with Brian about it a while ago and I think that's what we figured out. We need to request a token once and store it in cookies before sending more requests that load the token from cookies otherwise we will keep getting new tokens and storing them and requests with old ones won't work.

You may not need the `verifyRole` flag in the session service if you don't need to call `this.sessionService.downgradeToGuest`.

Is any request sent in parallel with `this.cartService.get()` in step 1? Doesn't look like it. But the sub components may start sending requests. That may have been why I had the big wrapper `ng-if` in `branded-checkout-step-1.tpl.html` to keep sub components from rendering until `this.cartService.get()` had completed.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -1,15 +1,13 @@
 import angular from 'angular';
 import find from 'lodash/find';
 import every from 'lodash/every';
-import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/operator/mergeMap';
```
OzzieOrca: This still needed?

adammeyer: Intelij said it wasn't.  What do you think uses it?

OzzieOrca: I think it was used by a line you deleted. Idk if imports that aren't assigned to a variable get marked correctly in an IDE since it can't check if the variable gets used

adammeyer: That's correct.  It was used, but I removed that code.

PR #739, opened 2018-03-22
Cart dropdown tweaks

j2trumpet:


adammeyer:
How is this being referenced?  Doesn't it need to be added to styles.scss?

PR #740, opened 2018-03-27
branded checkout: pass in default donor details after uri is loaded f…

adammeyer:
…rom server


adammeyer:
Thanks @OzzieOrca 

We still need to merge in the defaults, because they won't be saved on EP yet.

The only place where the donor-details attribute is used is on branded-checkout-step1.  So I think we're good there.

OzzieOrca:
K cool. That's sorta what I thought but wasn't sure.

PR #742, opened 2018-03-27
only update mailing address if set

adammeyer:


adammeyer:
found another bug.  This prevents mailing address to be set to 'undefined'.

PR #744, opened 2018-03-28
remove cru-nav and dependencies

adammeyer:


adammeyer:
Yes, the nav is all in AEM now.  The `navCartIcon` component is still around, putting that element directly in the AEM nav.

OzzieOrca:
Ok cool. Does that mean sign in is handled by AEM?

adammeyer:
Yes, I'm calling the services directly.

https://github.com/CruGlobal/giveaem/blob/451cb28f81ad9d5bf536fece5238ecc8b3d3f7fc/ui.apps/src/main/content/jcr_root/etc/designs/give/clientlibs/give-global/js/give-nav.js#L39

README.md
```diff
@@ -12,7 +12,7 @@ To use the `searchResults` component, include this code on your page:
 ```html
 <search-results ng-app="searchResults"></search-results>
 
-<script src="https://cru-givestage.s3.amazonaws.com/app.js"></script>
+<script src="https://give-static.cru.org/app.js"></script>
```
adammeyer: @OzzieOrca I updated the readme to use the production domain instead of staging.  Do you know why the staging domain was used originally?

OzzieOrca: Idk, maybe we weren't running production at the time. Or I figured people should test on staging first... Idk. This is fine. Maybe sticking the staging urls somewhere for reference would be good.

adammeyer: okay, added a note on the staging env to the readme

PR #745, opened 2018-03-29
only override donor details on first load of form

adammeyer:
This fixes an issue where a user would go back to edit their details from the review screen and they would be overridden by the defaults from the window object.

Let me know if you have a better way to detect this, I just noticed that that link doesn't exist on the first load, but does on the second.

### DIFF
--- a/src/common/components/contactInfo/contactInfo.component.js
+++ b/src/common/components/contactInfo/contactInfo.component.js
@@ -1,6 +1,7 @@
 import angular from 'angular';
 import 'angular-messages';
 import assign from 'lodash/assign';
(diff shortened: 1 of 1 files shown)


OzzieOrca:
I still think you should layer these on top of each other. `Object.assign({}, defaultDonorDetails, windowDonorDetails, apiDonorDetails)`. That may not handle the deeply nested address so it might need a lodash deep merge function or something. Maybe the api version has bad defaults or would overwrite something with an empty string. If doing this doesn't work, the code you have here is probably fine. I have no idea how consistently that field appears there.

PR #746, opened 2018-03-29
turn off autocomplete for bank and credit card fields

adammeyer:
Although this is helpful when testing, probably not great to have that remembered in the user's browser.


OzzieOrca:
Does this kill the browser's auto form filling? Seems like a good browser should detect that these are credit card fields and treat them differently. Chrome seems to.
https://support.google.com/chrome/answer/7665863
https://www.google.com/chrome/browser/privacy/whitepaper.html#autofill

I'm fine with doing this for banks. I think we should test credit cards a little more.

adammeyer:
I'll update it according to this guide. https://developers.google.com/web/fundamentals/design-and-ux/input/forms/#recommended_input_name_and_autocomplete_attribute_values

Sound good?

PR #747, opened 2018-03-29
update itemConfig if item is in cart

adammeyer:
ignore 'code' check since it changes if the gift frequency is changed


adammeyer:
Editing a gift from the review page was resetting all the options to the initial state and adding a second gift if the frequency was changed.

PR #750, opened 2018-04-02
remove homeSignIn component (no longer used)

adammeyer:


adammeyer:
Fund dev didn't like it

PR #753, opened 2018-04-06
redirect to cart after adding gift on mobile

adammeyer:


src/app/productConfig/productConfigModal/productConfig.modal.component.js
```diff
@@ -21,6 +21,8 @@ class ProductConfigModalController {
   $onInit(){
     this.initModalData();
     this.initializeParams();
+
+    this.isMobile = this.$window.innerWidth <= 575;
```
OzzieOrca: Do we do this anywhere else? Seems like it might be good to store as a constant somewhere.

PR #754, opened 2018-04-09
include 'name' as valid donor details field

adammeyer:


adammeyer:
Yes, this will take the whole 'name' object if it exists.

OzzieOrca:
Right. That's an object. Cool.

PR #765, opened 2018-04-26
Cart count cookie

adammeyer:
screen shot 2018-04-26 at 2 05 26 pm

Give and Cru.org will have a new design that shows the item count in the header.  Setting a cookie will avoid doing a cart get request on every page load.


adammeyer:
Need to be able to read on give.cru.org and cru.org, so localStorage won't work for that.

Yes, Safari should be good since both sites are cru.org.

Any idea on the $digest error on the test?  I'm stuck there.

PR #772, opened 2018-05-01
continue browsing button mobile cart

adammeyer:
I guess $document is an array


adammeyer:
that it's an array?  Or how I did it?

OzzieOrca:
That it's an array... The script.js on https://docs.angularjs.org/api/ng/service/$document has it too. It must be some jQuery/jQlite thing where it has an array of elements that matched the selectors. But idk how you'd have more than 1 document.

adammeyer:
ya, I agree, that is weird.

PR #775, opened 2018-05-04
[#EP-2111] Add MFA support to sign-in.

Omicron7:
This adds Multi-factor support to Sign-In.
It should continue to work without the cortex-gateway changes, it just won't show the MFA token field.

### DIFF
--- a/src/common/components/signInForm/signInForm.tpl.html
+++ b/src/common/components/signInForm/signInForm.tpl.html
@@ -2,34 +2,56 @@
   <span ng-switch-when="generic" translate>
     An error has occurred. Please try again or contact <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.
   </span>
+  <span ng-switch-when="mfa" translate>
+    Invalid Verification Code. Please try again.
+  </span>
   <span ng-switch-default>
     {{$ctrl.errorMessage | translate}}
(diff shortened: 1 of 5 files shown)


Omicron7:
Cortex Gateway change: https://github.com/CruGlobal/cortex_gateway/pull/52

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -2,34 +2,56 @@
...
-           placeholder="{{'Password'|translate}}"
-           autocomplete="off"
-           required
-           ng-disabled="$ctrl.isSigningIn"
-           ng-model="$ctrl.password">
+  <div ng-switch-when="mfa">
+    <p translate>Enter a verification code from your Authenticator app to continue.</p>
+    <div class="form-group">
+      <label for="mfa_token" translate>Verification Code</label>
+      <input id="mfa_token"
```
OzzieOrca: @Omicron7 Can we add autofocus here? I just used it for the first time and thought that would be nice.

adammeyer: 👍

Omicron7: @OzzieOrca You can indeed add autofocus there.

PR #776, opened 2018-05-10
Stop logging lastPurchaseLink error to cleanup rollbar

OzzieOrca:


adammeyer:
Why does this one happen so often?

OzzieOrca:
People must leave that tab open and it refreshes later... Idk. Or is reopened when they restart their browser... Or something else but those ideas sorta make sense... Idk if that accounts for that many reports though...

PR #777, opened 2018-05-14
clean and limit campaign code

adammeyer:


wrandall22:
So, in the use case we saw the other day, everything from the `?` forward should have been removed, but this seems to just remove the `?` and go down to 30 characters. I would think we would want, for `?` at least, to strip everything afterward (since it is likely a different query parameter that got tacked on).

adammeyer:
@mattdrees Liked that idea.  Updated.

PR #785, opened 2018-10-31
Add branded checkout attribute default-payment-type

OzzieOrca:
Request came from https://secure.helpscout.net/conversation/698088015/266883/


src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -89,6 +89,7 @@ class ExistingPaymentMethodsController {
         paymentMethod: existingPaymentMethod,
         disableCardNumber: !!existingPaymentMethod && !existingPaymentMethod['from-current-order'],
         mailingAddress: this.mailingAddress,
+        defaultPaymentType: () => this.defaultPaymentType,
```
OzzieOrca: Resolve is always weird... Idk why the others work. If I don't make it a function, it tries to use it as a component or something and I get `Unknown provider: creditCardProvider <- creditCard`.

Omicron7:
@j2trumpet New option added to branded checkout, we might want to add it to the WordPress plugin as well.

OzzieOrca:
Thanks guys!

PR #787, opened 2019-03-07
Allow Different Family Names

wrandall22:
[Helpscout ticket](https://secure.helpscout.net/conversation/795335774/302392?folderId=381778)

Sometimes the spouse family names are not the same. In these cases, show both spouse family names.

Old:
screen shot 2019-03-07 at 1 58 58 pm

New:
screen shot 2019-03-07 at 1 59 50 pm


wrandall22:
I think maybe this was never confirmed to be wanted for sure. I'll close it for now, and re-open if I ever hear back from DSG.

PR #790, opened 2019-06-10
EP-2128 - replace DTM with Launch

wrandall22:
This replaces the hard-coded script inside of `index.html`. I'm not actually sure if this is used anywhere besides a local machine test, or some other test that bypasses AEM.

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,6 +14,45 @@
       <script>
         window.digitalData = {};
       </script>
+      <script type="text/javascript">
+        ;(function(win, doc, style, timeout) {
+          var STYLE_ID = 'at-body-style';
+
+          function getParent() {
+            return doc.getElementsByTagName('head')[0];
+          }
+
+          function addStyle(parent, id, def) {
+            if (!parent) {
+              return;
+            }
+
+            var style = doc.createElement('style');
+            style.id = id;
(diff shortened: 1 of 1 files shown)


wrandall22:
It's a workaround that Adobe (I think) gave us to prevent flickering when Adobe Target is added to the page.

OzzieOrca:
Interesting...

wrandall22:
https://docs.adobe.com/help/en/target/using/implement-target/client-side/at-js/manage-flicker-with-atjs.html

OzzieOrca:
Those docs say
> We recommend using a tag manager such as Adobe DTM or the new Adobe Launch to add the pre-hiding snippet.

wrandall22:
Indeed they do. I'll ask Clark why he specifically required us to put the script before the Launch script instead of inside of it.

wrandall22:
From Clark:

> the script only works if it fires before and outside launch
> it's a bummer
> we've tried every combination with the firing order, and then got confirmation from Adobe that the script has to run before Launch and can't be triggered from inside

Sounds like they've already mentioned the issue in the documentation to Adobe.

OzzieOrca:
K that's fine. It's weird launch can't run stuff before target.

PR #791, opened 2019-06-20
Fix build and modals

Omicron7:
* This fixes the node-sass compile errors under Node 8.x.
* Fixes issue where modals would not show on local development.


Omicron7:
I used yarn to do the upgrades, it must lock it to the value given.
`yarn upgrade node-sass@4.5.3`

OzzieOrca:
Oh ya that makes sense. I've been using `yarn upgrade-interactive --latest` lately.

PR #792, opened 2019-06-21
Add self signed cert and use it with webpack-dev-server

Omicron7:
This adds a self signed cert for localhost.cru.org.
HTTPS on localhost is now required to do local development because of https://github.com/CruGlobal/cortex_gateway/pull/58


mattdrees:
alternatively, we could probably disable force_ssl in development mode. But this seems reasonable.

Omicron7:
Our development usually hits stage cortex, so this is probably better.

Omicron7:
This project is using Node 8. I'd suggest asdf or nvm to downgrade for this one.

PR #794, opened 2019-07-03
Add hideBankAccountLegalStatement flag to branded checkout

OzzieOrca:
https://secure.helpscout.net/conversation/784005984/298110?folderId=610613

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Add the following code to your page where appropriate. See the [Branded checkout
     day="<day>"
     default-payment-type="creditCard"
     donor-details="<window variable containing default values for donor's name and contact info>"
-    hide-bank-account-legal-statement="true"
+    hide-payment-type-options="true"
     on-order-completed="$event.$window.onOrderCompleted($event.purchase)"
     on-order-failed="$event.$window.onOrderFailed($event.donorDetails)">
 </branded-checkout>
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -63,6 +63,7 @@ Add the following code to your page where appropriate. See the [Branded checkout
     day="<day>"
     default-payment-type="creditCard"
     donor-details="<window variable containing default values for donor's name and contact info>"
+    hide-bank-account-legal-statement="true"
```
Omicron7: Let's not add this option to the README for now. FDS wanted it to be obscure so only very specific branded sites they tell us about could use it.

OzzieOrca: Guess I read the helpscout conversation wrong...

> Can you please confirm when you update the branded checkout so it doesn't include EFT disclosure text with references to Cru?

But the rest of it wants the whole thing removed. I'll rework this.

src/common/components/paymentMethods/paymentMethodForm/paymentMethodForm.tpl.html
```diff
@@ -45,7 +45,7 @@
 
 <div class="tab-indiv cart-tab show" ng-if="$ctrl.paymentType === 'bankAccount'">
 
-  <bank-account-form payment-form-state="$ctrl.paymentFormState" on-payment-form-state-change="$ctrl.onPaymentFormStateChange({ $event: $event })" payment-method="$ctrl.paymentMethod" hide-bank-account-legal-statement="$ctrl.hideBankAccountLegalStatement"></bank-account-form>
+  <bank-account-form payment-form-state="$ctrl.paymentFormState" on-payment-form-state-change="$ctrl.onPaymentFormStateChange({ $event: $event })" payment-method="$ctrl.paymentMethod" hide-payment-type-options="$ctrl.hidePaymentTypeOptions"></bank-account-form>
```
wrandall22: Seems like we might need to add the new property to the credit card form as well, since the default might be bank account form.

OzzieOrca: I figured since `default-payment-type` already exists, this would be enough. It's more flexible and the code is simpler.

OzzieOrca: Looks like I forgot to remove this line. It was left over from hiding the legal message. Thanks :)

PR #795, opened 2019-07-03
Upgrade all packages

OzzieOrca:
- Upgrade to webpack 4
- Upgrade to babel 7
- Skip rxjs major upgrade for now

I plan to run Prettier on the codebase after these are merged. I was just trying to keep formatting changes to a minimum and had to turn it off locally.

I clicked around most screens. We should do more testing on staging.


OzzieOrca:
Thanks :) You fine with me merging this? I don't want to block anything else you were trying to deploy if we want to test this for a bit.

`isBrandedCheckout` already existed but it started complaining about not having a default value with the new version. https://github.com/CruGlobal/give-web/search?q=isBrandedCheckout&unscoped_q=isBrandedCheckout

Omicron7:
Feel free to merge.

PR #797, opened 2019-07-11
HS-338328 - Change Ministry Landing Pages

wrandall22:
As per request from FDS, this changes the landing pages when a user clicks on `Details` for these ministries. There will likely be more to come, but this is the first batch.

[HelpScout ticket](https://secure.helpscout.net/conversation/892961764/338328/?folderId=381778)

PR #798, opened 2019-07-11
Revert "Upgrade all packages"

wrandall22:
Reverts CruGlobal/give-web#795

This is causing tests to fail intermittently and any further work is stopped.


OzzieOrca:
Did this fix things for now? Sorry about that. Sounded like tests were failing anyways. Guess I'll get back to this eventually...

wrandall22:
Yes, reverting this allowed me to move forward with #797

PR #806, opened 2019-08-12
Update packages

Omicron7:
This builds on top of @OzzieOrca previous PR to update all packages (except rxjs).

- Upgrade packages
- Switched testing framework to [Jest](https://jestjs.io/) and fixed all specs
- Added [JavaScript Standard Style](https://standardjs.com/) and fixed anything `--fix` didn't get.
- Removed eslint and other unnecessary packages
- Fixed travis build

Since I'm working on all current PR's, I'll update and rebase them off this.


package.json
```diff
@@ -75,7 +69,25 @@
...
+      "<rootDir>/jest/setup.js"
+    ],
+    "modulePaths": [
+      "<rootDir>/src"
+    ],
+    "transform": {
+      "^.+\\.js?$": "babel-jest",
+      "^.+\\.html$": "<rootDir>/jest/htmlTransform.js"
+    },
+    "verbose": true
```
OzzieOrca: Were you intending to keep this? Seems like it logs out successful tests also...

package.json
```diff
@@ -75,7 +69,25 @@
     "start": "webpack-dev-server",
     "build": "webpack -p",
     "build:analyze": "webpack -p --env.analyze",
-    "test": "karma start",
+    "test": "jest",
     "lint": "eslint 'src/**/*.js'"
+  },
+  "jest": {
```
OzzieOrca: I sorta like moving this to `jest.config.js` just to keep this file cleaner. But either is fine.

package.json
```diff
@@ -93,5 +94,14 @@
       "^.+\\.html$": "<rootDir>/jest/htmlTransform.js"
     },
     "verbose": true
+  },
+  "standard": {
+    "env": {
+      "jest": true
+    },
+    "ignore": [
+      "**/*.spec.js",
```
OzzieOrca: So you're not running standard on the tests?

src/app/analytics/analytics.factory.js
```diff
@@ -1,40 +1,39 @@
...
-import merge from 'lodash/merge';
-import isEmpty from 'lodash/isEmpty';
+import './analytics.module'
+import angular from 'angular'
+import get from 'lodash/get'
+import find from 'lodash/find'
+import sha3 from 'crypto-js/sha3'
+import merge from 'lodash/merge'
+import isEmpty from 'lodash/isEmpty'
+/* global localStorage */
```
OzzieOrca: Shouldn't this be part of the standard config? Or you could do `$window.localStorage`...

package.json
```diff
@@ -2,78 +2,100 @@
...
+    ],
+    "transform": {
+      "^.+\\.js?$": "babel-jest",
+      "^.+\\.html$": "<rootDir>/jest/htmlTransform.js"
+    },
+    "verbose": true
+  },
+  "standard": {
+    "env": {
+      "jest": true
```
OzzieOrca: Should something for the browser be included here? ESLint lets you specify different envs for different file globs. https://github.com/CruGlobal/missionhub-web/blob/master/.eslintrc.json#L10-L15

Omicron7:
@OzzieOrca I've tried prettier in a project after your recommendation. I found I disliked the customizations it allowed, and the fact that I still needed a linter. StandardJS solves most of this by having a very strict set of rules, which you can't modify. I agree, it leaves leeway for a programmer to format code slightly differently, but I'm ok with that.

Out of the box, StandardJS is not customizable. It's the ultimate preventer of bike shedding as no one developer can change any rules. That said, you can switch the parser to eslint and add eslint plugins if you want. Like Prettier, standardjs has plugins for all major editors and git hooks.

In my testing, prettier and standardjs are mutually exclusive. You can't customize prettier such that it produces standardjs style, so they are always auto correcting each other.

package.json
Omicron7: I liked the output it produced. `verbose: false` was harder in travis to see exactly what test in a file was failing.

package.json
Omicron7: I can enable it for them, but initial run through them had way more errors than I was willing to fix. Most of it was things like not handling promise rejections, which we rarely ever do in a test.

src/app/analytics/analytics.factory.js
Omicron7: I added it here instead of global config so that we'd get a warning if we used it in other places.

package.json
Omicron7: Only browser global it complained of was `localStorage`, so I didn't think it was necessary.

PR #810, opened 2019-09-18
Temporarily hardcode the help content

mattdrees:
For reasons we don't understand yet,
https://give.cru.org/designations/jcr:content/need-help-ipar/contentfragment.html
is failing with a 500 internal error.

For now, we'll hardcode the message.

### DIFF
--- a/src/app/checkout/help/help.component.js
+++ b/src/app/checkout/help/help.component.js
@@ -13,11 +13,16 @@ class CheckoutHelpController {
   }
 
   $onInit () {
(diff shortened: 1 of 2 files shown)


mattdrees:
Screen Shot 2019-09-17 at 6 50 52 PM

from staging

mattdrees:
thanks scotty!

mattdrees:
helpscout: https://secure.helpscout.net/conversation/958170004/365918/

PR #812, opened 2019-09-19
Revert "Temporarily hardcode the help content"

wrandall22:
Reverts CruGlobal/give-web#810

The fix was deployed this morning to no longer need this workaround. See https://github.com/CruGlobal/cru-aem-commons/pull/9

### DIFF
--- a/src/app/checkout/help/help.component.js
+++ b/src/app/checkout/help/help.component.js
@@ -13,16 +13,11 @@ class CheckoutHelpController {
   }
 
   $onInit () {
-    // this.$http.get('/designations/jcr:content/need-help-ipar/contentfragment.html').then((response) => {
-    //   this.helpHTML = response.data
-    // }, error => {
-    //   this.$log.error('Error loading give-need-help', error)
-    // })
-
-    this.helpHTML = '<div><h3 class="panel-name">Need Help?</h3>\n' +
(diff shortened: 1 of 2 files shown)


src/app/checkout/help/help.component.js
```diff
@@ -13,16 +13,11 @@ class CheckoutHelpController {
...
-    //   this.helpHTML = response.data
-    // }, error => {
-    //   this.$log.error('Error loading give-need-help', error)
-    // })
-
-    this.helpHTML = '<div><h3 class="panel-name">Need Help?</h3>\n' +
-        '<p>Call us at <b>(888)278-7233</b> from <b>9:00 a.m. to 5:00 p.m. ET,</b> Monday-Friday, or email us at <a href="mailto:eGift@cru.org" target="_top">eGift@cru.org</a>.</p>\n' +
-        '<p>We have also compiled a list of answers to <a>Frequently Asked Questions.</a></p>\n' +
-        '</div>'
+    this.$http.get('/designations/jcr:content/need-help-ipar/contentfragment.html').then((response) => {
```
mattdrees: do you think this url is still the best way to get the help content?

wrandall22: It's certainly an acceptable way to get it. There isn't a better place to get it from right now.

mattdrees: K. I wonder if a vanity url would be better. But not needed right now.

wrandall22: I'd be surprised if a vanity url would work for this (from AEM's vanity perspective, maybe Apache could do it fine) since it's not a page, it's just a JCR node.

PR #819, opened 2019-10-29
EP-2138 - Manifest and loader

wrandall22:
This PR includes a JS loader to asynchronously load the libraries we care about in order to (hopefully) make the branded checkout faster at loading. This is versioned, so will not affect ministries right away; only when they update the file name to version 2.


wrandall22:
cc @katoblakmask

wrandall22:
@Omicron7 would you be willing to resolve the conflicts such that the 2 files come out correctly?

wrandall22:
Thanks!

PR #820, opened 2019-11-15
Add Unto to ministry search results

wrandall22:
With GAiN renaming to Unto, they want to add Unto (alongside GAiN for now) onto the ministry search results page. At some point I would like this to become doable by AEM give site authors (see [Jira](https://jira.cru.org/browse/EP-2157) and [HelpScout](https://secure.helpscout.net/conversation/1006709892/385635/)).


wrandall22:
Thanks!

PR #824, opened 2019-11-26
Multilingual Functionality for Branded Checkout

moeloubani:
This adds translation functionality and some translations in Spanish, the translations are stored in the app.config.js file and the translation is done using angular-translate https://github.com/angular-translate/angular-translate. A 'language' parameter is used to control what language is shown with the current options being 'en' (the default) for English and 'es' for Spanish.


src/common/app.config.js
```diff
@@ -78,8 +78,326 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
...
+    CARD_EXP_MONTH_ERROR: 'Debes elegir un mes de vencimiento',
+    CARD_EXPIRED_ERROR: 'Su tarjeta de crédito ha caducado o seleccionó el mes o año incorrecto',
+    CARD_EXP_YEAR_ERROR: 'Debes elegir un año de vencimiento',
+    CARD_SEC_CODE_ERROR: 'Debes ingresar el código de seguridad',
+    MIN_LENGTH_CARD_SEC_CODE: 'El código de seguridad debe tener al menos 3 dígitos.',
+    MAX_LENGTH_CARD_SEC_CODE: 'El código de seguridad no puede tener más de 4 dígitos.',
+    LOCATION_OF_CODE_AMEX: 'Código de 4 dígitos en el frente de la tarjeta',
+    LOCATION_OF_CODE_OTHER: 'Código de 3 dígitos en el reverso de la tarjeta',
+    LOADING_ADDRESS: 'Cargando dirección postal...',
+    COUNTRY_LIST_ERROR: 'There was an error loading the list of countries. If you continue to see this message, contact <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.',
```
mattdrees: @moeloubani this one isn't Spanish. Did this and the ones below get missed? Or intentionally deferred?

moeloubani: Some of the translations have come yet from the global team working on them but we wanted to get the form out with the incomplete translations for now and add those in once we get them

mattdrees: great

mattdrees:
I want to touch base with Josh and/or Bill first, but pretty sure we will deploy this relatively soon.

src/common/app.config.js
```diff
@@ -76,10 +77,426 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
   }
 
   $qProvider.errorOnUnhandledRejections(false)
+
+  $translateProvider.translations('en', {
```
wrandall22: I think it's a little unfortunate to have these translations in this file. I would wonder if we could have separate language files that we pulled in instead (e.g. en.json, es.json) or if that would be too non-performant.

mattdrees: yeah, not ideal. Also, Brian said he'd like to replace this translation mechanism (angular-translate) with angular-gettext, which Scotty had actually already set up in this project. Brian just didn't realize it until recently.

So we'll do some cleanup. Not sure if it'll happen after peak giving or before.

moeloubani: I have some time over the next couple of days with Thanksgiving being a while ago in Canada, I can put the translations into a new branch with the commits I did to translate the way I did reverted and do it again using angular-gettext if that is ok

mattdrees: Yeah, that sounds good to me. @Omicron7 do you want to give Moe some direction on this?

PR #825, opened 2019-12-02
Remove translate directive from elements that contain AngularJS bindings

OzzieOrca:
https://secure.helpscout.net/conversation/1017916513/389669/

### DIFF
--- a/src/app/profile/profile.tpl.html
+++ b/src/app/profile/profile.tpl.html
@@ -136,7 +136,7 @@
                       <loading type="overlay" ng-if="$ctrl.emailLoading"></loading>
                       <form name="$ctrl.donorEmailForm" novalidate>
                         <div class="form-group is-required" ng-class="{'has-error': ($ctrl.donorEmailForm.donorsEmail | showErrors)}">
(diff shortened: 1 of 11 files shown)


OzzieOrca:
Ya I think most of these will need to be redone with however the library handles interpolation to combine strings with variables.

I've been using https://www.i18next.com/ on everything new (meaning missionhub-web and missionhub-react-native and any new React stuff I do). It'd be easy to convert to away from Angular or use in a React/Angular hybrid app since the filter is about the only framework specific thing.

Filter - https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/filters/translation.filter.js
Config - https://github.com/CruGlobal/missionhub-web/blob/master/src/i18n/i18next.config.ts
Upload/Download scripts - https://github.com/CruGlobal/missionhub-web/tree/master/bin
English translations - https://github.com/CruGlobal/missionhub-web/blob/master/src/i18n/locales/en-US.js

Here's how i18next handles interpolation: https://www.i18next.com/translation-function/interpolation

Looks like [MPDx](https://github.com/CruGlobal/mpdx_web) is using angular-gettext if that's a helpful resource.

PR #826, opened 2019-12-18
Add babel polyfill for IE 11 Promise support

Omicron7:
This adds a polyfill for IE11 by following this: https://stackoverflow.com/questions/52470124/promise-is-undefined-in-ie11-using-babel-polyfill

This has been confirmed working by Scott in IE 11.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,7 +1,11 @@
 // File required for jest, webpack includes different babel config
 module.exports = {
   presets: [
-    '@babel/preset-env'
+    ['@babel/preset-env', {
+      targets: {
+        browsers: ['last 1 version', 'ie >= 11']
+      }
+    }]
   ],
   plugins: [
     '@babel/plugin-transform-runtime'
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "give-web",
   "version": "1.0.0",
   "dependencies": {
+    "@babel/polyfill": "^7.7.0",
     "@babel/runtime-corejs2": "^7.0.0",
     "angular": "1.7.9",
     "angular-cookies": "^1.7.8",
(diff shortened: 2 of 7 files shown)


wrandall22:
I wonder if it affects other versions of IE earlier than 11

Omicron7:
This should polyfill for them as well. I think.

wrandall22:
Even with 
```
targets: {
        browsers: ['last 1 version', 'ie >= 11']
      }
```

babel.config.js
```diff
@@ -1,7 +1,11 @@
 // File required for jest, webpack includes different babel config
 module.exports = {
   presets: [
-    '@babel/preset-env'
+    ['@babel/preset-env', {
+      targets: {
+        browsers: ['last 1 version', 'ie >= 11']
```
OzzieOrca: The default is a little more forgiving than this https://github.com/browserslist/browserslist#full-list
```
> 0.5%, last 2 versions, Firefox ESR, not dead
```

Omicron7: I think `babel.config.js` is only used for jest.
webpack defines it's own target, maybe we want to relax it some? I've got no clue how what's put here changes whats imported in the entrypoints.

OzzieOrca: That makes sense. I just picked one to comment on since they both looked the same. Webpack might go find the babel.config.js if it doesn't have it's own but I'm not sure. I know I've had trouble in the past... Ya, idk. I assume it dynamically does something in @babel/polyfill...

OzzieOrca: Idk what's best either, one version just seemed a little strict.

OzzieOrca: https://browserl.ist/?q=last+1+version%2C+ie+%3E%3D+11 has 51.8% global coverage
https://browserl.ist/?q=defaults%2C+ie+%3E%3D+11 and https://browserl.ist/?q=defaults have 91.34% global coverage

PR #830, opened 2020-01-22
Update ministries to reflect GAiN > unto name change

OzzieOrca:
[Helpscout ticket](https://secure.helpscout.net/conversation/1061869260/405976/)

Should not go to production until Feb 1

### DIFF
--- a/src/app/searchResults/searchResults.ministries.js
+++ b/src/app/searchResults/searchResults.ministries.js
@@ -66,12 +66,6 @@ export default [
     facet: 'ministry'
   },
   {
-    name: 'Global Aid Network (GAiN)',
-    designationNumber: '2807640',
-    path: '/gain-greatestneeds',
-    facet: 'ministry'
-  },
-  {
     name: 'Global Greatest Needs',
     designationNumber: '0703618',
     path: '/2017-globalgreatestneeds.html',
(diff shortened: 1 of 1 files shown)


wrandall22:
Is there a ticket for this? Jira or Helpscout?

wrandall22:
We should put this in stage beforehand so they can look at it as well.

src/app/searchResults/searchResults.ministries.js
```diff
@@ -120,9 +114,9 @@ export default [
     facet: 'ministry'
   },
   {
-    name: 'Unto™',
-    designationNumber: '1071573',
-    path: '/1071573',
+    name: 'unto™',
```
wrandall22: Should use `Unto™` as before. Helpscout ticket was updated.

src/app/searchResults/searchResults.ministries.js
```diff
@@ -121,8 +115,8 @@ export default [
   },
   {
     name: 'Unto™',
```
OzzieOrca: Is the `™` desired? It must have blended in with the quotes when I was looking at it before and I didn't notice it.

wrandall22: I believe it is.

PR #831, opened 2020-01-23
HS-406435: Fix translation of retry buttons and make suggestedAmounts…

OzzieOrca:
… take campaign into account

This still doesn't use `campaignPage` when looking up the Facebook pixel since EP doesn't seem to expose it post checkout for the thank you page.


wrandall22:
Were you able to test this out on stage?

PR #832, opened 2020-01-29
[EP-2157] Create a way for AEM authors to manage ministries

sebamelzi:
This PR depends on https://github.com/CruGlobal/giveaem/pull/414 to get the ministries list.


src/app/searchResults/searchResults.component.js
```diff
@@ -49,7 +48,17 @@ class SearchResultsController {
     this.searchParams.type = type
 
     if (!this.searchParams.keyword && includes(['ministries', '', undefined], this.searchParams.type)) {
-      this.searchResults = ministries
+      const path = this.$location.$$path
```
wrandall22: Not sure what `$location.$$path` is doing. Is it supposed to be `$location.path()`?

src/common/services/api/designations.service.js
```diff
@@ -255,6 +255,26 @@ class DesignationsService {
...
   }
+
+  ministriesList (pagePath) {
+    const path = pagePath + '/jcr:content/content-parsys/designation_search_r.json'
+    return Observable.from(this.$http.get(this.envService.read('apiUrl') + '/' + path))
+      .map((response) => {
+        const ministries = []
+        if (response.data.ministries) {
+            angular.forEach(response.data.ministries, (v, k) => {
+              const m = JSON.parse(v)
```
wrandall22: Not a big fan of single letter variable names.

src/app/searchResults/searchResults.spec.js
```diff
@@ -57,12 +56,13 @@ describe('searchResults', function () {
     })
 
     it('pulls ministry list', () => {
-      jest.spyOn($ctrl.designationsService, 'productSearch').mockReturnValue(Observable.of([]))
+      jest.spyOn($ctrl.designationsService, 'ministriesList').mockReturnValue(Observable.of([]))
```
wrandall22: The test might be more interesting if there was at least one result instead of an empty array.

src/app/searchResults/searchResults.component.js
sebamelzi: Fixed

src/common/services/api/designations.service.js
sebamelzi: Done

src/app/searchResults/searchResults.spec.js
sebamelzi: Added one result to test

sebamelzi:
Solved the lint errors

wrandall22:
Not sure why, but can't get the observable subscribe to work. I am always getting an error response with -1 status. Not very helpful information in the error unfortunately:
Screen Shot 2020-02-03 at 10 55 30 AM

wrandall22:
Looking like it's probably a CORS issue locally.

wrandall22:
That was it, fixed CORS, now I can test it.

wrandall22:
One thing that seems to be an issue (unless I'm doing something wrong) is that I can't see my changes until I publish them. This is because the angular app is pointing at the dispatchers (rightly so) instead of the author.

sebamelzi:
Yes, this is so. The author must publish the changes. We have some problem with that? Or you want the changes to be published automatically?

wrandall22:
The issue is that they won't see their changes (in the author environment) until they are published, which is a bit unfortunate.

sebamelzi:
I understand. One option would be publish the changes automatically (we must publish the whole page). Another simple option maybe would be add a message in the multi-field dialog, something like "Changes in the ministries list will be refresh after the page is published".

wrandall22:
I would be hesitant to publish anything automatically. I'll talk to the stakeholders and see what they think and get back on this. I had hoped we could do something like the SPA editor where the authoring is built into the angular and the users could just author the pieces simply.

PR #836, opened 2020-02-19
EP-2167 - Donor to Contact DSG on Failed Verification Questions

wrandall22:
### OKR
Increase donor satisfaction and reduce costs by reducing duplicate donor accounts

[Jira](https://jira.cru.org/browse/EP-2167)

After meeting with FDS and DSG, they wanted to (as a quick fix) stop creating new accounts when people fail verification questions. This PR is the client side of that request. See [cortex](https://github.com/CruGlobal/give-ep-extensions/pull/587) for the API side changes.

### DIFF
--- a/src/common/components/failedVerificationModal/failedVerificationModal.component.js
+++ b/src/common/components/failedVerificationModal/failedVerificationModal.component.js
@@ -1,0 +1,20 @@
+import angular from 'angular'
+import template from './failedVerificationModal.tpl.html'
+
+const componentName = 'failedVerificationModal'
+
+class FailedVerificationModalController {
+  /* @ngInject */
+  constructor () /* eslint-disable-line no-useless-constructor */ {}
+}
+
+export default angular
+  .module(componentName, [])
+  .component(componentName, {
+    controller: FailedVerificationModalController,
+    templateUrl: template,
+    bindings: {
+      title: '@',
+      onOk: '&'
+    }
+  })
--- a/src/common/components/failedVerificationModal/failedVerificationModal.component.js
+++ b/src/common/components/failedVerificationModal/failedVerificationModal.component.js
@@ -3,15 +3,9 @@ import template from './failedVerificationModal.tpl.html'
 
 const componentName = 'failedVerificationModal'
 
-class FailedVerificationModalController {
-  /* @ngInject */
-  constructor () /* eslint-disable-line no-useless-constructor */ {}
-}
-
 export default angular
   .module(componentName, [])
   .component(componentName, {
-    controller: FailedVerificationModalController,
     templateUrl: template,
     bindings: {
       title: '@',
--- a/src/common/components/userMatchModal/userMatchModal.component.js
+++ b/src/common/components/userMatchModal/userMatchModal.component.js
@@ -12,8 +12,9 @@ const componentName = 'userMatchModal'
 
 class UserMatchModalController {
   /* @ngInject */
-  constructor ($log, gettext, profileService, verificationService, analyticsFactory) {
+  constructor ($log, $window, gettext, profileService, verificationService, analyticsFactory) {
     this.$log = $log
+    this.$window = $window
     this.gettext = gettext
     this.profileService = profileService
     this.verificationService = verificationService
@@ -145,10 +146,15 @@ class UserMatchModalController {
         this.changeMatchState('success')
       },
       error => {
+        this.setLoading({ loading: false })
         this.$log.debug('Failed verification questions', error)
         this.changeMatchState('failure')
       })
     }
+  }
+
+  onFailure () {
+    this.$window.location = '/'
   }
 }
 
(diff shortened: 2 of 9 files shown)


wrandall22:
@OzzieOrca I can't seem to figure out how to close the modal and go back to the home page on click of the Close button.

OzzieOrca:
Communication between components in AngularJS is way too hard. The parent has access to the `close` function of the modal.

Also AngularJS hijacks some of the browser location things. I can't remember if that's just `$location` or `window.location` too. https://docs.angularjs.org/guide/$location#what-does-it-not-do- says `$window.location.href` will cause a whole page refresh. Is it currently refreshing?

wrandall22:
It's currently doing nothing when I hit the close button. However, I am noticing some CORS issues in stage. I think they are unrelated (maybe related to the translation changes).

src/common/components/userMatchModal/userMatchModal.component.js
```diff
@@ -166,6 +174,7 @@ export default angular
       modalTitle: '=',
       setLoading: '&',
       onStateChange: '&',
-      onSuccess: '&'
+      onSuccess: '&',
+      onFailure: '&'
```
OzzieOrca: Oh. Remove this. It's probably overwriting your function. These bindings make things available on `this`. Since the template using using the `on-failure` binding, it's probably undefined.

wrandall22: If I had not defined an `onFailure` function, what would it call? I see several places where `onSuccess` is being called, but is not defined in the controller.

OzzieOrca: Bindings are element attributes that come from where the component is used. Somewhere there's a `<user-match-modal on-success="some angular expression">`. That expression would be run when `this.onSuccess()` is called.

wrandall22:
@OzzieOrca I think it's working now. Going to message DSG/FDS to QA, but I think at least the code can be reviewed at this point.

src/common/components/failedVerificationModal/failedVerificationModal.component.js
```diff
@@ -0,0 +1,20 @@
...
+
+class FailedVerificationModalController {
+  /* @ngInject */
+  constructor () /* eslint-disable-line no-useless-constructor */ {}
+}
+
+export default angular
+  .module(componentName, [])
+  .component(componentName, {
+    controller: FailedVerificationModalController,
```
OzzieOrca: I think you can just delete the controller if it's not needed.
```suggestion

```

src/common/components/userMatchModal/userMatchModal.tpl.html
```diff
@@ -38,10 +38,13 @@
       <button type="button" class="btn btn-md btn-block btn-primary" ng-click="$ctrl.onSuccess()" translate>Close</button>
     </p>
   </div>
-  <div ng-switch-when="success-failure" ng-init="$ctrl.analyticsFactory.track('aa-registration-match-new-account-success')">
-    <p translate>Your account is now activated.</p>
+  <div ng-switch-when="failure" ng-init="$ctrl.analyticsFactory.track('aa-registration-match-failed-questions')">
+    <p translate>
+      Unfortunately we are unable to verify your account.
+      Please contact us at 888-278-7233, option 1, option 1 or email eGift@cru.org.
```
OzzieOrca: Why is this content duplicated in 2 places? I don't quite remember how these modals work...

wrandall22: I just happened to use the same text in the failed verification modal as I did here. This is the text supplied by the stakeholders upon failure of the questions. This location is just text on an existing modal, but the other I had to create a new modal or hijack the register account modal, which seemed weird to me.

src/common/components/failedVerificationModal/failedVerificationModal.component.js
wrandall22: I'll give it a try. Wasn't sure if directives had to have controllers backing them. Was basing this off of the retry modal.

src/common/components/userMatchModal/userMatchModal.tpl.html
wrandall22: Might be able to use the new directive instead?

src/common/components/failedVerificationModal/failedVerificationModal.component.js
wrandall22: We're using `$ctrl.title` and `$ctrl.onOk()`. How would I do those calls if I don't have a controller?

OzzieOrca: Does it break? I assume it attaches the bindings to the view model. I don't think a controller is necessary but I could be wrong. The `this` of the controller is the same as the view model. AngularJS is confusing...

src/common/components/userMatchModal/userMatchModal.tpl.html
OzzieOrca: I don't have my head wrapped around this... The state changes are weird.

PR #837, opened 2020-02-21
Add translation sanitization strategy

OzzieOrca:
- Remove unused translate library directive from searchParam bindings

I clicked around and didn't see that this broke anything obvious. We still have 2 translation libraries running around which is great... :( ~~I pushed this to the staging branch too.~~ Looks like this broke a bunch of tests. Seems very unrelated to the errors but this is Angular after all. Guess I'll have to revisit it later.

https://rollbar.com/Cru/all/items/?offset=0&sort=last_desc&status=active&assigned_user=&framework=&date_from=&date_to=&activated_from=&activated_to=&timezone=America%2FNew_York&enc_query=ZT28xQRidC6me%2FYo595F1mriq3BFD%2B9GHu9iftuVd1Hb3IyP8RbXyXVvdQHK470e&levels=30&projects=78978&date_filtering=seen

### DIFF
--- a/src/app/searchResults/searchResults.tpl.html
+++ b/src/app/searchResults/searchResults.tpl.html
@@ -2,10 +2,10 @@
   <section class="is-top">
     <div class="container">
       <h1 ng-if="!$ctrl.searchParams.type || $ctrl.searchParams.type === 'people'">
-          <span translate ng-if="$ctrl.searchParams.keyword">
+          <span ng-if="$ctrl.searchParams.keyword">
             Search Results for <strong>{{$ctrl.searchParams.keyword}}</strong>
           </span>
-        <span translate ng-if="$ctrl.searchParams.first_name || $ctrl.searchParams.last_name">
+        <span ng-if="$ctrl.searchParams.first_name || $ctrl.searchParams.last_name">
             Search Results for <strong>{{$ctrl.searchParams.first_name}} {{$ctrl.searchParams.last_name}}</strong>
           </span>
         <em ng-if="$ctrl.searchResults.length">({{$ctrl.searchResults.length}} results)</em>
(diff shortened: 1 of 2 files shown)


OzzieOrca:
@wrandall22 Can you review this? I think I'll deploy it tomorrow if you're good with that. It's on staging.

wrandall22:
Is the removal of the empty `translate` attributes on the search page significant? It seems like the change to `app.config.js` is the one that really fixes this issue. Just noting since there are empty `translate` elements all over the place.

OzzieOrca:
It fixes a production bug https://give.cru.org/search-results.html?fName=test&lName=test&type=people
Screen Shot 2020-02-28 at 8 30 10 AM
We've got 2 translation libraries running around and neither one is really used right...

wrandall22:
Do you think that all of these other empty `translate` attributes will cause similar issues elsewhere on the site?

moeloubani:
I've been meaning to get the translation work using gettext pushed into production this week but I've been sidetracked by a caching issue on the WP side of things. I noticed these comments and I don't know if that will help but it is all done here and it has been merged into staging: https://github.com/CruGlobal/give-web/tree/multilingual-using-gettext In the end we didn't end up needing the Spanish form at FamilyLife but I thought it would be good to have that in there in case anyone else needs to translate things.

OzzieOrca:
They seem to cause issues if there are bindings/interpolation inside them. Or sometimes other html elements. Idk. Haven't dug too deeply but I've deleted some of them that were causing issues in previous PRs.

If we were starting from scratch, my suggestion would be to use https://www.i18next.com/. I've been using it on missionhub-web (both AngularJS and React) and missionhub-react-native and it's been great. This is the extent of the AngularJS integration https://github.com/CruGlobal/missionhub-web/blob/master/app/assets/javascripts/angular/filters/translation.filter.js, then you can just use it as a pipe `{{ 'myComponentNamespace:myTranslatedString' | t }}`. It isn't gettext style so you have to manually put the strings in a JSON file somewhere but it's simpler for all other parts of the translation process.

PR #838, opened 2020-02-26
Scroll to top of <branded-checkout> component

OzzieOrca:
Scroll to top of <branded-checkout> component instead of top of page.

https://secure.helpscout.net/conversation/1093344788/419578/


wrandall22:
Looks like the tests failed.

PR #843, opened 2020-04-03
EP-2177 - Enable Angular/React hybrid

wrandall22:
This PR is pretty simple, it just adds the libraries necessary to do the AngularJS/ReactJS hybrid solution.


.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "presets": [
+    "@babel/preset-react"
+  ]
+}
```
OzzieOrca, line 5: Use https://github.com/CruGlobal/give-web/blob/master/babel.config.js instead. It's the newer config file. Idk what having both will do...

package.json
```diff
@@ -24,6 +24,10 @@
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.11",
     "moment": "^2.24.0",
+    "prop-types": "^15.7.2",
```
OzzieOrca, line 27: I'd prefer to use TypeScript.

OzzieOrca: I don't think you need this package anymore.

PR #846, opened 2020-04-24
EP-2163 - Donors cover CC fee

wrandall22:
This PR allows donors to choose to cover the credit card processing fee that is associated with credit card gifts. This way, their entire gift goes to the ministry/staff person/student that they intended.

Note: It may be better to store the fee coverage server side, feel free to push back on the way I decided to do this. I basically just made it work; so there may be better ways to handle this task

[Jira](https://jira.cru.org/browse/EP-2163)


src/common/services/api/order.service.js
```diff
@@ -316,6 +340,87 @@ class Order {
...
+    return this.$filter('number')(newAmount, 2)
+  }
+
+  calculateAmountWithFees (originalAmount) {
+    originalAmount = parseFloat(originalAmount)
+    return (originalAmount * 0.0235) +
+      (originalAmount * 0.0235 * 0.0235) +
+      (originalAmount * 0.0235 * 0.0235 * 0.0235) +
+      (originalAmount * 0.0235 * 0.0235 * 0.0235 * 0.0235) +
+      originalAmount
```
OzzieOrca, line 366: What does this math represent?

wrandall22: It's mentioned in the Jira ticket (linked in the description). However, the basic idea is that it is a 2.35% processing fee.

OzzieOrca: Seems like there's got to be a better and more accurate way to calculate that. Does this work? Maybe there's something I don't have context for. Seems like you can get it down to a multiplication by a constant with you want. My example computations seem to match your document.

Equations:
```
amt_gift     = amt_with_fee - amt_with_fee * 0.0235
amt_gift     = (1 - 0.0235) * amt_with_fee
amt_with_fee = amt_gift / (1 - 0.0235)
amt_with_fee = amt_gift / 0.9765
amt_with_fee = amt_gift * (1 / 0.9765)
amt_with_fee = amt_gift * 1.0240655402
```
Examples:
```
     10.240 =      10 / (1 - 0.0235)
    102.406 =     100 / (1 - 0.0235)
  1,024.065 =   1,000 / (1 - 0.0235)
 10,240.655 =  10,000 / (1 - 0.0235)
102,406.554 = 100,000 / (1 - 0.0235)
```

wrandall22: Hmm, not sure. I was given the calculation from the fund development team, it wasn't something I came up with.

wrandall22: The numbers seem to come out the same way, so maybe that's fine.

OzzieOrca: Maybe the goal is to avoid division, in which case I guess the approximation makes sense. If the precision of the approximation isn't enough, seems like Cru picks up the tab. But looks like you've got some pretty large gifts covered.

wrandall22: Looks like it's slightly different at $1M:

```
1024065.5328550625
1024065.540194572452637 = 1,000,000 / (1 - 0.0235)
```

OzzieOrca: Cuz the approximation runs out of precision and JS division still has precision... The approximation isn't increasing it enough to cover the whole fee. It's off by a whole cent...
`1024065.5328550625 - (1024065.5328550625 * .0235) = 999,999.992`

OzzieOrca: Anyways this was a fun excursion in math but either solution is probably fine. I was just wondering why. Maybe some comment in the code about how the formula was provided by fund development would be good.

wrandall22: I'll run your calculation by FDS and see what they think. They still have to do some more testing so there is time to update it if needed.

wrandall22: Looks like `amt_gift     = amt_with_fee - amt_with_fee * 0.0235` is more precise than `amt_gift     = (1 - 0.0235) * amt_with_fee`. At 102,406.55 the first gives `100000.000075` and the latter gives `99999.996075`

OzzieOrca: Must be a fun artifact of floating point number impentations. Cuz algebraically those should be equal.

wrandall22: Ran the calculations in the console and they both came back with the right number, using the calculator app came with the wrong numbers.

wrandall22: Good news, FDS approved the change in calculation, thanks Scotty!

src/common/services/api/order.service.js
```diff
@@ -359,11 +357,7 @@ class Order {
 
   calculateAmountWithFees (originalAmount) {
     originalAmount = parseFloat(originalAmount)
-    return (originalAmount * 0.0235) +
-      (originalAmount * 0.0235 * 0.0235) +
-      (originalAmount * 0.0235 * 0.0235 * 0.0235) +
-      (originalAmount * 0.0235 * 0.0235 * 0.0235 * 0.0235) +
-      originalAmount
+    return originalAmount / 0.9765
```
OzzieOrca, line 360: Can you document where this came from in the code? Maybe a comment with some of the equations from https://github.com/CruGlobal/give-web/pull/846#discussion_r424686632 showing how this was derived and that the `.0235` is the 2.35% card processing fee. I forsee myself and others being confused what this magical number is.

src/common/services/api/order.service.js
```diff
@@ -407,7 +401,7 @@ class Order {
 
   calculatePriceWithoutFees (originalAmount) {
     originalAmount = parseFloat(originalAmount)
-    const newAmount = originalAmount / this.removeFeesDividend
+    const newAmount = originalAmount - (originalAmount * 0.0235)
```
OzzieOrca, line 404: You've got 2 different versions of this. 
```suggestion
    const newAmount = originalAmount * 0.9765
```
would be the inverse of the equation you're using above in `calculateAmountWithFees`. I'm fine with using the original equation (`amt_gift     = amt_with_fee - amt_with_fee * 0.0235`) but you're using the reduced version above.

wrandall22:
There seems to be intermittent issues with this code where sometimes it will set the amounts and other times it won't. My best guess is that this is due to the asynchronous nature of the code. I tried to use `.pipe` to make it synchronous, but ended up breaking things worse. Any thoughts on how we might more reliably set the amounts?

OzzieOrca:
How does this interact with the API? What amount do you push? You could always store the base value and when you render or send the amount to the API you could run it through a function to add the fee on top depending on the checkbox state... That might touch too many parts of the app though if they don't have access to both those values.

Just glancing at this I'm not sure how to fix it. Sounds like a race condition or one of the 2 state values (both the amount and the cover fee decision seem to be holding the same state but in different forms). Does something come back from the API and overwrite the changes you've made locally? Or are you loading from local storage while updating it? The mutable state and imperitive style of Angular code is hard to follow.

wrandall22:
My intention with the code is that it pushes the newly calculated amount to the API when the user moves on from the payment method page (it does the update cart action, which does a DELETE on the item(s) and POSTs a new item configuration to the API). I thought maybe the users are going too quickly between pages. Each page loads the cart data from the API, which, I assume, will overwrite the amount if the change hasn't made it to the API yet. I'm not putting the amount itself in local storage, but maybe should.

wrandall22:
Do you think it's possible to work `orderService.editGifts` to return an `Observable`, then update the payment method components to continue on in the `subscribe()` function of that `Observable`? Or do you think storing the calculated amount in local storage is a better way to go?

OzzieOrca:
I don't know what's best without digging into the code. editGifts sounds like a one time mutation, it wouldn't query for updates would it? I can take a close look if you need me to.

wrandall22:
Yeah, I would appreciate it. I don't know this stuff as well as you do, and could use your expert opinion.

wrandall22:
Squashed a bunch of stuff together. I'm hopeful that no more bug fixes are needed, so I think a re-review would be good. One thing I will note is that this particular feature makes internationalization more difficult, if we were ever to show non-US money on the give site. The API gives back an internationalized price already, but here we are bypassing the API and assuming US dollars.

src/common/services/api/order.service.js
```diff
@@ -316,6 +380,93 @@ class Order {
...
+  }
+
+  calculateAmountWithoutFees (originalAmount) {
+    originalAmount = parseFloat(originalAmount)
+    return originalAmount * this.FEE_DERIVATIVE
+  }
+
+  editGifts (cartData) {
+    const observables = []
+    angular.forEach(cartData.items, item => {
```
OzzieOrca, line 462: `cartData.items.map(...)` is probably cleaner than a `forEach` and a `push`. As long as `cartData.items` is always an array. Otherwise you could do `(cartData.items || []).map(...)`

wrandall22:
I was able to reduce some code duplication, but not sure how I would go about doing more.

PR #849, opened 2020-05-28
SIEBUPG-296 - Fix edit recurring gift checkout

wrandall22:
[Jira](https://jira.cru.org/browse/SIEBUPG-296)

Currently (in production) the code is not accounting for the fact that recurring donations have a code of `{designationNumber}_{frequency}` and try to lookup the infinity.json of a non-existing AEM node. This removes the frequency when calculating the path in AEM.

### DIFF
--- a/src/common/services/api/designations.service.js
+++ b/src/common/services/api/designations.service.js
@@ -185,6 +185,11 @@ class DesignationsService {
       .slice(0, 5)
       .join('/')
 
+    const underscorePosition = code.indexOf('_')
+    if (underscorePosition > 0) {
+      code = code.substring(0, underscorePosition)
+    }
+
     return campaignPage
       ? `/content/give/us/en/campaigns/${c}/${code}/${campaignPage}.infinity.json`
       : `/content/give/us/en/designations/${c}/${code}.infinity.json`
--- a/src/common/services/api/designations.service.js
+++ b/src/common/services/api/designations.service.js
@@ -185,14 +185,11 @@ class DesignationsService {
       .slice(0, 5)
       .join('/')
 
-    const underscorePosition = code.indexOf('_')
-    if (underscorePosition > 0) {
-      code = code.substring(0, underscorePosition)
-    }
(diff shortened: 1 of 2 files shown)


src/common/services/api/designations.service.js
```diff
@@ -185,6 +185,11 @@ class DesignationsService {
       .slice(0, 5)
       .join('/')
 
+    const underscorePosition = code.indexOf('_')
+    if (underscorePosition > 0) {
+      code = code.substring(0, underscorePosition)
+    }
```
OzzieOrca, line 191: What you have is fine but it could be a one liner (which is maybe more computationally since it processes the whole string instead of just to the first `_`).
```suggestion
    [code] = code.split('_');
```

wrandall22: Hmm, that one liner appears to break all the tests. It works if I do this...
```
const other = code.split('_')

    return campaignPage
      ? `/content/give/us/en/campaigns/${c}/${other[0]}/${campaignPage}.infinity.json`
      : `/content/give/us/en/designations/${c}/${other[0]}.infinity.json`
```

wrandall22: Also, this works:
```
const [other] = code.split('_')

return campaignPage
      ? `/content/give/us/en/campaigns/${c}/${other}/${campaignPage}.infinity.json`
      : `/content/give/us/en/designations/${c}/${other}.infinity.json`
```

OzzieOrca: Something in the test tooling must not support array destructuring (and other newer JS features). Oh well, that's fine.
```suggestion
    code = code.split('_')[0];
```
is the same thing.

I like that you gave the variable a new name since I like to try and keep things immutable. This codebase doesn't value that.

OzzieOrca: Huh that's interesting. Maybe assignment is a different feature than variable declarations. Whatever. I'm fine with whatever. This is trivial.

PR #851, opened 2020-06-03
HS-453518 - Fix Newsletter Issue

wrandall22:
[Helpscout](https://secure.helpscout.net/conversation/1183539709/453518/)

When we went from jQuery `3.4.1` to `3.5.1` this form broke. After some testing locally, making this tag a non-empty tag fixed the problem.


wrandall22:
Can be seen working in stage [here](https://give-stage2.cru.org/0559826).

PR #852, opened 2020-06-05
Handle Siebel being down

wrandall22:
This PR is to give users a better experience when Siebel is down and they attempt to login. This is dependent upon https://github.com/CruGlobal/give-ep-extensions/pull/604

### DIFF
--- a/src/common/components/signInForm/signInForm.component.js
+++ b/src/common/components/signInForm/signInForm.component.js
@@ -51,6 +51,14 @@ class SignInFormController {
               this.signInState = 'identity'
               this.onFailure()
           }
+        } else if (error && error.data && error.data.code && error.data.code === 'SIEB-DOWN') {
+          if (error && error.config && error.config.data && error.config.data.password) {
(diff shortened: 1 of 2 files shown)


wrandall22:
Siebel database (and thus login) will be down tonight, so hoping to get this merged before then.

src/common/components/signInForm/signInForm.component.js
```diff
@@ -51,6 +51,14 @@ class SignInFormController {
               this.signInState = 'identity'
               this.onFailure()
           }
+        } else if (error && error.data && error.data.code && error.data.code === 'SIEB-DOWN') {
```
OzzieOrca, line 54: Does the tooling support this? 
```suggestion
        } else if (error?.data?.code === 'SIEB-DOWN') {
```
You can merge as is since it's probably not worth the time testing it.

wrandall22: Hmm, not sure, there isn't any typescript in this project as far as I know.

PR #854, opened 2020-06-16
EP-2187: Remove search tabs

OzzieOrca:
Should I merge this to staging? Not sure what the release process has been here lately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "mini-css-extract-plugin": "^0.8.0",
     "moment-locales-webpack-plugin": "^1.1.0",
     "ngtemplate-loader": "^2.0.1",
-    "node-sass": "4.12.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "mini-css-extract-plugin": "^0.8.0",
     "moment-locales-webpack-plugin": "^1.1.0",
     "ngtemplate-loader": "^2.0.1",
-    "node-sass": "4.12.0",
+    "node-sass": "4.14.1",
```
OzzieOrca, line 49: `node-sass` always seems to need to be upgrade to work with newer versions of node...

wrandall22:
Yeah, merge to staging so the stakeholders can see it in action. When they give the okay, we can merge to production.

PR #858, opened 2020-07-27
Fix cart count by replacing angular-gettext pluralize with ng-pluralize

OzzieOrca:
https://secure.helpscout.net/conversation/1233344938/468614/?

https://angular-gettext.rocketeer.be/dev-guide/annotate/#plurals isn't working for some reason. Maybe because we're using https://angular-translate.github.io/ as the main translation library but angular-gettext is still around. This string doesn't appear to be translated anyways.

### DIFF
--- a/src/common/components/nav/navCart/navCart.tpl.html
+++ b/src/common/components/nav/navCart/navCart.tpl.html
@@ -52,7 +52,15 @@
 
   <div class="row repeating-row sub-total" ng-if="$ctrl.hasItems">
     <div class="col-xs-9">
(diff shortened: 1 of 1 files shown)


wrandall22:
Have you been able to test this in stage yet?

OzzieOrca:
Just pushed it to staging. I tested locally and it was always showing 1 before this change.

wrandall22:
When it is live in staging, can you tell them (in the HS ticket) to test in stage?

OzzieOrca:
Should I merge this? Haven't heard back on helpscout...

wrandall22:
At this point, I would go ahead and do it.

PR #863, opened 2020-10-05
EP-2190 - Branded checkout cover fees

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2190)
[OKR - Improve online giving experience, and provide more giving and engagement opportunities: Encourage generous giving from our donors](https://docs.google.com/presentation/d/1zOAeUeY0F_FgjJmsejfsdVGJ3UL9H26pt__z6N2CmU8/edit?ts=5f43dd6f&pli=1#slide=id.g928c48312d_1_198)

This PR adds the "cover fees" checkbox to branded checkout as an opt-in. We went with an opt-in solution because adding a new section to the branded checkout might look bad on individual client implementations depending on how they did their styling.

This uses a separate set of local storage variables in order to not affect the standard checkout.

### DIFF
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -6,6 +6,7 @@ import contactInfo from 'common/components/contactInfo/contactInfo.component'
 import checkoutStep2 from 'app/checkout/step-2/step-2.component'
 
 import cartService from 'common/services/api/cart.service'
+import orderService from 'common/services/api/order.service'
 
 import template from './branded-checkout-step-1.tpl.html'
 
@@ -13,9 +14,10 @@ const componentName = 'brandedCheckoutStep1'
 
 class BrandedCheckoutStep1Controller {
   /* @ngInject */
-  constructor ($log, cartService) {
+  constructor ($log, cartService, orderService) {
     this.$log = $log
     this.cartService = cartService
+    this.orderService = orderService
   }
 
   $onInit () {
@@ -62,6 +64,13 @@ class BrandedCheckoutStep1Controller {
 
         // add campaign page
         this.itemConfig['campaign-page'] = this.campaignPage
+
+        if (this.orderService.retrieveBrandedCoverFeeDecision()) {
+          this.itemConfig.priceWithFee = item.price
+          this.itemConfig.price = item.price
+          this.itemConfig.amountWithFee = item.amount
+          this.itemConfig.coverFees = true
+        }
       }
       this.loadingProductConfig = false
     },
@@ -152,7 +161,8 @@ export default angular
     productConfigForm.name,
     contactInfo.name,
     checkoutStep2.name,
-    cartService.name
+    cartService.name,
+    orderService.name
   ])
   .component(componentName, {
     controller: BrandedCheckoutStep1Controller,
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -65,7 +65,7 @@ class BrandedCheckoutStep1Controller {
         // add campaign page
         this.itemConfig['campaign-page'] = this.campaignPage
 
-        if (this.orderService.retrieveBrandedCoverFeeDecision()) {
+        if (this.orderService.retrieveCoverFeeDecision()) {
           this.itemConfig.priceWithFee = item.price
           this.itemConfig.price = item.price
           this.itemConfig.amountWithFee = item.amount
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -177,6 +177,7 @@ export default angular
       donorDetails: '<',
       defaultPaymentType: '<',
       hidePaymentTypeOptions: '<',
+      showCoverFees: '<',
       next: '&',
       onPaymentFailed: '&'
     }
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -66,8 +66,6 @@ class BrandedCheckoutStep1Controller {
         this.itemConfig['campaign-page'] = this.campaignPage
 
         if (this.orderService.retrieveCoverFeeDecision()) {
-          this.itemConfig.priceWithFee = item.price
-          this.itemConfig.price = item.price
           this.itemConfig.amountWithFee = item.amount
           this.itemConfig.coverFees = true
         }
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -45,6 +45,7 @@ class BrandedCheckoutStep1Controller {
         break
     }
     this.itemConfig['recurring-day-of-month'] = this.day
+    this.itemConfig.frequency = this.frequency
   }
 
   initCart () {
--- a/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
+++ b/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
@@ -167,6 +167,7 @@ export default angular
       defaultPaymentType: '<',
       hidePaymentTypeOptions: '<',
       cartData: '<',
+      brandedCheckoutItem: '<',
       onPaymentFormStateChange: '&',
       onLoad: '&'
     }
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -31,6 +31,8 @@ import { giveGiftParams } from '../giveGiftParams'
 import loading from 'common/components/loading/loading.component'
 import analyticsFactory from 'app/analytics/analytics.factory'
 
+import { brandedCheckoutAmountUpdatedEvent } from 'common/components/paymentMethods/coverFees/coverFees.component'
+
 import template from './productConfigForm.tpl.html'
 
 export const brandedCoverFeeCheckedEvent = 'brandedCoverFeeCheckedEvent'
@@ -67,9 +69,9 @@ class ProductConfigFormController {
     this.$rootScope.$on(brandedCoverFeeCheckedEvent, () => {
       this.initItemConfig()
       if (this.selectableAmounts.includes(this.itemConfig.amount)) {
-        this.changeAmount(this.itemConfig.amount)
+        this.changeAmount(this.itemConfig.amount, true)
       } else {
-        this.changeCustomAmount(this.itemConfig.amount)
+        this.changeCustomAmount(this.itemConfig.amount, true)
       }
     })
   }
@@ -249,20 +251,30 @@ class ProductConfigFormController {
     }
   }
 
-  changeAmount (amount) {
+  changeAmount (amount, retainCoverFees) {
     this.itemConfigForm.$setDirty()
     this.checkAmountChanged(amount)
     this.itemConfig.amount = amount
     this.customAmount = ''
     this.customInputActive = false
+    if (!retainCoverFees && this.amountChanged) {
+      this.orderService.clearBrandedCoverFees()
+      this.itemConfig.coverFees = false
+      this.$scope.$emit(brandedCheckoutAmountUpdatedEvent)
+    }
     this.updateQueryParam({ key: giveGiftParams.amount, value: amount })
   }
 
-  changeCustomAmount (amount) {
+  changeCustomAmount (amount, retainCoverFees) {
     this.checkAmountChanged(amount)
     this.itemConfig.amount = amount
     this.customAmount = amount
     this.customInputActive = true
+    if (!retainCoverFees && this.amountChanged) {
+      this.orderService.clearBrandedCoverFees()
+      this.itemConfig.coverFees = false
+      this.$scope.$emit(brandedCheckoutAmountUpdatedEvent)
+    }
     this.updateQueryParam({ key: giveGiftParams.amount, value: amount })
   }
 
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -210,6 +210,11 @@ class ProductConfigFormController {
     this.errorChangingFrequency = false
     const lastFrequency = this.productData.frequency
     this.productData.frequency = product.name
+
+    if (this.isBrandedCheckout) {
+      this.itemConfig.frequency = product.display
+    }
+
     this.updateQueryParam({ key: giveGiftParams.frequency, value: product.name })
     if (product.selectAction) {
       this.changingFrequency = true
@@ -351,6 +356,7 @@ export default angular
       isEdit: '<',
       uri: '<',
       defaultFrequency: '<',
+      isBrandedCheckout: '<',
       disableSessionRestart: '@',
       updateQueryParam: '&',
       submitted: '<',
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -283,6 +283,9 @@ class ProductConfigFormController {
     if (this.itemConfig.amount && amount) {
       this.amountChanged = this.itemConfig.amount !== amount
     }
+    if (!this.itemConfig.amount && amount) {
+      this.amountChanged = true
+    }
   }
 
   changeStartDay (day, month) {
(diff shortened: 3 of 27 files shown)


reldredge71 (commented):
This looks good!  I just have a few clarifying questions about Angular.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -166,6 +177,7 @@ export default angular
       donorDetails: '<',
       defaultPaymentType: '<',
       hidePaymentTypeOptions: '<',
+      showCoverFees: '<',
```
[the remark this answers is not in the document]
wrandall22: Yes, this is an Angular thing. Basically, it tells the component (in the HTML) that `show-cover-fees` is a valid attribute on the component and that it can accept expressions as the value of the attribute. [Here is a blog](https://blog.krawaller.se/posts/dissecting-bindings-in-angularjs/) related to it.

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -261,6 +292,10 @@ class ProductConfigFormController {
     this.updateQueryParam({ key: giveGiftParams.month, value: month })
   }
 
+  updateDecimalPlaces () {
+    this.customAmount = this.$filter('number')(this.customAmount, 2)
```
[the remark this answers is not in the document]
wrandall22: This expression is basically turning a number into a formatted string (e.g. `51.2` into `'51.20'`

src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
```diff
@@ -56,15 +58,45 @@ describe('product config form component', function () {
...
 
       expect($ctrl.initItemConfig).toHaveBeenCalled()
       expect($ctrl.loadData).toHaveBeenCalled()
       expect($ctrl.waitForFormInitialization).toHaveBeenCalled()
+      expect($ctrl.$rootScope.$on).toHaveBeenCalledWith(brandedCoverFeeCheckedEvent, expect.any(Function))
+      $ctrl.$rootScope.$on.mock.calls[0][1]()
+    })
+
+    it('should handle brandedCoverFeeCheckedEvent for selectable amounts', () => {
+      jest.spyOn($ctrl, 'initItemConfig').mockImplementation(() => {})
```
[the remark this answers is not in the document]
wrandall22: I do not, thanks for catching it.

src/app/productConfig/productConfigForm/productConfigForm.component.js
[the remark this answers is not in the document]
reldredge71: Gotcha, I did notice that the output was a numeric string with two decimal points, I was wondering how `filter` did that.  I was thinking that `filter` was for getting a subset of values from an array, but it looks like the syntax `$filter('number')` is a specific method for formatting numeric strings.

reldredge71 (approved):

src/app/productConfig/productConfigForm/productConfigForm.component.js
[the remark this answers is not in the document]
wrandall22: Yeah, the documentation for it is [here](https://docs.angularjs.org/api/ng/filter/filter)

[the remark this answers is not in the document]
wrandall22: Yeah, the documentation for `$filter` is [here](https://docs.angularjs.org/api/ng/service/$filter) and [here](https://docs.angularjs.org/api/ng/filter/number)

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -232,20 +253,30 @@ class ProductConfigFormController {
...
     this.updateQueryParam({ key: giveGiftParams.amount, value: amount })
   }
 
-  changeCustomAmount (amount) {
+  changeCustomAmount (amount, retainCoverFees) {
     this.checkAmountChanged(amount)
     this.itemConfig.amount = amount
     this.customAmount = amount
     this.customInputActive = true
+    if (!retainCoverFees && this.amountChanged) {
```
wrandall22, line 275: For some reason, this doesn't work the first time. Maybe `this.amountChanged` is false.

wrandall22: It looks like `itemConfig.amount` is undefined.

src/common/services/api/order.service.js
```diff
@@ -352,6 +352,27 @@ class Order {
     return cartData
   }
 
+  storeBrandedCoverFeeDecision (coverFees) {
+    this.localStorage.setItem('brandedCoverFees', angular.toJson(coverFees))
```
OzzieOrca, line 356: I saw in your comment that you purposely had this using a different key. Seems like branded and give.cru.org should always be running on a different domain and local storage won't be shared (except maybe in development where it shouldn't matter.

If we think there's a reason to keep it seperate, that's fine, just seems like more complexity and code to maintain that we could potentially reuse.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -45,6 +47,7 @@ class BrandedCheckoutStep1Controller {
         break
     }
     this.itemConfig['recurring-day-of-month'] = this.day
+    this.itemConfig.frequency = this.frequency
```
OzzieOrca, line 50: This seems unrelated but it must be needed for some reason? AngularJS makes the flow of data really hard to follow. Maybe something has always been broken?

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -167,6 +167,7 @@ export default angular
       defaultPaymentType: '<',
       hidePaymentTypeOptions: '<',
       cartData: '<',
+      brandedCheckoutItem: '<',
```
OzzieOrca, line 170: This seems really annoying to have to pass around 😞 Could we pass the `showCoverFees` bool around instead (or maybe stash it somewhere? In React this is called prop drilling and can be solved by using context. Idk if there's a good solution in AngularJS.

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -60,6 +65,16 @@ class ProductConfigFormController {
     this.initItemConfig()
     this.loadData()
     this.waitForFormInitialization()
+
+    this.$rootScope.$on(brandedCoverFeeCheckedEvent, () => {
```
OzzieOrca, line 69: Why does this need an event? How did the regular checkout cover fee do it?

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -210,6 +226,11 @@ class ProductConfigFormController {
     this.errorChangingFrequency = false
     const lastFrequency = this.productData.frequency
     this.productData.frequency = product.name
+
+    if (this.isBrandedCheckout) {
+      this.itemConfig.frequency = product.display
```
OzzieOrca, line 231: Same question as before, was this something that has been broken in branded checkout for a while? Or did cover fees change the behavior of something?

src/app/productConfig/productConfigForm/productConfigForm.component.js
OzzieOrca, line 296: Is this a string and not a number? How is the decimal place getting off? Maybe we should parse it as a number if we're getting it from the API as a string... Unless the math is causing rounding errors. You seem to have had to call this a bunch. Maybe this filter belongs in the template instead.

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -351,6 +386,7 @@ export default angular
       isEdit: '<',
       uri: '<',
       defaultFrequency: '<',
+      isBrandedCheckout: '<',
```
OzzieOrca, line 389: You can use `envService.read('isBrandedCheckout')` instead of having to pass it around.

src/common/components/paymentMethods/coverFees/coverFees.component.js
```diff
@@ -20,16 +24,16 @@ class CoverFeesController {
   }
 
   $onInit () {
+    this.$rootScope.$on(brandedCheckoutAmountUpdatedEvent, () => {
+      this.feesCalculated = false
+      this.$onInit()
```
OzzieOrca, line 29: Will calling this again add another event listener that will start calling this multiple times?

OzzieOrca (changes requested):
Idk Bill, this seems like a whole lot of code. I think it's going to be hard to maintain in the future.

Maybe we can reuse more code between branded and regular checkout...

Maybe I should have pushed back a little more on the original cover fees PR but I think with this implementation it makes the code hard to follow. It would be sorta nice if we could store the checkbox decision in local storage as a boolean, use that as a template filter to increase the prices just on the view layer (and not change any values in the data later), and then send the boolean along with the checkout submit request and let the API do the final math. Idk, maybe we need to make sure the user reviews the final API adjusted values before submitting... But having to constantly update product prices and make sure everything refreshes after seems a little unwieldy.

Maybe the API could always serve 2 prices for products (1 with and 1 without cover fees). On the client we could just pick whichever one matches the local storage boolean at render time (in the template). And we could send that boolean to the server sometime. If we did it at checkout submission, I guess it would have to return 2 prices for all the totals too.

If you've got this working, we can just go with it. I don't really want to make anyone spend more time in the depths of AngularJS but maybe there's something we can do to make this a little cleaner...

src/common/services/api/order.service.js
wrandall22: That makes sense. My only hesitation is the testing that FDS will be doing in stage. Hitting S3 directly is not really usable for user acceptance testing, and I've put a small branded checkout directly onto give-stage2. This would have conflicting values when going between branded and standard checkout. But, as you say, it is just for testing.

src/app/branded/step-1/branded-checkout-step-1.component.js
wrandall22: It seems when the initial implementation was done, just the bare minimum was brought into this branded checkout `itemConfig` object. It's not that something was broken, but that this value wasn't used, so it wasn't put in the object.

src/common/components/paymentMethods/coverFees/coverFees.component.js
wrandall22: Hmm, possibly. It might be getting called twice. I think I started with it in the constructor, but something wasn't working. I might try that again.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: It seems when the initial implementation was done, just the bare minimum was brought into this branded checkout `itemConfig` object. It's not that something was broken, but that this value wasn't used, so it wasn't put in the object.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: It is a number typically, and as a number, it drops the ending `0` (numbers prefer `51.2` over `51.20`). When I put it in the template, it caused real problems with the user typing in the number. It would try to format it on every key stroke which was very annoying.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: They both use events. Regular checkout uses the `cartUpdatedEvent`.

wrandall22: The major difference is that in branded checkout, the product config form is always showing on the page; whereas in regular checkout, it is brought up when the user goes to edit a gift (which calls `$onInit()`)

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
wrandall22: So, there isn't really a need to pass around the boolean since it is in local storage. We could maybe put this branded checkout item also in local storage and keep it updated there instead. This object holds the frequency and amounts that I care about primarily. I think that would end up with more code, though.

wrandall22:
I think long term I do want to push some of this to the API side, especially since they want to know in the gift confirmation email whether or not someone chose to cover fees. However, I'm concerned that it will take longer than I have in T3 to accomplish.

wrandall22:
> It would be sorta nice if we could store the checkbox decision in local storage as a boolean, use that as a template filter to increase the prices just on the view layer (and not change any values in the data later)

This sounds interesting. We're already storing it as a boolean in local storage. I don't really know how to turn that into a template filter.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: @OzzieOrca do you think I should default the amount to `50` (e.g at line 88) or change the logic for checking the amount changed to not require a defined itemConfig amount (line 285)? I'm leaning towards the latter.

src/common/services/api/order.service.js
wrandall22: Going to change it to use the regular cover-fees local storage.

src/app/productConfig/productConfigForm/productConfigForm.component.js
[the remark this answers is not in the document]
wrandall22: So, it appears that this commit broke the amount validation

wrandall22:
It looks like I might be able to drop `priceWithFee` and, in the case of branded checkout, also `price` if I use `amountWithFee` with a currency filter.

wrandall22:
@OzzieOrca any other comments on this? I reduced the code a bit. I feel like there should be a way to use two-way binding more effectively in terms of getting the `brandedCheckoutItem` passed around, but not seeing how so far.

src/common/components/paymentMethods/coverFees/coverFees.tpl.html
```diff
@@ -1,4 +1,20 @@
...
+      I want my gift to go even further for ministry!
+      To help cover credit/debit card processing fees, please adjust my gift amount to
+      {{$ctrl.brandedCheckoutItem.amountWithFee | currency}}
+      <span class="text-lowercase"
+            ng-if="$ctrl.brandedCheckoutItem.frequency && $ctrl.brandedCheckoutItem.frequency.toLowerCase() !== 'single'">
+        {{$ctrl.brandedCheckoutItem.frequency}}.
+      </span>
+    </span>
+  </label>
   <label ng-if="$ctrl.cartData.items.length === 1">
```
OzzieOrca, line 18: Can you share this HTML? Maybe in the component, you can decide which set of data to use and set a `$ctrl.item` variable or something.

src/common/services/api/order.service.js
```diff
@@ -422,27 +415,31 @@ class Order {
...
-        }
-      }
-
-      item.amount = round(parseFloat(newAmount), 2)
-      item.price = `$${this.$filter('number')(newAmount, 2)}`
-      newTotal += item.amount
+      newTotal += this.updatePrice(item, cartData.coverFees)
     })
 
     cartData.cartTotal = newTotal
```
OzzieOrca, line 421: This seems like a good scenario for reduce:
```ts
cartData.cartTotal = cartData.items.reduce(
  (total, item) => total + this.updatePrice(item, cartData.coverFees),
  0
);
```

OzzieOrca (approved):
Looks a lot simpler than it was, thanks!

src/common/components/paymentMethods/coverFees/coverFees.tpl.html
wrandall22: So, if I share the HTML; the `ng-model` doesn't do what it needs to do. Do you know off-hand a way to keep that `coverFees` attribute in sync with both objects without doing it manually?

wrandall22: I figured something out. It seems that because the `coverFees` is at the cart level not the item level, I need to update it there in `updatePrices()`. But, on `brandedCheckoutItem` I don't need to because it is at the item level.

OzzieOrca: If you use the parent object, it should be passed by reference and update. Unless I'm understanding it wrong or AngularJS is doing something weird. Something like:
```ts
$onInit () {
  this.objectContainingCoverFees = this.brandedCheckoutItem || this.cartData;
}
```
and then
```ts
ng-model="$ctrl.objectContainingCoverFees.coverFees"
```

You could maybe use `ng-change` and `ng-value` instead of `ng-model`. Or bind it to a temporary variable you initialize and then update in `updatePrices`.

That's weird the data is stored inconsistently. Is that some API thing or have you been putting it in 2 different places?

wrandall22: It's because branded checkout doesn't have a cart. It is just a single item. For regular checkout, the `coverFees` decision covers all gifts in the cart.

Regular checkout:
```
cart ->
  coverFees: boolean,
  items: [
    { amount, frequency, amountWithFee, etc }
  ]
```

Branded Checkout:
```
item: {
  coverFees: boolean,
  amount, frequency, amountWithFee, etc
}
```

OzzieOrca (approved):
👍

PR #864, opened 2020-10-16
SIEBUPG-299: Fix modal close icon

OzzieOrca:
https://jira.cru.org/browse/SIEBUPG-299

These images should really be run through Webpack's file loader. I tried quickly but there's a lot of legacy AEM CSS in this project and I was having trouble wiring up all the icon font css imports.

### DIFF
--- a/src/assets/scss/_modal.scss
+++ b/src/assets/scss/_modal.scss
@@ -17,7 +17,7 @@
     }
 
     button.close {
-      background: url("assets/img/close-mobile.png");
+      background: url("/assets/img/close-mobile.png");
       background-size: contain;
       width: 27px;
       height: 27px;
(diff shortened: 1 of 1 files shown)


wrandall22 (approved):

wrandall22:
This worked in stage, but so far (after 28 minutes) not seeing the effect in prod. Maybe it's cache.

OzzieOrca:
I did a Cloudfront invalidation on `manifest.json` and it seems to be working now. https://console.aws.amazon.com/cloudfront/home?region=us-east-1#distribution-settings:E51L08TW3241I

The cache busting on this repo is relying on a manual Cloudfront invalidation that seems to be broken https://travis-ci.com/github/CruGlobal/give-web/builds/191240985#L1063. It really should set up sane caching rules up front and have unique resource hashes that can be cached long term and change when updated.

wrandall22:
Thanks for doing that. Yeah, there are probably several things about this project that should/could be improved.

PR #866, opened 2020-11-12
EP-2191 - Migrate-To-GA

TheNoodleMoose:
This pr will be adding all of the new GTM e-commerce events for the GiveSite. I plan to work through the events in batches. The documentation that I am working through can be found [here](https://docs.google.com/spreadsheets/d/14_AEOJ_CeP_c1H9nE8TbS5wXOzUryC5E2uQBiFQJuKc/edit#gid=1140205977).

[Jira](https://jira.cru.org/browse/EP-2191)

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -199,6 +199,58 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // Error caught in analyticsFactory.cartView
       }
     },
+    checkoutStepEvent: function (step, cart) {
+      const cartObject = cart.items.map((cartItem) => {
+        return {
+          name: cartItem.displayName.toLowerCase(),
+          id: cartItem.code,
+          price: cartItem.amount.toString(),
+          branch: 'cru',
+          category: cartItem.designationType.toLowerCase(),
+          variant: cartItem.frequency.toLowerCase(),
+          quantity: '1'
+        }
+      })
+      try {
+        if (typeof $window.dataLayer !== 'undefined') {
+          $window.dataLayer.push({
+            event: 'checkout-step',
+            cartId: cart.id,
+            ecommerce: {
+              currencyCode: 'USD',
+              checkout: {
+                actionField: {
+                  step: step,
+                  option: ''
+                },
+                products: [
+                  ...cartObject
+                ]
+              }
+            }
+          })
+        }
+      } catch (e) {
+        // Error caught in analyticsFactory.checkoutStepEvent
+      }
+    },
+    checkoutStepOptionEvent: function (option, step) {
+      try {
+        $window.dataLayer.push({
+          event: 'checkout-option',
+          ecommerce: {
+            checkout_option: {
+              actionField: {
+                step: step.toLowerCase(),
+                option: option.toLowerCase()
+              }
+            }
+          }
+        })
+      } catch (e) {
+        // Error caught in analyticsFactory.checkoutStepOptionEvent
+      }
+    },
     editRecurringDonation: function (giftData) {
       try {
         var frequency = ''
@@ -306,11 +358,11 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // Error caught in analyticsFactory.getSetProductCategory
       }
     },
-    giveGiftModal: function (productCode) {
+    giveGiftModal: function (productData) {
       try {
         var product = [{
           productInfo: {
-            productID: productCode
+            productID: productData.code
           },
           attributes: {
             siebel: {
@@ -320,6 +372,23 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         }]
 
         $window.digitalData.product = product
+        $window.dataLayer.push({
+          event: 'give-gift-modal',
+          ecommerce: {
+            currencyCode: 'USD',
+            detail: {
+              products: [{
+                name: productData.displayName.toLowerCase(),
+                id: productData.code,
+                price: undefined,
+                brand: 'cru',
+                category: productData.designationType.toLowerCase(),
+                variant: undefined,
+                quantity: '1'
+              }]
+            }
+          }
+        })
         this.setEvent('give gift modal')
         this.pageLoaded()
       } catch (e) {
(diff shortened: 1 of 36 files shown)


TheNoodleMoose:
This first batch adds events for add-to-cart, remove-from-cart, view-cart, and view-mini-cart. Additional events that do not require additional data have also been added.

src/app/analytics/analytics.factory.js
```diff
@@ -150,7 +191,12 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
       try {
         // Build products variable
         this.buildProductVar(cartData)
-
+        // Send GTM Advance Ecommerce event
+        if (typeof $window.dataLayer !== 'undefined') {
+          $window.dataLayer.push({
+            event: callType === 'mini-cart' ? 'view-mini-cart' : 'view-cart'
+          })
+        }
```
TheNoodleMoose, line 199: So for this, I noticed ```callType``` wasn't actually being used so I decided to use it. They want to differentiate between viewing the mini cart and then actually going to view the cart. So where I call the analytics function for the view-mini-cart event, I pass 'mini-cart' as a parameter so we can differentiate the two events. Let me know if I should update the ```callType``` parameter name to something else.

src/app/profile/yourGiving/editRecurringGifts/step4/confirmRecurringGifts.component.js
```diff
@@ -32,6 +32,7 @@ class ConfirmRecurringGiftsController {
 
   saveChanges () {
     this.analyticsFactory.track('aa-edit-recurring-submit')
+    this.analyticsFactory.trackGTM('ga-edit-recurring-submit')
```
TheNoodleMoose, line 35: A lot of these events don't need any additional data, so I just used the ```trackGTM``` function and changed the event name to align with google analytics.

src/common/services/session/sessionModal.service.js
```diff
@@ -46,12 +46,34 @@ const SessionModalService = /* @ngInject */ function ($uibModal, $log, modalStat
...
+          }
         })
     }
 
     if (options.openAnalyticsEvent) {
       currentModal.opened.then(() => {
         analyticsFactory.track(options.openAnalyticsEvent)
+        switch (type) {
+          case 'sign-in':
+          case 'sign-up':
```
TheNoodleMoose, line 68: I didn't know if these needed to be two separate events or not. There wasn't anything in the documentation so I just used the same analytics call for both. I'm going to bring it up to Clark though.

src/app/analytics/analytics.factory.js
```diff
@@ -110,6 +110,26 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
+        // Send GTM Advance Ecommerce event
+        if (typeof $window.dataLayer !== 'undefined') {
+          $window.dataLayer.push({
+            event: 'add-to-cart',
+            ecommerce: {
+              currencyCode: 'USD',
+              add: {
+                products: [{
+                  name: productData.displayName.toLowerCase(),
+                  id: productData.designationNumber.toLowerCase(),
```
wrandall22, line 122: Do you need to lowercase a number?

TheNoodleMoose: That I do not, haha. I'll remove that.

src/app/analytics/analytics.factory.js
```diff
@@ -150,7 +191,12 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
       try {
         // Build products variable
         this.buildProductVar(cartData)
-
+        // Send GTM Advance Ecommerce event
+        if (typeof $window.dataLayer !== 'undefined') {
```
wrandall22, line 195: So, `cartView` isn't currently called anywhere. Maybe if you aren't going to use the `cartData` parameter, it should be removed from the definition?

src/app/analytics/analytics.factory.js
wrandall22: Since the current code doesn't use it, you could make it a boolean if you wanted

TheNoodleMoose: Yeah, that sounds good. I'll change it to ```isMiniCart``` and then have a default value of false.

src/common/components/nav/navCartIcon.component.js
```diff
@@ -9,8 +9,9 @@ const componentName = 'navCartIcon'
 
 class NavCartIconController {
   /* @ngInject */
-  constructor ($rootScope) {
+  constructor ($rootScope, analyticsFactory) {
     this.$rootScope = $rootScope
+    this.analyticsFactory = analyticsFactory
```
wrandall22, line 14: You also need to add this to the `.module` piece at the bottom (see `stopStartRecurringGifts.modal.component.js`)

wrandall22: And you need to import it at the top I believe.

src/app/analytics/analytics.factory.js
```diff
@@ -540,6 +586,16 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
       } catch (e) {
         // Error caught in analyticsFactory.track
       }
+    },
+    trackGTM: function (eventName, data) {
```
wrandall22, line 590: So far you aren't sending `data` in any of the calls to this function.  Are you planning to use it in the future?

src/common/components/nav/navCartIcon.component.js
TheNoodleMoose: Okay went ahead and added that.

src/app/analytics/analytics.factory.js
TheNoodleMoose: Yeah, good point. Any event that needs data I'll probably just make a custom analytics function for so I'll remove the data parameter.

src/app/analytics/analytics.factory.js
```diff
@@ -187,18 +187,16 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
         // Send GTM Advance Ecommerce event
         if (typeof $window.dataLayer !== 'undefined') {
           $window.dataLayer.push({
-            event: callType === 'mini-cart' ? 'view-mini-cart' : 'view-cart'
+            event: isMiniCart ? 'view-mini-cart' : 'view-cart'
           })
         }
         // Call DTM direct call rule
-        if (typeof callType !== 'undefined' && callType === 'customLink') {
+        if (isMiniCart) {
```
wrandall22, line 199: This block doesn't really make sense.

TheNoodleMoose: Yeah, I'm not sure what the logic before was, but the event being fired is an ```aa-view-minicart``` event. So I figured we would only want that to be called if ```isMiniCart``` is true?

wrandall22: It was never called, so why not just remove the block?

TheNoodleMoose: Ah good point, yep will do!

wrandall22:
This is probably ok to push to staging, but you should add more tests before I approve the PR.

TheNoodleMoose:
@wrandall22 Here's the second batch of events. I made sure to add tests for these events while I was adding them. I'm going to make another commit to add tests for the first batch.

wrandall22:
It looks fine to push to staging, but looks like it wants more tests for prod.

TheNoodleMoose:
Okay, I'll add more tests, I think there was a big section that didn't have tests initially for AA and that's why there's such a big miss in coverage.

TheNoodleMoose:
Just pushed it to staging

wrandall22 (approved):

TheNoodleMoose:
@wrandall22 Just got the green light from Clark that everything looks good on stage! 🥳 Should I merge it in now or wait till the morning?

wrandall22:
I would wait at this point.

TheNoodleMoose:
Sounds good, I'll merge in the morning.

PR #867, opened 2020-11-20
Add event for Optimize

wrandall22:
We are trying to use Optimize to replace parts of the SPA piece of give site. According to [this doc](https://support.google.com/optimize/answer/7008840?hl=en) the way to do that is to have Optimize hook into an activation event. We already had something like this for Adobe by the looks of it, so I just tacked on an event for Google in the same function.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -414,6 +414,10 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
           }
         }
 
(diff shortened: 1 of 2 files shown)


TheNoodleMoose (approved):
So this just sends an event to GTM that triggers optimize to load something? Looks good to me.

wrandall22:
It's an event that Optimize can listen for in order to know that Angular has already been loaded. However, so far I can't get an Angular replacement to work, so this PR might be unnecessary.

PR #868, opened 2020-11-23
Fix-Transaction-GTM-Event

TheNoodleMoose:
Clark realized that transaction events aren't actually firing on give stage or prod. I think it has something to do with where the call is made, and the dataLayer being uninitialized during page loads(from review checkout to the thank you summary page). So over the weekend, I thought of a solution to prevent duplicate transaction event calls on the thank you summary page. Basically when the purchase is made I store their cart in localStorage in order for use in the transaction event. I then check localStorage for ```transactionId``` to see if it matches the purchase id coming into the function. If it doesn't I know this is a new purchase, so we need to send a transaction event and also store this purchase id in localStorage as ```transactionId```. That way if they refresh the page or something to that extent, the ```transactionId``` is set in localStorage, and the id's will now match so it won't send a duplicate transaction event. I then also remove the cart from local storage since it's no longer needed. Let me know what you think of this solution. I thought going with localStorage would be better anyway since the digitalData layer will probably go away sometime in the future.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -484,9 +484,9 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // The purchaseId number from the last purchase
         const lastTransactionId = sessionStorage.getItem('transactionId')
         // The purchaseId number from the pruchase data being passed in
-        const currentTransactionId = purchaseData.rawData['purchase-number']
+        const currentTransactionId = purchaseData && purchaseData.rawData['purchase-number']
         // If the lastTransactionId and the current one do not match, we need to send an analytics event for the transaction
-        if (lastTransactionId !== currentTransactionId) {
+        if (purchaseData && lastTransactionId !== currentTransactionId) {
           // Set the transactionId in localStorage to be the one that is passed in
           sessionStorage.setItem('transactionId', currentTransactionId)
           const cartObject = transactionCart.items.map((cartItem) => {
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -482,13 +482,13 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // Parse the cart object of the last purchase
         const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
         // The purchaseId number from the last purchase
-        const lastTransactionId = localStorage.getItem('transactionId')
+        const lastTransactionId = sessionStorage.getItem('transactionId')
         // The purchaseId number from the pruchase data being passed in
         const currentTransactionId = purchaseData.rawData['purchase-number']
         // If the lastTransactionId and the current one do not match, we need to send an analytics event for the transaction
         if (lastTransactionId !== currentTransactionId) {
           // Set the transactionId in localStorage to be the one that is passed in
-          localStorage.setItem('transactionId', currentTransactionId)
+          sessionStorage.setItem('transactionId', currentTransactionId)
           const cartObject = transactionCart.items.map((cartItem) => {
             return {
               name: cartItem.displayName.toLowerCase(),
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -464,6 +464,8 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // Build cart data layer
         this.setDonorDetails(donorDetails)
         this.buildProductVar(cartData)
+        // Stringify the cartObject and store in localStorage for the transactionEvent
+        localStorage.setItem('transactionCart', JSON.stringify(cartData))
       } catch (e) {
         // Error caught in analyticsFactory.purchase
       }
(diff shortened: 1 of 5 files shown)


src/app/checkout/step-3/step-3.component.js
```diff
@@ -161,20 +161,6 @@ class Step3Controller {
...
-    }
-    // Load purchase data
-    this.profileService.getPurchase(lastPurchaseLink)
-      .subscribe((data) => {
-      // Run ecommerce transaction event
-        this.analyticsFactory.transactionEvent(data, this.cartData)
-      },
-      (error) => {
-        this.$log.error('Error loading purchase data for transaction event', error)
-      })
```
TheNoodleMoose, line 177: This solution also gets rid of the additional calls to the api which is great.

src/app/analytics/analytics.factory.js
```diff
@@ -475,42 +477,55 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
+        const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
+        // The purchaseId number from the last purchase
+        const lastTransactionId = localStorage.getItem('transactionId')
+        // The purchaseId number from the pruchase data being passed in
+        const currentTransactionId = purchaseData.rawData['purchase-number']
+        // If the lastTransactionId and the current one do not match, we need to send an analytics event for the transaction
+        if (lastTransactionId !== currentTransactionId) {
+          // Set the transactionId in localStorage to be the one that is passed in
+          localStorage.setItem('transactionId', currentTransactionId)
+          const cartObject = transactionCart.items.map((cartItem) => {
```
wrandall22, line 492: This `const` should only be alive within this `if` block being defined like this. Shouldn't it be defined above the `if` block instead?

wrandall22: Nevermind, it is also used inside the `if`. Hard to see in GitHub.

src/app/analytics/analytics.factory.js
```diff
@@ -475,42 +477,55 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
+        // Parse the cart object of the last purchase
+        const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
+        // The purchaseId number from the last purchase
+        const lastTransactionId = localStorage.getItem('transactionId')
+        // The purchaseId number from the pruchase data being passed in
+        const currentTransactionId = purchaseData.rawData['purchase-number']
+        // If the lastTransactionId and the current one do not match, we need to send an analytics event for the transaction
+        if (lastTransactionId !== currentTransactionId) {
+          // Set the transactionId in localStorage to be the one that is passed in
+          localStorage.setItem('transactionId', currentTransactionId)
```
wrandall22, line 491: So, it looks like `transactionId` is never cleared from local storage. I wonder if this could introduce a security hole, especially on public computers. Maybe it should be cleared on sign-in and/or sign-out events?

TheNoodleMoose: I also thought about doing sessionStorage instead, do you think that would be better?

wrandall22: How would the thank you summary load in a new browser session?

TheNoodleMoose: You can disregard that last part haha. Should I go with sessionStorage instead so we make sure it's cleared after their session ends regardless of being signed in or not?

TheNoodleMoose: Actually, I think that is the way to go because we do the same thing for the ```lastPurchaseLink```, ```storedCvvs```, etc

wrandall22: So, I thought of one way that might hit the thank you page without going through checkout, which is site crawlers. You can go to https://give-stage2.cru.org/thank-you.html directly without a cart. It throws an error, but we should probably also account for that case of bots/web crawlers.

src/app/analytics/analytics.factory.js
```diff
@@ -482,13 +482,13 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
         // Parse the cart object of the last purchase
         const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
         // The purchaseId number from the last purchase
-        const lastTransactionId = localStorage.getItem('transactionId')
+        const lastTransactionId = sessionStorage.getItem('transactionId')
```
wrandall22, line 485: `sessionStorage` is not currently available to this factory I think.

TheNoodleMoose: It seems to work. Is it because ```$window``` is exposed in the factory? The other places we use sessionStorage just do ```this.sessionStorage = $window.sessionStorage```
Screen Shot 2020-11-23 at 10 01 01 AM

wrandall22: Interesting. Not sure why.

src/app/analytics/analytics.factory.js
```diff
@@ -475,42 +477,55 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
-                products: [
-                  ...cartObject
-                ]
-              }
+        // Parse the cart object of the last purchase
+        const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
+        // The purchaseId number from the last purchase
+        const lastTransactionId = sessionStorage.getItem('transactionId')
+        // The purchaseId number from the pruchase data being passed in
+        const currentTransactionId = purchaseData.rawData['purchase-number']
```
wrandall22, line 487: Perhaps in order to account for web crawlers, we should ensure `purchaseData` is defined?

TheNoodleMoose: Yep, sounds good to me!

wrandall22:
Feel free to push to staging/test in stage.

TheNoodleMoose:
@wrandall22 Clark gave the thumbs up that this works as expected now on staging. Should I wait for a code review from Scotty or do you think this is good as is for prod?

wrandall22:
I'd like to hear @OzzieOrca opinion on the above conversation.

src/app/analytics/analytics.factory.js
OzzieOrca: `localStorage` and `sessionStorage` are both JS globals because everything on the `window` object is considered to be global. Best practices in AngularJS seem to suggest you get these from `$window`. I think that's to make mocking in tests a little easier. But idk if it matters too much.

wrandall22 (approved):

PR #870, opened 2020-11-24
Update-Checkout-Step-Event

TheNoodleMoose:
We thought we could pass this as a string with the value but it looks like it needs to be a number.


wrandall22:
This looks ok. Are we sure it is just those 3? `cart` and `thankYou` don't get steps?

TheNoodleMoose:
Brenton said just those 3 steps.

wrandall22 (approved):

TheNoodleMoose:
Just pushed it to stage as well

PR #872, opened 2020-11-25
Update-Transaction-Event-Totals

TheNoodleMoose:
We noticed in Google Analytics that some of the transaction totals were ```$0.00```. Figured out that this is caused by doing monthly, quarterly, or annual donations for a future date. Brenton requested that we just add all the totals of each donation and send that instead.


wrandall22 (approved):
Recurring donations do not factor into `monetary-total` as you surmised.

TheNoodleMoose:
Pushed to stage as well.

wrandall22:
My preferred workflow would be to push it to staging and test there before merging.

TheNoodleMoose:
Okay, yeah that makes sense and will do for future prs. Brenton was pretty confident it would work and wanted it in prod asap, but testing on staging first makes sense.

PR #875, opened 2020-11-30
fix-transaction-credit-card-events

TheNoodleMoose:


src/app/checkout/step-2/step-2.component.js
```diff
@@ -50,7 +50,7 @@ class Step2Controller {
   }
 
   handlePaymentChange (selectedPaymentMethod) {
-    this.defaultPaymentType = selectedPaymentMethod['account-type']
+    this.defaultPaymentType = selectedPaymentMethod['account-type'] ? selectedPaymentMethod['account-type'] : selectedPaymentMethod['card-type']
```
OzzieOrca, line 53: Is `this.defaultPaymentType` only used for analytics? Maybe naming it `this.analyticsPaymentType` or something would be better.

OzzieOrca (approved):
Feel free to merge as is.

src/app/checkout/step-2/step-2.component.js
TheNoodleMoose: I think it's used in other components (```payment-method-form``` and ```checkout-existing-payment-methods``` maybe?). ```defaultPaymentType``` was already defined, I just added the function for analytics purposes.

OzzieOrca: It looks like you changed the value of the default though in your last PR... Which seems a little weird... Is there a way to access the current value?

TheNoodleMoose: Yeah looking at my change in that pr, I don't think ```defaultPaymentType``` even had a value which is why I had to add this function to update it here. I think when this component is initialized the current value is just ```undefined```. I can change it to ```this.analyticsPaymentType``` here since it's only used for that purpose if you think that would be better?

OzzieOrca: `default-payment-type` can be set when using branded checkout https://github.com/CruGlobal/give-web#branded-checkout-config and it is probably passed through to here. It may be undefined unless set which is a little weird.

I think it would be better to separate concerns and not abuse the `defaultPaymentType` variable by making it hold analytics state also.

PR #877, opened 2020-12-10
Change-Product-Names-To-Use-Designation-Number

TheNoodleMoose:
I know we still have a code feeze, but another pr that I'm making just to have the work done for when the freeze is over. Part of the team was a little uneasy with being able to see the names of people and how much they were getting donated. So they requested we switch to using the designation number instead.


wrandall22 (approved):

TheNoodleMoose:
Just pushed this to staging.

TheNoodleMoose:
@wrandall22 This looks good on staging. Am I good to merge?

PR #879, opened 2020-12-31
Fix cover fees edit gift

wrandall22:
This PR attempts to fix issues like [this one](https://rollbar.com/Cru/give-web/items/58574/occurrences/147101637267/). I can't reproduce it in a site. I don't know if my test is actually mimicking the issue effectively.

### DIFF
--- a/src/app/checkout/checkout.component.js
+++ b/src/app/checkout/checkout.component.js
@@ -112,6 +112,7 @@ class CheckoutController {
         })
         .subscribe((data) => {
           this.orderService.addFeesToNewGiftIfNecessary(data)
+          this.orderService.storeCartData(data)
           this.cartData = data
           this.analyticsFactory.buildProductVar(data)
         },
--- a/src/app/checkout/checkout.component.js
+++ b/src/app/checkout/checkout.component.js
@@ -112,7 +112,6 @@ class CheckoutController {
         })
         .subscribe((data) => {
           this.orderService.addFeesToNewGiftIfNecessary(data)
-          this.orderService.storeCartData(data)
           this.cartData = data
           this.analyticsFactory.buildProductVar(data)
         },
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -15,6 +15,7 @@ import 'rxjs/add/operator/mergeAll'
 import 'rxjs/add/operator/mergeMap'
 import 'rxjs/add/operator/catch'
 import 'rxjs/add/observable/from'
+import 'rxjs/add/observable/concat'
 
 import cortexApiService from '../cortexApi.service'
 import commonService from './common.service'
@@ -162,8 +163,10 @@ class Cart {
   }
 
   editItem (oldUri, uri, data) {
-    return this.deleteItem(oldUri)
-      .switchMap(() => this.addItem(uri, data, true))
+    return Observable.concat(
+      this.deleteItem(oldUri),
+      this.addItem(uri, data, true)
+    )
   }
 
   deleteItem (uri) {
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,7 +9,8 @@ module.exports = {
     'angular',
     'angular-mocks',
     'jest-date-mock',
-    '<rootDir>/jest/setup.js'
+    '<rootDir>/jest/setup.js',
+    'jest-extended'
   ],
   moduleNameMapper: {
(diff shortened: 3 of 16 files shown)


src/common/services/api/cart.service.js
```diff
@@ -162,8 +163,10 @@ class Cart {
   }
 
   editItem (oldUri, uri, data) {
-    return this.deleteItem(oldUri)
-      .switchMap(() => this.addItem(uri, data, true))
+    return Observable.concat(
+      this.deleteItem(oldUri),
+      this.addItem(uri, data, true)
```
TheNoodleMoose, line 168: The ```concat``` just prevents ```this.addItem``` from starting until ```this.deleteItem``` finishes right?

TheNoodleMoose (approved):
@wrandall22 I don't seem to have access to the rollbar error, but from what you explained about how ```additem``` was firing before the ```deleteItem``` function finished, this solution makes sense.

wrandall22:
What is your username in rollbar?

TheNoodleMoose:
> What is your username in rollbar?

TheNoodleMoose

wrandall22:
You should have access now.

TheNoodleMoose:
@wrandall22 I still don't seem to have access to it. I have access to https://rollbar.com/Cru/give/ but not https://rollbar.com/Cru/give-web/.

wrandall22:
Ok, should be there now.

src/common/services/api/cart.service.js
wrandall22: That's my understanding. See the [docs](https://www.learnrxjs.io/learn-rxjs/operators/combination/concat).

wrandall22:
Before the fix commit:

```
console.log src/common/services/api/order.service.spec.js:727
    Deleting item

console.log src/common/services/api/order.service.spec.js:727
    Deleting item

console.log src/common/services/api/order.service.spec.js:737
    GET purchase form

  console.log src/common/services/api/order.service.spec.js:731
    Adding item

  console.log src/common/services/api/order.service.spec.js:731
    Adding item

  console.log src/common/services/api/order.service.spec.js:741
    POST purchase
```

After the fix commit:
```
console.log src/common/services/api/order.service.spec.js:727
    Deleting item

  console.log src/common/services/api/order.service.spec.js:731
    Adding item

  console.log src/common/services/api/order.service.spec.js:727
    Deleting item

  console.log src/common/services/api/order.service.spec.js:731
    Adding item

  console.log src/common/services/api/order.service.spec.js:737
    GET purchase form

  console.log src/common/services/api/order.service.spec.js:741
    POST purchase
```

TheNoodleMoose:
Looks good to me!

TheNoodleMoose (approved):

TheNoodleMoose (approved):

src/app/checkout/checkout.component.js
```diff
@@ -112,6 +112,7 @@ class CheckoutController {
         })
         .subscribe((data) => {
           this.orderService.addFeesToNewGiftIfNecessary(data)
+          this.orderService.storeCartData(data)
```
wrandall22, line 115: So, `addFeesToNewGiftIfNecessary` actually calls `storeCartData` so I need to remove these that I added.

TheNoodleMoose: Ah yep in the ```updatePrices``` function.

TheNoodleMoose (approved):

wrandall22:
More problems found: can't edit a gift in my cart.

TheNoodleMoose:
Is that a result of these changes or related to the sync issue with the backend that we are trying to resolve?

wrandall22:
I believe it is a result of my `concat` commit.

wrandall22:
It is deleting the item successfully, but failing on the add.

wrandall22:
A side effect of the `concat` is that briefly the cart shows up as empty when editing a gift.

TheNoodleMoose (approved):
Is the cart appearing empty too noticeable or do you think it's okay from a user perspective?

wrandall22:
It's not horrible. However, I have another guess at what might be happening here, which would be, if the last item of the cart is removed, maybe the cart itself is deleted and prevents adding another item to it. This would be an unfortunate side effect of doing a delete/add vs a put when changing things in the cart. However, I get 403 responses whenever I try to do a put on an item in the cart, so I can't just switch to that.

wrandall22:
Going to close this PR. I plan to implement the API version of cover fees this week if I can manage it.

PR #881, opened 2021-01-07
Use api for cover fees

wrandall22:
This depends on changes in `give-ep-extensions` to go live and should not be merged before that.

This PR re-applies the cover credit card fees checkbox option, and sends the flag to the server to process the amount instead of editing the gifts during submission, which was causing some donors to not be able to make it through checkout.

I still want to figure out a way to make the client code cleaner (e.g. I'm not a big fan of tracking the cart in local storage). However, that may be in a separate PR, and I don't have any good ideas yet for how to do it. We could display individual amounts with a filter, but not easily for the rate totals and sums.


TheNoodleMoose:
@wrandall22  just want to make sure I have my thoughts correct on how this new addition works and how it was working before.

1. We send the cover fees decision on order submission in order for the server to correctly calculate the amount on the backend.
2. We still do the calculation of the cover fees amount on the front end simply for display purposes. 
3. Before this, we were editing the amount of each item in the cart and sending that updated amount to the backend based on their cover fee decision and then using that amount for the transaction. While this simply just holds an unedited item amount on the backend and only worries about calculating for the transaction after they submit(determining whether to add cover fees amounts by what we now send).

wrandall22:
That is basically correct. There is a bit of nuance to number 3. The way the editing of gifts works is that it deletes the gift and adds a new one with the new values. I believe this was the cause of the issues, that sometimes the process of deleting and adding the gifts was deleting the cart entirely.

TheNoodleMoose (approved):

OzzieOrca:
I'm not seeing where the cover fee decision gets sent to the API. Maybe I missed something but I've stared a this code for a while and am still confused about how it works.

Is there a reason the API can't handle the whole thing? The client really shouldn't be having to do these fee conversions locally...

The API or client could calculate costs both with and without cover fees and the client could potentially switch between them in the view depending on the cover fee decision boolean. Although the client shouldn't be making calculations that are non-authoritative that the API will recalculate later (potentially differently if there's a bug on the client or API).

src/common/services/api/order.service.js
```diff
@@ -254,6 +254,7 @@ class Order {
     return this.getPurchaseForm()
       .mergeMap((data) => {
         const postData = cvv ? { 'security-code': cvv } : {}
+        postData['cover-cc-fees'] = !!this.retrieveCoverFeeDecision()
```
wrandall22, line 257: @OzzieOrca this is where it is being sent to the API

wrandall22:
It's a bit tricky. The cost is driven by the user input and is determined before the cover fees decision is made. However, since the calculation is fixed it can possibly be done ahead of time and passed to the client. It would then have to be calculated again (can probably use the same code to do it in order to reduce bugs) after the gift is submitted since amounts are not part of the order submission. We'll also have to do the calculations for the frequency totals separately and pass that as a new piece of information from the API.

wrandall22:
Unfortunately, this code has introduced double fee addition in branded checkout. I've been working on moving all of the calculations to the server to avoid this (as per Scotty's suggestion). I think the API work is done, but haven't been able to test it against the client yet.

wrandall22:
It looks like we can do the calculations on the API side for standard checkout, but not for branded checkout because the amount is never on the server until the gift is submitted.

OzzieOrca:
Really? I can look some more but shouldn't the review page be loading gifts and amounts from the API? The cart on the first page might not update... We could possibly make it save the gift and refresh it somehow when clicking the checkbox...

Is there a way to implement this as some sort of cart level fee like tax/VAT/shipping/etc so we don't have to change cart amounts? Seems like EP would have some sort of feature like that since it's a general ecommerce thing...

wrandall22:
The cover fees box is on page 1, so it needs to show the modified amounts on page 1. So sending it to the API between page 1 and 2 won't work out. If we send it after they alter the amount, that could really harm the performance.

wrandall22:
I think in the `branded-checkout-step-1` component we could do the calculation one time on the amount and store it similar to how the api is returning it. Might give that a try a bit.

PR #886, opened 2021-02-02
Fix issues with cover fees

wrandall22:
Found a couple of issues with the cover-fees code. The primary one being that the wrong amount shows up in the nav-cart (showing the numerical instead of display amount). Secondarily, the old code would show the cart total, which is based off of the one-time total, the new code showed either the cart total or the first rate total, which could be a recurring total. This PR puts that back to the way it was. Third, I added some tests to show some of these failures and fixes.

### DIFF
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -106,9 +106,11 @@ class Cart {
     })
 
     let cartTotal
+    let cartTotalDisplay
     const frequencyTotals = map(cartResponse.rateTotals, rateTotal => {
       if (rateTotal.recurrence.interval === 'NA') {
         cartTotal = rateTotal.cost.amount
+        cartTotalDisplay = rateTotal.cost.display
       }
       return {
         frequency: rateTotal.recurrence.display,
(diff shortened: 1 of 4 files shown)


wrandall22:
Can someone please review this today so I can get it live?

TheNoodleMoose:
@wrandall22 I started taking a look yesterday but got pulled away. I'll finish up my review now.

src/common/services/api/cart.service.js
```diff
@@ -67,64 +67,73 @@ class Cart {
...
+        frequency: rateTotal.recurrence.display,
+        amount: rateTotal.cost.amount,
+        amountWithFees: rateTotal.cost['amount-with-fees'],
+        total: rateTotal.cost.display,
+        totalWithFees: rateTotal.cost['display-with-fees']
+      }
+    })
+    if (cartTotal) {
+      frequencyTotals.unshift(frequencyTotals.pop())
+    }
```
TheNoodleMoose, line 125: Could you explain the need for this logic?

TheNoodleMoose (approved):

src/common/services/api/cart.service.js
wrandall22: Yeah, so the display is expecting a specific order, with `Single` gifts rate total being first. The API is returning `Single` gifts rate total last. Maybe there's a better way to do it.

wrandall22: I can put a comment in the code as well.

TheNoodleMoose: Ah okay, that makes sense. Yeah, I think a comment would be good.

TheNoodleMoose (approved):

wrandall22:
Thanks for reviewing!

PR #887, opened 2021-02-02
Fix cache invalidation issue

wrandall22:
The cache invalidation has been broken for awhile. This should hopefully fix it.


OzzieOrca (approved):
Thanks for doing this!

PR #890, opened 2021-02-16
additional-transaction-event-custom-dimensions

TheNoodleMoose:
Here's a list of what the custom dimension values should be:

1.  dimension1: Donor Type
2. dimension2: Donor Account Number
3. dimension3: Donation Type
4. dimension4: Donation Frequency
6. dimension6: Payment Type
7. dimension7: Siebel Order ID
8. dimension8: Siebel Product Type
9. dimension9: Siebel Campaign Code

It was decided that ```dimension5``` was no longer needed for now so it was removed after discussing with the analytics team.


src/app/analytics/analytics.factory.js
```diff
@@ -537,9 +537,15 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
-              dimension3: cartItem.frequency.toLowerCase() === 'single' ? 'one-time' : 'recurring',
+              dimension1: localStorage.getItem('gaDonorType'),
+              dimension2: localStorage.getItem('gaDonorAcct') || undefined,
+              dimension3: cartItem.designationType.toLowerCase(),
               dimension4: cartItem.frequency.toLowerCase(),
-              dimension8: cartItem.designationType.toLowerCase()
+              dimension5: cartItem.frequency.toLowerCase() === 'single' ? 'one-time' : 'recurring',
+              dimension6: purchaseData.paymentMeans['account-type'] ? purchaseData.paymentMeans['account-type'].toLowerCase() : purchaseData.paymentMeans['card-type'].toLowerCase(),
+              dimension7: purchaseData.rawData['purchase-number'],
+              dimension8: 'designation',
```
TheNoodleMoose, line 547: Any references in the codebase to ```Siebel Product Type``` was always set to ```designation``` so I did the same thing here.

src/app/analytics/analytics.factory.js
```diff
@@ -537,9 +537,15 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
               category: cartItem.designationType.toLowerCase(),
               variant: cartItem.frequency.toLowerCase(),
               quantity: '1',
-              dimension3: cartItem.frequency.toLowerCase() === 'single' ? 'one-time' : 'recurring',
+              dimension1: localStorage.getItem('gaDonorType'),
+              dimension2: localStorage.getItem('gaDonorAcct') || undefined,
+              dimension3: cartItem.designationType.toLowerCase(),
```
wrandall22, line 542: This looks like dimension 8

wrandall22: Well, maybe not...But I don't know what "Donation Type" really means.

TheNoodleMoose: Do you mean the old dimension 8? Brenton originally created the first three dimensions I added(dimension3, 4, 8) but then Lauren went through and remapped the values so things got switched around.

src/app/analytics/analytics.factory.js
wrandall22: Yeah, there are other types, but we don't currently support them in the give site. In the future we might pull in others.

src/app/analytics/analytics.factory.js
wrandall22: Donation Type does not scream Designation Type to me. But maybe that's what they meant.

src/app/analytics/analytics.factory.js
```diff
@@ -537,9 +537,15 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
...
               dimension4: cartItem.frequency.toLowerCase(),
-              dimension8: cartItem.designationType.toLowerCase()
+              dimension5: cartItem.frequency.toLowerCase() === 'single' ? 'one-time' : 'recurring',
```
wrandall22, line 544: I'm not a fan of the wording between 4 and 5. I would have swapped these if I were interpreting them. Are we sure this is the correct mapping?

TheNoodleMoose: Yeah same. I believe four is just supposed to be what the frequnce value is while five is supposed to confirm if it's a reoccurring type or not. I'll double check but I believe this is what they wanted.

TheNoodleMoose:
@wrandall22 So I talked with Lauren To, and she said that the value I had set for ```dimension5``` should actually be the value for ```dimension3```(which makes more sense). We then decided that ```dimension5``` may not be needed so we are removing that one for the time being.

wrandall22:
So, now we're no longer sending designation type? (e.g. Staff, Ministry)

TheNoodleMoose:
Technically we send it as the ```category``` for the event so they have been reporting on that.

TheNoodleMoose:
And there is always the possibility for them to add more in the future so if they decide they would rather not report on category and would rather have a distinct custom dimension for the designation type they can do that.

wrandall22:
Gotcha. Can you update the description in this PR to the new reality?

TheNoodleMoose:
Updated the description.

wrandall22:
Dimension 3 in the code does not match the description.

wrandall22:
> Dimension 3 in the code does not match the description.

Unless Donation Type means one-time vs recurring?

TheNoodleMoose:
Yeah, I confirmed with Lauren that's what it should be. I guess for donation type they want to know if it's a recurring donation or a one-time donation.

wrandall22:
Has this gone through `staging` yet?

TheNoodleMoose:
No, I'll push it there first, let Lauren QA it and then we can push to prod if it looks good to her.

TheNoodleMoose:
@wrandall22 So had more talks with Lauren. They changed dimension2 from being product scoped to session-scoped because they want that value whenever it's available in the session. So I removed it from the transaction event and instead created a custom javascript function within GTM to check for the value, similar to how we did it here.

TheNoodleMoose:
Lauren tested the changes on staging and said it looked good so we can merge this in.

wrandall22:
So GTM is using the same local storage variable?

TheNoodleMoose:
Yeah this is the function in GTM:
```
function () {
    var donorAcctNumber = localStorage.getItem('gaDonorAcct') || undefined
    return donorAcctNumber;
}
```

TheNoodleMoose:
And she confirmed that it showed up as dimension2  as expected :

Screen Shot 2021-02-17 at 4 40 07 PM

wrandall22 (approved):

PR #905, opened 2021-07-02
Remove-Remaining-Adobe-Analytics-Events

TheNoodleMoose:
I broke this up into two commits:
1.  Removing remaining adobe events that have been migrated to google analytics.
2. Removing the remaining adobe events that were not migrated.

I confirmed with Lauren and Clark, but all events of importance were migrated over during the initial migration phase. These events that were removed in the second commit are no longer of importance. However, I split this into two commits in case they become important again in the future. That way, we'll have a way to identify what remaining events need to be added to google analytics.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -130,11 +130,6 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
             }
           })
         }
-
-        // Call DTM direct call rule
-        if (typeof $window._satellite !== 'undefined') {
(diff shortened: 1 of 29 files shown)


wrandall22:
Thanks for splitting these commits up, that is a good breakpoint. I would suggest adding a 3rd commit to rename `trackGTM` to `track`.

wrandall22:
Feel free to push this to the staging branch, but we'll wait for Tuesday to merge into master.

wrandall22 (approved):

TheNoodleMoose:
Pushed this to staging and it looks good.

PR #907, opened 2021-07-13
Update Transaction Event Payment Type Values

TheNoodleMoose:
As per request, they wanted to update the payment type values from ```visa```, ```amex```, etc to just be either ```bank account``` or ```credit card```.


wrandall22 (approved):

TheNoodleMoose:
Just tested on staging, looks good
Screen Shot 2021-07-13 at 9 28 00 AM

PR #909, opened 2021-07-16
GTM-CheckoutStepEvent-Brand-Fix

TheNoodleMoose:
While investigating setting ```brand``` to have a value of an ```organization id```, I happened to notice this small typo. I guess this was missed in QA, and must not have been causing issues this whole time. Also went and double-checked the docs, https://developers.google.com/tag-manager/enhanced-ecommerce#checkoutstep, just to confirm the typo.


wrandall22 (approved):

TheNoodleMoose:
Tested on staging. ```brand``` is now there in the ecommerce object
Screen Shot 2021-07-16 at 1 41 05 PM

PR #910, opened 2021-07-16
EP-2260: Radio Station Select on Branded Checkout

reldredge71:
Adds an optional Radio Station selector to the branded checkout flow.  It could probably use some polishing, let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,9 @@ Add the following code to your page where appropriate. See the [Branded checkout
     donor-details="<window variable containing default values for donor's name and contact info>"
     show-cover-fees="true"
     on-order-completed="$event.$window.onOrderCompleted($event.purchase)"
-    on-order-failed="$event.$window.onOrderFailed($event.donorDetails)">
+    on-order-failed="$event.$window.onOrderFailed($event.donorDetails)"
+    radio-station-api-url="https://api.domain.com/getStations"
+    radio-station-radius="100">
 </branded-checkout>
 
 <script src="https://give-static.cru.org/branded-checkout.v2.js"></script>
@@ -128,6 +130,9 @@ The `<branded-checkout>` element is where the branded checkout Angular app will 
 - `on-order-completed` - an Angular expression that is executed when the order was submitted successfully - *Optional* -  provides 2 variables:
   - `$event.$window` - Provides access to the browser's global `window` object. This allows you to call a custom callback function like `onOrderCompleted` in the example.
   - `$event.purchase` - contains the order's details that are loaded for the thank you page
+- `radio-station-api-url` - Provides a URL path for fetching a list of radio stations in the donor's vicinity - *Optional*
+- `radio-station-radius` - Provides a radius (in miles) for fetching a list of radio stations in the donor's vicinity - *Optional*
+
 
 #### Server-side configuration for a new branded checkout domain
 1. Figure out what domain you will be hosting the branded checkout form on. For example, `myministry.org`
--- a/branded-checkout.html
+++ b/branded-checkout.html
@@ -11,7 +11,7 @@
 </head>
 <body>
 
-<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
+<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="https://api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
 <script src="dev.v2.js"></script>
 
 </body>
--- a/branded-checkout.html
+++ b/branded-checkout.html
@@ -11,8 +11,8 @@
 </head>
 <body>
 
-<branded-checkout designation-number="2294554" show-cover-fees="true" ng-cloak></branded-checkout>
-<script src="branded-checkout.v2.js"></script>
+<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="https://api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
+<script src="dev.v2.js"></script>
 
 </body>
 </html>
--- a/branded-checkout.html
+++ b/branded-checkout.html
@@ -11,7 +11,7 @@
 </head>
 <body>
 
-<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="https://api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
+<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
 <script src="dev.v2.js"></script>
 
 </body>
--- a/src/app/branded/step-2/branded-checkout-step-2.component.js
+++ b/src/app/branded/step-2/branded-checkout-step-2.component.js
@@ -2,6 +2,7 @@ import angular from 'angular'
 
 import checkoutStep3 from 'app/checkout/step-3/step-3.component'
 import cartService from 'common/services/api/cart.service'
+import orderService from '../../../common/services/api/order.service'
 
 import template from './branded-checkout-step-2.tpl.html'
 
@@ -9,9 +10,10 @@ const componentName = 'brandedCheckoutStep2'
 
 class BrandedCheckoutStep2Controller {
   /* @ngInject */
-  constructor ($log, cartService) {
+  constructor ($log, cartService, orderService) {
     this.$log = $log
     this.cartService = cartService
+    this.orderService = orderService
   }
 
   $onInit () {
@@ -30,6 +32,10 @@ class BrandedCheckoutStep2Controller {
       )
   }
 
+  loadRadioStation () {
+    this.radioStationName = this.orderService.retrieveRadioStationData().description
+  }
+
   changeStep (newStep) {
     if (newStep === 'thankYou') {
       this.next()
@@ -42,7 +48,8 @@ class BrandedCheckoutStep2Controller {
 export default angular
   .module(componentName, [
     checkoutStep3.name,
-    cartService.name
+    cartService.name,
+    orderService.name
   ])
   .component(componentName, {
     controller: BrandedCheckoutStep2Controller,
--- a/src/app/branded/step-2/branded-checkout-step-2.component.js
+++ b/src/app/branded/step-2/branded-checkout-step-2.component.js
@@ -14,10 +14,13 @@ class BrandedCheckoutStep2Controller {
     this.$log = $log
     this.cartService = cartService
     this.orderService = orderService
+
+    this.radioStationName = undefined
   }
 
   $onInit () {
     this.loadCart()
+    this.loadRadioStation()
   }
 
   loadCart () {
@@ -33,7 +36,7 @@ class BrandedCheckoutStep2Controller {
   }
 
   loadRadioStation () {
-    this.radioStationName = this.orderService.retrieveRadioStationData().description
+    this.radioStationName = this.orderService.retrieveRadioStationName()
   }
 
   changeStep (newStep) {
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -73,10 +73,10 @@
             <div>
               {{$ctrl.donorDetails.email}}
             </div>
-              <div class="form-group">
-                <label class="" translate>{{'RADIO_STATION'}}</label>
-                <p>{{$ctrl.radioStationName}}</p>
-              </div>
+            <div class="form-group" ng-if="$ctrl.radioStationName">
+              <label class="" translate>{{'RADIO_STATION'}}</label>
+              <p>{{$ctrl.radioStationName}}</p>
+            </div>
             <loading ng-if="!$ctrl.donorDetails"></loading>
           </div>
         </div>
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -73,6 +73,10 @@
             <div>
               {{$ctrl.donorDetails.email}}
             </div>
+              <div class="form-group">
+                <label class="" translate>{{'RADIO_STATION'}}</label>
+                <p>{{$ctrl.radioStationName}}</p>
+              </div>
             <loading ng-if="!$ctrl.donorDetails"></loading>
           </div>
         </div>
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -73,9 +73,9 @@
             <div>
               {{$ctrl.donorDetails.email}}
             </div>
-            <div class="form-group" ng-if="$ctrl.radioStationName">
-              <label class="" translate>{{'RADIO_STATION'}}</label>
-              <p>{{$ctrl.radioStationName}}</p>
+            <div ng-if="$ctrl.donorDetails && $ctrl.radioStationName">
+              <translate>{{'PREFERRED_RADIO_STATION'}}</translate>
+              {{$ctrl.radioStationName}}
             </div>
             <loading ng-if="!$ctrl.donorDetails"></loading>
           </div>
--- a/src/common/components/addressForm/addressForm.component.spec.js
+++ b/src/common/components/addressForm/addressForm.component.spec.js
@@ -19,7 +19,8 @@ describe('addressForm', function () {
         $valid: false,
         $setSubmitted: jest.fn()
       },
-      onSubmit: jest.fn()
+      onSubmit: jest.fn(),
+      onAddressChanged: jest.fn()
     })
   }))
 
@@ -112,6 +113,16 @@ describe('addressForm', function () {
 
       expect(self.controller.$log.error.logs[0]).toEqual(['Error loading regions.', 'some error'])
       expect(self.controller.loadingRegionsError).toEqual(true)
+    })
+  })
+
+  describe('onPostalCodeChanged', () => {
+    it('should trigger onAddressChanged callback', () => {
+      self.controller.countries = countriesResponse._element
+      jest.spyOn(self.controller, 'onAddressChanged')
+      self.controller.onPostalCodeChanged()
+
+      expect(self.controller.onAddressChanged).toHaveBeenCalledWith()
     })
   })
 })
--- a/src/common/components/contactInfo/contactInfo.component.js
+++ b/src/common/components/contactInfo/contactInfo.component.js
@@ -41,6 +41,7 @@ class Step1Controller {
     }
 
     this.loadDonorDetails(donorDetailsDefaults)
+    this.loadRadioStations()
     this.waitForFormInitialization()
 
     this.$scope.$on(SignInEvent, () => {
@@ -51,6 +52,9 @@ class Step1Controller {
   $onChanges (changes) {
     if (changes.submitted.currentValue === true) {
       this.submitDetails()
+    }
+    if (changes.donorDetails.address.postalCode) {
+      this.loadDonorDetails()
     }
   }
 
@@ -108,6 +112,20 @@ class Step1Controller {
       })
   }
 
+  loadRadioStations () {
+    if (this.requestRadioStation) {
+      this.loadingRadioStationsError = false
+      this.geographiesService.getRadioStations(this.donorDetails.address.postalCode) //Need new service
+        .subscribe((data) => {
+          this.radioStations = data
+        },
+        error => {
+          this.loadingRadioStationsError = true
+          this.$log.error('Error loading radio stations.', error)
+        })
+    }
+  }
+
   submitDetails () {
     this.detailsForm.$setSubmitted()
     if (this.detailsForm.$valid) {
@@ -150,6 +168,7 @@ export default angular
     bindings: {
       submitted: '<',
       donorDetails: '=?',
-      onSubmit: '&'
+      onSubmit: '&',
+      requestRadioStation: '<'
     }
   })
--- a/src/common/components/contactInfo/contactInfo.component.js
+++ b/src/common/components/contactInfo/contactInfo.component.js
@@ -22,10 +22,11 @@ const componentName = 'contactInfo'
 
 class Step1Controller {
   /* @ngInject */
-  constructor ($log, $scope, orderService, sessionService, analyticsFactory) {
+  constructor ($log, $scope, orderService, radioStationService, sessionService, analyticsFactory) {
     this.$log = $log
     this.$scope = $scope
     this.orderService = orderService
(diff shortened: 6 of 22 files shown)


wrandall22:
Make sure to run `yarn lint` locally.

src/common/components/contactInfo/contactInfo.component.spec.js
```diff
@@ -195,16 +197,93 @@ describe('contactInfo', function () {
...
+      self.controller.radioStationApiUrl = radioStationApiUrl
+      self.controller.radioStationRadius = radioStationRadius
+      self.controller.requestRadioStation = true
+      self.controller.donorDetails = { mailingAddress: { } }
+
+      jest.spyOn(self.controller.radioStationsService, 'getRadioStations').mockImplementation(() => Observable.of([]))
+      self.controller.loadRadioStations()
+
+      expect(self.controller.radioStationsService.getRadioStations).not.toHaveBeenCalled()
+      expect(self.controller.radioStations).toEqual(undefined)
```
wrandall22, line 229: You can use `toBeUndefined()` instead.

src/common/components/contactInfo/contactInfo.component.spec.js
```diff
@@ -195,16 +197,93 @@ describe('contactInfo', function () {
...
+      self.controller.radioStationApiUrl = undefined
+      self.controller.radioStationRadius = undefined
+      self.controller.requestRadioStation = false
+      self.controller.donorDetails = { mailingAddress: { postalCode } }
+
+      jest.spyOn(self.controller.radioStationsService, 'getRadioStations').mockImplementation(() => Observable.of([]))
+      self.controller.loadRadioStations()
+
+      expect(self.controller.radioStationsService.getRadioStations).not.toHaveBeenCalled()
+      expect(self.controller.radioStations).toEqual(undefined)
```
wrandall22, line 216: You can use `toBeUndefined()` instead.

src/common/components/contactInfo/contactInfo.component.spec.js
```diff
@@ -195,16 +197,93 @@ describe('contactInfo', function () {
...
+      self.controller.radioStationApiUrl = radioStationApiUrl
+      self.controller.radioStationRadius = radioStationRadius
+      self.controller.requestRadioStation = true
+      self.controller.donorDetails = { mailingAddress: { postalCode } }
+
+      jest.spyOn(self.controller.radioStationsService, 'getRadioStations').mockImplementation(() => Observable.throw('some error'))
+      self.controller.loadRadioStations()
+
+      expect(self.controller.radioStationsService.getRadioStations).toHaveBeenCalledWith(radioStationApiUrl, postalCode, radioStationRadius)
+      expect(self.controller.radioStations).toEqual(undefined)
```
wrandall22, line 255: You can use `toBeUndefined()` instead.

src/common/components/contactInfo/contactInfo.component.spec.js
```diff
@@ -228,8 +310,8 @@ describe('contactInfo', function () {
         emailFormUri: '/emails/crugive',
         'donor-type': 'Staff'
       })
-
       expect(self.controller.orderService.addEmail).toHaveBeenCalledWith('someone@asdf.com', '/emails/crugive')
+      expect(self.controller.orderService.storeRadioStationData).toHaveBeenCalledWith(radioStationData)
```
wrandall22, line 314: It seems like this should be split out into a separate test `it('should save radio station')`

src/common/services/api/order.service.spec.js
```diff
@@ -711,7 +711,23 @@ describe('order service', () => {
...
+
+      self.orderService.submit()
+        .subscribe((data) => {
+          expect(data).toEqual(purchaseResponse)
+        })
+
+      self.$httpBackend.flush()
+    })
+
+    it('should send the radio cover letters to the server', () => {
```
wrandall22, line 725: One thing I've noticed with these tests that follow the pattern:
```
someService.function()
  .subscribe((data) => {
    expect(stuff)
})
```
Is that we need to use the [done](https://jestjs.io/docs/asynchronous) callback in order for it to actually be testing something.

src/common/services/api/radioStations.service.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    self.radioStationService = radioStationService
+    self.$httpBackend = $httpBackend
+  }))
+
+  afterEach(() => {
+    self.$httpBackend.verifyNoOutstandingExpectation()
+    self.$httpBackend.verifyNoOutstandingRequest()
+  })
+
+  it('getRadioStations', () => {
```
wrandall22: This doesn't make sense. It should be `describe('getRadioStations')` with some `it(...)` calls underneath that make sense.

src/common/services/api/radioStations.service.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    const expectedRadioStations = [{ Description: 'Radio Station', MediaId: 'WXYZ' }, { Description: 'Other Station', MediaId: 'ZYXW' }]
+
+    self.$httpBackend.expectGET('https://api.domain.com/getStations/333333/100')
+      .respond(200, expectedRadioStations)
+    self.radioStationsService.getRadioStations('https://api.domain.com/getStations', '33333', '100')
+      .subscribe((data) => {
+        expect(data).toEqual(expectedRadioStations)
```
wrandall22: Use the `done` callback here.

src/common/services/api/radioStations.service.spec.js
```diff
@@ -1,30 +1,33 @@
...
-
-    self.$httpBackend.expectGET('https://api.domain.com/getStations/333333/100')
-      .respond(200, expectedRadioStations)
-    self.radioStationsService.getRadioStations('https://api.domain.com/getStations', '33333', '100')
-      .subscribe((data) => {
-        expect(data).toEqual(expectedRadioStations)
-      })
-    self.$httpBackend.flush()
+  describe('getRadioStations', () => {
+    it('returns list of radio stations', () => {
```
wrandall22, line 21:
```suggestion
    it('returns a list of radio stations', () => {
```

src/common/services/api/radioStations.service.js
```diff
@@ -5,7 +5,8 @@ import 'rxjs/add/operator/mergeMap'
 import 'rxjs/add/operator/pluck'
 import 'rxjs/add/observable/of'
 import 'rxjs/add/observable/throw'
-import appConfig from 'common/app.config'
+import 'rxjs/add/operator/do'
```
wrandall22, line 8: This commit (b20e29ea049039ed29d5e9ae4573fc1fd90b3638) should be squashed.

src/common/services/api/radioStations.service.spec.js
```diff
@@ -7,9 +7,9 @@ describe('radio station service', () => {
   beforeEach(angular.mock.module(module.name))
   var self = {}
 
-  beforeEach(inject((_radioStationService_, _$httpBackend_) => {
-    self.radioStationService = _radioStationService_
-    self.$httpBackend = _$httpBackend_
+  beforeEach(inject((radioStationsService, $httpBackend) => {
```
wrandall22, line 10: This commit (a7d920254dfbc86abed3a811c96048ba5c86f659) should be squashed.

src/common/components/contactInfo/contactInfo.component.js
```diff
@@ -117,7 +117,7 @@ class Step1Controller {
   }
 
   loadRadioStations () {
-    const postalCode = this.donorDetails.mailingAddress?.postalCode;
+    const postalCode = this.donorDetails.mailingAddress.postalCode
```
wrandall22, line 120: This commit (9fc87f7685e69f8296553ee954c2e4aa46af79c7) should be broken up and squashed with the commits that introduced the lint errors.

wrandall22 (changes requested):
Most of my comments have to do with keeping a cleaner git history. But there are a couple of functional changes as well.

wrandall22:
Also make sure to rebase `master` and force push to get the latest master changes.

src/common/services/api/radioStations.service.js
```diff
@@ -0,0 +1,38 @@
...
+    this.sessionStorage = $window.sessionStorage
+    this.localStorage = $window.localStorage
+    this.$log = $log
+    this.$filter = $filter
+  }
+
+  getRadioStations(radioStationApiUrl, postalCode, radioStationRadius) {
+    return Observable.from(this.$http({
+      method: 'GET',
+      url: radioStationApiUrl + '/' + postalCode + '/' + radioStationRadius,
```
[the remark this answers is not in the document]
reldredge71: Yeah I think it's possible for that to happen.  I wonder what a more robust solution would be.

src/common/services/api/radioStations.service.js
```diff
@@ -6,6 +6,7 @@ import 'rxjs/add/operator/combineLatest'
 import 'rxjs/add/operator/pluck'
 import 'rxjs/add/observable/of'
 import 'rxjs/add/observable/throw'
+import appConfig from 'common/app.config'
```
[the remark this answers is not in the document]
reldredge71: I was trying to look at the other services to figure out how to access `$http`.  At one point I thought that I needed appConfig to do that.  I can remove it now though.

src/common/components/contactInfo/contactInfo.component.js
```diff
@@ -44,7 +44,8 @@ class Step1Controller {
 
     this.requestRadioStation = !!(this.radioStationApiUrl && this.radioStationRadius)
 
-    this.radioStation = undefined
+    this.radioStationName = undefined
+    this.radioStationCallLetters = undefined
```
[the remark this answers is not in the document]
reldredge71: I was thinking it would be better to explicitly declare an initial value of `undefined` for these values... but it doesn't seem like this is needed.  They will resolve to undefined anyway, until they're given values from the other methods.

src/common/components/addressForm/addressForm.component.spec.js
```diff
@@ -114,4 +115,14 @@ describe('addressForm', function () {
...
     })
   })
+
+  describe('onPostalCodeChanged', () => {
+    it('should trigger onAddressChanged callback', () => {
+      self.controller.countries = countriesResponse._element
+      jest.spyOn(self.controller, 'onAddressChanged')
+      self.controller.onPostalCodeChanged()
+
+      expect(self.controller.onAddressChanged).toHaveBeenCalledWith()
```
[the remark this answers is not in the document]
reldredge71: I developed a philosophy from working in other JavaScript projects that I want to make the strongest assertion possible.  Here I want to assert two things: 1) that it was called, and 2) the inputs for the method call are correct.  `toHaveBeenCalledWith()` does both, while `toHaveBeenCalled()` does only one.

In contrast, if I want to assert that something was not called, I exclusively use `.not.toHaveBeenCalled()` over `.not.toHaveBeenCalledWith()`.  The latter is not strong enough of an assertion, because the method might in fact be called with different inputs and we would not know.  `.not.toHaveBeenCalled()` is the stronger assertion in this case.

I eventually just developed a practice of using `.toHaveBeenCalledWith()` and `.not.toHaveBeenCalled()`.  That being said, if you think it's better that I use `toHaveBeenCalled()` here, I'm fine making that change.

src/common/services/api/order.service.spec.js
reldredge71: Would you want me to go through the other tests in this file and add `done()` to them?

src/common/services/api/order.service.spec.js
```diff
@@ -733,6 +733,7 @@ describe('order service', () => {
       self.orderService.submit()
         .subscribe((data) => {
           expect(data).toEqual(purchaseResponse)
+          done()
```
wrandall22, line 736: It's not enough to add `done()` here, you also have to send it as a parameter in the `it`

src/common/services/api/radioStations.service.js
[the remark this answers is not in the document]
wrandall22: One option is to have a whitelist of domains or something.

src/common/services/api/radioStations.service.js
```diff
@@ -19,14 +19,14 @@ class RadioStations {
     this.$filter = $filter
   }
 
-  getRadioStations(radioStationApiUrl, postalCode, radioStationRadius) {
+  getRadioStations (radioStationApiUrl, postalCode, radioStationRadius) {
```
wrandall22, line 14: This commit (dbc64fc294bf37dfd8d782be6c8ab040f84c5fa4) should be squashed with 92d24a86535a5695596ec1c476884b4e0006a276

src/common/components/contactInfo/contactInfo.component.js
```diff
@@ -41,6 +41,7 @@ class Step1Controller {
     }
 
     this.loadDonorDetails(donorDetailsDefaults)
+    this.loadRadioStations()
```
wrandall22, line 51: This commit (13592e625187459d15f769d4e5e85cc2d5753449) should be squashed with 1282eedaacdd82fa9f941146ded196bc91055d24 and given a nice name about contact info being done

src/common/services/api/radioStations.service.js
```diff
@@ -5,7 +5,8 @@ import 'rxjs/add/operator/mergeMap'
 import 'rxjs/add/operator/pluck'
 import 'rxjs/add/observable/of'
 import 'rxjs/add/observable/throw'
-import appConfig from 'common/app.config'
```
wrandall22: This removal should be squashed with 2e34bf2ebe49cf9c7e5545066e8edf07ca785808 or perhaps 2e34bf2ebe49cf9c7e5545066e8edf07ca785808 should be removed altogether.

src/common/services/api/radioStations.service.js
```diff
@@ -1,12 +1,6 @@
 import angular from 'angular'
 import { Observable } from 'rxjs/Observable'
 import 'rxjs/add/operator/map'
-import 'rxjs/add/operator/mergeMap'
```
wrandall22: Commit ca986c72c6132e886c3d6b7e794feb125d083194 should be squashed

branded-checkout.html
```diff
@@ -11,8 +11,8 @@
 </head>
 <body>
 
-<branded-checkout designation-number="2294554" show-cover-fees="true" radio-station-api-url="https://api.familylife.com/FamilyLifeAPI.svc/media/getStations" radio-station-radius="100" ng-cloak></branded-checkout>
-<script src="dev.v2.js"></script>
+<branded-checkout designation-number="2294554" show-cover-fees="true" ng-cloak></branded-checkout>
+<script src="branded-checkout.v2.js"></script>
```
wrandall22: Commit 876257a02f528d9c688c71719864fe14d7685599 should be squashed

src/common/services/api/radioStations.service.spec.js
```diff
@@ -18,7 +18,7 @@ describe('radio station service', () => {
   })
 
   describe('getRadioStations', () => {
-    it('returns list of radio stations', () => {
+    it('returns a list of radio stations', () => {
```
wrandall22: Commit 3b3d60e1b1656e8a993c83bd922dbadbc65afaa5 should be squashed

src/common/services/api/order.service.spec.js
```diff
@@ -722,7 +722,7 @@ describe('order service', () => {
       self.$httpBackend.flush()
     })
 
-    it('should send the radio cover letters to the server', () => {
+    it('should send the radio cover letters to the server', (done) => {
```
wrandall22, line 725: Commit 86e1dc52d09362227fb5bcc916c46aa844b16949 should be squashed

wrandall22 (changes requested):
Also 103a71596da46680d1bc4ebd3a4fd1437b49f26b should go away

src/common/services/api/order.service.spec.js
wrandall22: You can if you want.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -220,5 +220,24 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'CONTACT_INFO'
...
       </div>
     </div>
+
+    <div class="mb" ng-if="$ctrl.requestRadioStation">
+      <h4 class="panel-title  border-bottom-small  visible" translate>{{'RADIO_STATION'}}</h4>
+      <div class="form-group">
+        <select class="form-control form-control-subtle"
+          name="radioStation"
+          ng-model="$ctrl.radioStation"
+          ng-options="v.Description as v['Description'] for v in $ctrl.radioStations | orderBy:'\'Description\''"
```
wrandall22, line 237: `orderBy:'\'Description\''` seems a bit weird to me. Do we need the quotes in the text for some reason?

src/common/components/contactInfo/contactInfo.component.js
```diff
@@ -139,6 +137,7 @@ class Step1Controller {
   }
 
   submitDetails () {
+    debugger
```
wrandall22: This should be removed

src/common/components/contactInfo/contactInfo.component.js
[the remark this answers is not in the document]
wrandall22: I think they should be removed.

src/app/branded/step-2/branded-checkout-step-2.component.js
```diff
@@ -2,20 +2,25 @@ import angular from 'angular'
...
 
 class BrandedCheckoutStep2Controller {
   /* @ngInject */
-  constructor ($log, cartService) {
+  constructor ($log, cartService, orderService) {
     this.$log = $log
     this.cartService = cartService
+    this.orderService = orderService
+
+    this.radioStationName = undefined
```
wrandall22, line 18: This line should be removed.

src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -73,9 +73,9 @@
             <div>
               {{$ctrl.donorDetails.email}}
             </div>
-            <div class="form-group" ng-if="$ctrl.radioStationName">
```
wrandall22: Commit 3d93b0b77a4aa9650a0e3c872463b370ff5e1ed3 should be split up and squashed with the commits it is fixing.

wrandall22 (changes requested):
Git history is looking a lot better! A few comments.

src/common/components/contactInfo/contactInfo.tpl.html
reldredge71: Turns out they're not needed!  I'll clean that up.

src/common/services/api/radioStations.service.js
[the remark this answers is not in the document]
reldredge71: A whitelist could work... will we have to maintain that as more organizations use branded checkout?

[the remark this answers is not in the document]
wrandall22: The maintenance would be specific to ministries who want to use radio stations, which should be very few.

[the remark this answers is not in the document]
reldredge71: Are you thinking of implementing a whitelist on the API side, or should I whitelist things on the client side before a request is made?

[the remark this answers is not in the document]
wrandall22: Hmm, to actually be secure it probably has to be on the server. Otherwise, I, as the malicious attacker could use local overrides in my Chrome dev tools to modify the whitelist to include my malicious api.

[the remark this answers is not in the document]
reldredge71: Yeah I was thinking that.  That makes sense.  The whitelist should be in a secure location that only we would modify.

README.md
```diff
@@ -128,6 +130,9 @@ The `<branded-checkout>` element is where the branded checkout Angular app will
 - `on-order-completed` - an Angular expression that is executed when the order was submitted successfully - *Optional* -  provides 2 variables:
   - `$event.$window` - Provides access to the browser's global `window` object. This allows you to call a custom callback function like `onOrderCompleted` in the example.
   - `$event.purchase` - contains the order's details that are loaded for the thank you page
+- `radio-station-api-url` - Provides a URL path for fetching a list of radio stations in the donor's vicinity - *Optional*
```
wrandall22, line 133: If we're adding a whitelist server-side, we will need to add instructions for those using these properties to contact us to add their API url to the whitelist.

reldredge71: Should I mention a specific person/email to contact?

wrandall22: Probably `help@cru.org` like the other place says.

src/common/services/api/radioStations.service.js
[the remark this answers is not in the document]
wrandall22: I wonder; since the attacker could use dev tools to just remove the code that looks at the allowed list, are we really gaining any protection?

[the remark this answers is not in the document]
wrandall22: What we do for TSYS device ID is that we have the ministry send in a keyword that is mapped to a real value on the server. Maybe we should do this instead of having them send in the API url itself?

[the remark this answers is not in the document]
wrandall22: So, in this case, the server could build `radioStationApiUrl` that is shown here if you hit an endpoint and pass in an identifier (probably the same one that is passed for the tsys identifier, which would reduce what the ministries would need to configure and keep ids the same in the system)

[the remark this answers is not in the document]
reldredge71: Hmm this is teaching me a lot about cybersecurity!  Yeah what you suggested at first makes it slightly more complicated to hack, but it's still not the most secure solution.  I see the benefit of having the URLs stored in the server and accessed through a keyword.  If we returned the URL from the server and used that, could it still get overridden?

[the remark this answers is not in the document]
wrandall22: That's a good point. Any code on the front-end could be rewritten, including all of our calls to our own API. So maybe it's not worth the effort to use an allowed list at all...It's not too difficult to set up I suppose and it provides a little extra security. @mattdrees do you have an opinion on this discussion?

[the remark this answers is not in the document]
mattdrees: I think I need a little more context.

> . Let's say a malicious agent builds out a branded checkout entering in a malicious URL here. Could they somehow pass data back to our server this way?

It seems to me that the malicious agent can enter zip codes, and get back radio stations. Which doesn't seem like a huge problem to me. Am I missing some bigger things that you're concerned about?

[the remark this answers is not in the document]
wrandall22: The thought would be they could enter zip codes and get back anything, since they control the API (e.g. scripts for XSS, etc). However, as I think about it, an attacker can change any of the javascript code locally such that any api call can hit any api and return any data to be sent to our api, so maybe it's a moot point.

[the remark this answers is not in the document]
mattdrees: So yeah, essentially this is a public api. There's nothing you can do to stop people from scraping your zip code-> radio database, if they want to. They could hire out people to collect it manually, and bypass any kind of captcha, even.

So I'd say don't bother with a whitelist.

The things you need to think about are whether a bad guy can do harm to a good user; somehow steal the info they type in or something. And if you have an XSS vuln, it's really hard to prevent that. Maybe a full CSP could. But a javascript-based solution wouldn't, because XSS by definition allows the attacker to muck up your javascript.

PR #911, opened 2021-07-21
GTM-Update-Brand-To-OrgId

TheNoodleMoose:
After discussion with Cat and Lauren, they decided to go with using the orgId for the brand value. The following events were updated:
1. add-to-cart
2. remove-from-cart
3. checkout-step
4. give-gift-modal
5. product-detail-click
6. transaction

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -38,7 +38,8 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
             item = {
               productInfo: {
                 productID: cartData.items[i].designationNumber,
-                designationType: cartData.items[i].designationType
+                designationType: cartData.items[i].designationType,
+                orgId: cartData.items[i].orgId ? cartData.items[i].orgId : 'cru'
               },
               price: {
                 basePrice: cartData.items[i].amount
@@ -121,7 +122,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
                   name: productData.designationNumber,
                   id: productData.designationNumber,
                   price: itemConfig.amount.toString(),
-                  brand: 'cru',
+                  brand: productData.orgId,
                   category: productData.designationType.toLowerCase(),
                   variant: frequencyObj.display.toLowerCase(),
                   quantity: '1'
@@ -165,7 +166,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
                     name: item.designationNumber,
                     id: item.designationNumber,
                     price: item.amount.toString(),
-                    brand: 'cru',
+                    brand: item.orgId,
                     category: item.designationType.toLowerCase(),
                     variant: item.frequency.toLowerCase(),
                     quantity: '1'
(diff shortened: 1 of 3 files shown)


src/app/analytics/analytics.factory.js
```diff
@@ -38,7 +38,8 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
             item = {
               productInfo: {
                 productID: cartData.items[i].designationNumber,
-                designationType: cartData.items[i].designationType
+                designationType: cartData.items[i].designationType,
+                orgId: cartData.items[i].orgId ?? 'cru'
```
TheNoodleMoose, line 42: Not sure if the ```?? cru``` is actually needed. From what I observed, every product had an orgId.

src/common/services/api/designations.service.js
```diff
@@ -109,7 +109,8 @@ class DesignationsService {
           type: hit.designation_type || null,
           facet: hit.facet || null,
           startMonth: hit.start_date ? moment(hit.start_date).format('YYYY-MM-01') : null,
-          ministry: hit.organization_id ? ministryIds[hit.organization_id] : null
+          ministry: hit.organization_id ? ministryIds[hit.organization_id] : null,
+          orgId: hit.organization_id ?? null
```
TheNoodleMoose, line 113: Same thing here as well.

TheNoodleMoose: I went ahead and updated this to "cru" in my most recent commit.

src/common/services/api/cart.service.js
```diff
@@ -89,6 +89,7 @@ class Cart {
       return {
         uri: item.self.uri,
         code: item.itemCode.code,
+        orgId: item.itemDefinition.details[1].value,
```
wrandall22, line 92: This seems a little fragile. Maybe an `orgId` variable should be set similar to how `designationType` is being set?

src/common/services/api/designations.service.js
```diff
@@ -164,7 +165,8 @@ class DesignationsService {
         displayName: data.definition['display-name'],
         designationType: designationType,
         code: data.code.code,
-        designationNumber: data.code['product-code']
+        designationNumber: data.code['product-code'],
+        orgId: data.definition.details[1].value
```
wrandall22, line 169: See other comment about fragility

src/app/analytics/analytics.factory.js
```diff
@@ -39,7 +39,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
               productInfo: {
                 productID: cartData.items[i].designationNumber,
                 designationType: cartData.items[i].designationType,
-                orgId: cartData.items[i].orgId ?? 'cru'
+                orgId: cartData.items[i].orgId ? cartData.items[i].orgId : 'cru'
```
wrandall22, line 42: This commit should be squashed into the other.

wrandall22 (changes requested):
I assume you've tested to make sure every spot you added an orgId actually has it available?

src/common/services/api/cart.service.js
TheNoodleMoose: Ah yeah, I like that approach better. I'll go ahead and update it.

TheNoodleMoose:
@wrandall22 Yeah I tested the events with staff, ministries, and campaigns. Each seemed to have a respective orgId. Lauren is going to do some thorough QA once this goes on staging.

wrandall22 (approved):

TheNoodleMoose:
Just pushed to staging. I will update after Lauren is finished with QA.

PR #914, opened 2021-07-23
Payment method updates

wrandall22:
This PR is the first round of fixing payment methods to work with the new 8.1 API. Some notable changes:

- The `type` of payment methods has changed, and can no longer be relied upon for logic between credit cards and bank accounts
- The JSON structure changed, moving the properties we care about into a new `payment-instrument-identification-form` object (besides billing address, which is still outside)
- The location of payment methods has changed

This does not
 - Handle the recurring gifts link on payment methods
 - Handle billing address properly

This PR also taught me there could probably be a good chunk of refactoring done to re-use code between `order.service.js` and `profile.service.js`.


TheNoodleMoose:
@wrandall22 Do you think this would have any effect on the GTM transaction event? https://github.com/CruGlobal/give-web/blob/a64b3b6f9da6c3d03b5e237b053a5e0ef0a5c235/src/app/analytics/analytics.factory.js#L546

wrandall22:
Quite possible. I haven't gotten that far yet, so I am not sure what the json looks like on the purchase for payment methods.

TheNoodleMoose:
Okay, once it's on staging I can QA that event and see if there are any necessary changes.

wrandall22:
cc @reldredge71

PR #918, opened 2021-07-30
EP-2276 - Fix address handling

wrandall22:
With the upgrade, mailing address now has a nested `address` object in it, so we need to handle this on the front-end.

In addition, I needed a convenient way to logout when doing local development, so I went ahead and added that.


wrandall22:
cc @reldredge71

PR #919, opened 2021-08-10
EP-2282 - Fix submit gift

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2282)

This PR fixes the checkout such that one can make it to the Thank You page. There are still bugs in checkout to be fixed, that will be done in separate PRs. I also cleaned up some test stuff that the earlier contractors had left as technical debt.


wrandall22:
cc @reldredge71

PR #921, opened 2021-08-11
EP-2264 - Handle payment methods on purchase

wrandall22:
This PR fixes payment method data display on the Thank You page. At some point maybe it would be worth restructuring the objects in use to match the new structure of the json, but for now it is not worth the time.


wrandall22:
cc @reldredge71

PR #922, opened 2021-08-11
EP-2285 - Fix cover fees display

wrandall22:
Cost is an array here, so we need to take that into account when display amounts with fees.


wrandall22:
cc @reldredge71

PR #923, opened 2021-08-11
EP-2261: BCO Select Premium

reldredge71:
Create option for including a premium with branded checkout.

### DIFF
--- a/src/app/branded/step-1/branded-checkout-step-1.spec.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.spec.js
@@ -271,6 +271,29 @@ describe('branded checkout step 1', () => {
     })
   })
 
+  describe('onSelectPremiumOption', () => {
+    beforeEach(() => {
+      $ctrl.initItemConfig()
+      $ctrl.premiumCode = '112233'
+    })
+
+    it('premium selected', () => {
+      $ctrl.premiumSelected = true
+
+      $ctrl.onSelectPremiumOption()
+
+      expect($ctrl.itemConfig['premium-code']).toEqual($ctrl.premiumCode)
+    })
+
+    it('premium deselected', () => {
+      $ctrl.premiumSelected = false
+
+      $ctrl.onSelectPremiumOption()
+
+      expect($ctrl.itemConfig['premium-code']).toEqual(undefined)
+    })
+  })
+
   describe('checkSuccessfulSubmission', () => {
     beforeEach(() => {
       $ctrl.resetSubmission()
--- a/src/app/branded/step-2/branded-checkout-step-2.component.js
+++ b/src/app/branded/step-2/branded-checkout-step-2.component.js
@@ -57,6 +57,8 @@ export default angular
     templateUrl: template,
     bindings: {
       previous: '&',
-      next: '&'
+      next: '&',
+      premiumName: '<',
+      itemConfig: '<'
     }
   })
--- a/src/app/checkout/step-3/step-3.component.js
+++ b/src/app/checkout/step-3/step-3.component.js
@@ -190,6 +190,8 @@ export default angular
       onSubmitted: '&',
       onSubmittingOrder: '&',
       submittingOrder: '<',
-      radioStationName: '<'
+      radioStationName: '<',
+      premiumName: '<',
+      brandedCheckoutItem: '<'
     }
   })
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -150,6 +150,22 @@
     </div>
   </div>
 
+  <div class="mb" ng-if="!!$ctrl.brandedCheckoutItem['premium-code']">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <translate>{{'PREMIUM'}}</translate>
+        <button id="changePremiumButton" class="btn btn-default btn-panel-head pull-right" ng-click="$ctrl.changeStep({newStep: 'premium'})" translate>{{'CHANGE'}}</button>
+      </div>
+      <div class="panel-body">
+        <div class="row">
+          <div class="col-md-12">
+            {{$ctrl.premiumName}}
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+
   <div class="mb">
     <div class="panel panel-default">
       <div class="panel-heading">
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ The `<branded-checkout>` element is where the branded checkout Angular app will 
 - `designation-number` - the designation number you would like donors to give to - **Required**
 - `campaign-page` - the campaign page you would like to use, used for suggested amounts - *Optional*
 - `campaign-code` - the campaign code you would like to use - *Optional*
+- `premium-code` - the premium code you would like to use - *Optional* - make sure you use the proper premium code associated with the campaign
+- `premium-name` - the name of the premium offered to donors - *optional*
+- `premium-image-url` - the custom image to appear when selecting a premium - *optional*
 - `tsys-device` - the device name that corresponds to the TSYS Merchant Account which will be used for tokenizing your site's credit cards with TSYS - **Required** - Will be provided by DPS when adding your domain to the TSYS whitelist. `cru` is the default and corresponds with Cru's main TSYS Merchant ID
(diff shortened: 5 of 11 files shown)


src/common/components/premiumSelect/premiumSelect.component.js
```diff
@@ -0,0 +1,24 @@
+import angular from 'angular'
+
+import template from './premiumSelect.tpl.html'
+
+const componentName = 'premiumSelect'
+
+class PremiumSelect {
```
wrandall22, line 9: For the sake of naming convention, this should probably be `PremiumSelectController`

src/common/components/premiumSelect/premiumSelect.component.js
```diff
@@ -0,0 +1,24 @@
...
+  /* @ngInject */
+  constructor ($log, $scope) {
+    this.$log = $log
+    this.$scope = $scope
```
wrandall22, line 13: It looks like these aren't used.

src/common/components/premiumSelect/premiumSelect.tpl.html
```diff
@@ -1,26 +1,29 @@
...
-      <label class="radio-inline">
-        <input
-          type="radio"
-          name="premiumSelect"
-          ng-model="$ctrl.premiumSelected"
-          ng-value="false">
-          <translate>{{'NO_THANK_YOU'}}</translate>
-      </label>
+    <div class="col-sm-4">
+        <img ng-src="{{$ctrl.premiumImageUrl}}" style="width: 200px; height: 200px" />
```
wrandall22, line 28: Are these width/height from the current FL site?

src/common/services/api/order.service.js
```diff
@@ -253,6 +253,7 @@ class Order {
         const postData = cvv ? { 'security-code': cvv } : {}
         postData['cover-cc-fees'] = !!this.retrieveCoverFeeDecision()
         postData['radio-call-letters'] = this.retrieveRadioStationCallLetters()
+        postData['premium-code'] = this.retrievePremiumCode()
```
wrandall22, line 256: This isn't here, it is at the same place `campaign-code` is at.

src/app/checkout/step-3/step-3.component.js
```diff
@@ -190,6 +190,7 @@ export default angular
       onSubmitted: '&',
       onSubmittingOrder: '&',
       submittingOrder: '<',
+      premiumName: '<',
```
wrandall22, line 193: I don't think we want to display premiums on standard checkout for now. So no changes to `step-3`.

src/app/branded/step-2/branded-checkout-step-2.spec.js
```diff
@@ -47,6 +50,16 @@ describe('branded checkout step 2', () => {
     })
   })
 
+  describe('loadPremium', () => {
+    it('should load radio station name', () => {
```
wrandall22, line 54: Name is wrong here (copy/paste error)

wrandall22 (changes requested):
Looks like there is also some code that was added but not tested.

src/app/checkout/step-3/step-3.component.js
reldredge71: `branded-checkout-step-2` is basically just a wrapper around `checkout/step-3`.  I want to show the premium for branded checkout, but not for standard checkout.  So I added conditional logic to `step-3` so that it would render the premium section only if we pass `premiumName` to it.  Would that be fine?  Otherwise, I'm not sure where else it make sense to display the premium in https://github.com/CruGlobal/give-web/blob/master/src/app/branded/step-2/branded-checkout-step-2.tpl.html

wrandall22: Ah, I see. The one thing that would need to be tested with that is switching between branded and standard checkout. If you're storing premium name in local storage, it will still be populated when going to standard checkout, right? Though, given my other comments in this PR, you might be able to get rid of it from local storage. Even so, it would be stored on the gift and we'd have the same issue. At that point we have 2 options: show it on standard checkout in the rare cases it is still in there, but do it nicely; or put some sort of variable in to know whether or not it is branded checkout and only show it there.

src/common/services/api/order.service.js
reldredge71: I saw that `campaign-code` is added to `itemConfig` in this file: https://github.com/CruGlobal/give-web/blob/master/src/app/branded/step-1/branded-checkout-step-1.component.js#L34
Should I add `premium-code` to itemConfig instead?  I was trying to follow the app flow to see what happens to the properties in `itemConfig`, I can't see where this data gets submitted to the API.

src/app/checkout/step-3/step-3.component.js
reldredge71: Yeah I'm thinking that we don't want to display it or submit it if it's not branded checkout.  One idea I have is to handle it similar to `checkout/step-2`: step 2 has a `brandedCheckoutItem` property that essentially kills two birds with one stone.  It lets step 2 know that it's branded checkout, and it contains the relevant data needed for branded checkout.  I could do something similar here.  It just so happens that `brandedCheckoutItem` in this context is set to `itemConfig`, which seems to be the place I should be storing `premiumCode`.

reldredge71: https://github.com/CruGlobal/give-web/blob/master/src/app/branded/step-1/branded-checkout-step-1.tpl.html#L37

src/common/services/api/order.service.js
wrandall22: Yes, `itemConfig` is the place to put it.

src/app/checkout/step-3/step-3.component.js
wrandall22: I'm ok with you using that.

src/app/branded/step-2/branded-checkout-step-2.component.js
```diff
@@ -33,6 +34,10 @@ class BrandedCheckoutStep2Controller {
       )
   }
 
+  loadPremium () {
+    this.premiumName = this.orderService.retrievePremiumName()
```
wrandall22, line 38: `orderService.retrievePremiumName()` does not exist.

wrandall22:
You should be able to test this locally against `give-stage2` since the server code is already there.

src/app/branded/step-1/branded-checkout-step-1.spec.js
```diff
@@ -271,6 +271,30 @@ describe('branded checkout step 1', () => {
     })
   })
 
+  describe('onSelectPremiumOption', () => {
+    beforeEach(() => {
+      //jest.spyOn($ctrl, 'onSelectPremiumOption').mockImplementation(() => {})
```
wrandall22, line 276: Either delete or uncomment.

src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -150,6 +150,22 @@
     </div>
   </div>
 
+  <div class="mb" ng-if="!!$ctrl.brandedCheckoutItem['premium-code']">
```
wrandall22, line 153: Am I understanding this correctly that, if they chose to add a premium, this section will show up, but if they chose not to add a premium, this section will not show up?

wrandall22 (changes requested):

src/app/checkout/step-3/step-3.tpl.html
reldredge71: Yes, that is what I was thinking.  Although maybe we can render this section if it is branded checkout and if the branded checkout configuration has premium data; and if the premium was not selected, we could show the phrase "none" in this box?

wrandall22: It should probably be asked in Jira, but I don't think it should hold up putting something on staging necessarily if we don't get an answer before fixing the other issues here.

wrandall22:
In terms of deploying to stage. My preference would be to have you merge this into `staging` locally (make sure to pull the latest `staging` first) then push `staging`. Then we should switch the base of this PR to `master`.

reldredge71:
Sounds good.  Is there anything else you want me to change first?  Would you want me to change the image size at all?

wrandall22:
I think QA can flush out the image size. I don't see a premium on their site right now to compare.

PR #924, opened 2021-08-12
GTM-Add-Credit-Card-Cover-Fee-Event

TheNoodleMoose:
This adds a new GTM event for cover fees, but only when the user has actually covered credit card fees for that transaction. This event still needs to be created in GTM, then QA can begin.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -509,6 +509,8 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
     },
     transactionEvent: function (purchaseData) {
       try {
+        // The value of whether or not user is covering credit card fees for the transaction
+        const coverFees = localStorage.getItem('coverFees')
         // Parse the cart object of the last purchase
         const transactionCart = JSON.parse(localStorage.getItem('transactionCart'))
         // The purchaseId number from the last purchase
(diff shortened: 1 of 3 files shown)


src/app/analytics/analytics.factory.js
```diff
@@ -509,6 +509,8 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, sessionSer
     },
     transactionEvent: function (purchaseData) {
       try {
+        // The value of whether or not user is covering credit card fees for the transaction
+        const coverFees = localStorage.getItem('coverFees')
```
wrandall22, line 513: Is this triggered before or after [this line](https://github.com/CruGlobal/give-web/blob/master/src/app/checkout/step-3/step-3.component.js#L146)?

wrandall22: Also, see `retrieveCoverFeeDecision`. This should get the data the same way.

TheNoodleMoose: Ah looks like it's called after 🙁 I thought I was getting a value but testing it again now I'm getting ```null```.  Do you see another way to approach it? Maybe setting a temp sessionStorage with the value of ```coverFees``` and then deleting it after the transaction event?

wrandall22: 2 options:

1. Move the deletion of the variable
2. Make another event

wrandall22:
Please add the `on staging` label when you push this to stage.

wrandall22 (approved):

TheNoodleMoose:
This has been QA'd, but I'll merge it on Monday.

PR #928, opened 2021-08-18
Update staging domain

wrandall22:
It may or may not also be worth updating the `imgDomain` from prod to staging. This could mean some up front work to migrate some images into the staging environment that aren't already there, which may not be worth it at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,4 +38,5 @@ deploy:
       branch: master
 
 after_deploy:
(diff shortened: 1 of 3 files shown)


wrandall22:
This PR also required updating the `S3_GIVE_DOMAIN` environment variable in Travis.

reldredge71 (approved):

PR #932, opened 2021-08-20
EP-2261: Updates to Premium Selection

reldredge71:
Applies these changes:
- change the "Premium" header to "Choose a Resource"
- add single quotes around premium name
- allow the premium image height to resize itself and maintain aspect ratio with width

### DIFF
--- a/src/app/branded/step-1/branded-checkout-step-1.tpl.html
+++ b/src/app/branded/step-1/branded-checkout-step-1.tpl.html
@@ -57,7 +57,7 @@
                 ng-model="$ctrl.premiumSelected"
                 ng-value="true"
                 ng-change="$ctrl.onSelectPremiumOption()">
-                {{$ctrl.premiumName}}
+                {{$ctrl.premiumSelectLabel}}
             </label>
           </div>
           <div class="row">
--- a/src/app/branded/step-1/branded-checkout-step-1.tpl.html
+++ b/src/app/branded/step-1/branded-checkout-step-1.tpl.html
@@ -48,7 +48,7 @@
     <div class="panel-body">
       <h3 class="panel-name" translate>{{'CHOOSE_RESOURCE'}}</h3>
       <div class="row">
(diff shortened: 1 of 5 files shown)


src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -159,7 +159,7 @@
       <div class="panel-body">
         <div class="row">
           <div class="col-md-12">
-            {{$ctrl.premiumName}}
+            {{$ctrl.premiumSelectLabel}}
```
wrandall22, line 162: Seems like you could have put the single quotes here instead of making a new variable.

wrandall22: Though, I suppose it's in 2 places. So, either way is probably ok.

src/app/branded/step-1/branded-checkout-step-1.tpl.html
```diff
@@ -73,7 +73,7 @@ <h3 class="panel-name" translate>{{'CHOOSE_RESOURCE'}}</h3>
           </div>
         </div>
         <div class="col-sm-4">
-          <img ng-src="{{$ctrl.premiumImageUrl}}" style="width: 200px; height: 200px" />
+          <img ng-src="{{$ctrl.premiumImageUrl}}" style="width: 200px" />
```
wrandall22, line 76: It's probably better to make a class so that ministries can override this style.

src/app/checkout/step-3/step-3.tpl.html
reldredge71: I wasn't positive that I could use that string concatenation technique inside the html file, but if I can do that, then I would prefer to not create a new variable.

wrandall22: You don't need string concatenation, it's just HTML so you can put the quotes outside of the `{{}}`

PR #936, opened 2021-09-15
Update Sign In Wording

reldredge71:
Use updated wording for the Sign In screen headers, descriptions, and button labels.


wrandall22:
Are you planning to do the rearrangement in a separate PR? I think I would lean towards the same PR but a second commit.

reldredge71:
From some of the comments in Jira, it sounded like they mentioned the rearrangement but were only asking for the rewording at the moment: https://jira.cru.org/browse/EP-2273?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true But I can do the rearrangement as well.

wrandall22:
The rearrangement is the main part of this task, the rewording is kind of just tacked on.

src/app/signIn/signIn.tpl.html
```diff
@@ -5,9 +5,16 @@
         <div class="panel sign-in-panel">
           <div class="panel-body">
             <div class="row vdivide">
+              <div class="col-md-6 sign-in-section sign-in-hide-for-desktop">
```
reldredge71, line 8: I basically duplicated the "New Partners" section, and added conditional classes that hides one and shows the other.  Ideally it would be great to have just one instance of this section and have conditional logic that flips the order of the sections... but this was what I could find.

wrandall22: It sounded like they would be ok with rearranging the two on desktop as well during our last meeting.

wrandall22:
Good for staging.

wrandall22:
Did this get merged/tested in stage?

reldredge71:
I've tested it locally, I haven't merged it to stage yet.  I probably should have taken off the "On Staging" label when I made the additional commits.

wrandall22 (approved):

PR #940, opened 2021-10-04
EP-2293: Integrate GitHub Actions

reldredge71:
Setup GitHub Actions for Give-Web.  Tried to replicate the Stage and Prod build commands from Travis.


.github/workflows/build.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: pip install --user awscli
+      
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        
+      - name: Deploy to S3
+        run: aws s3 sync ./public/ s3://cru-givestage --delete
```
frett, line 44: I'm pretty sure this gets run on PRs against staging, you should probably only trigger the deploy logic if it is a push to staging.

frett:
It probably makes sense to separate out the lint checks, tests, and codecov upload to a separate job, then make the stage_build and prod_build jobs dependent on it.

the static checks job shouldn't need an `if` conditional, it can always run. The 2 deploy jobs would need an if conditional that only runs it if it's for pushes on the appropriate branch.

.github/workflows/build.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: codecov/codecov-action@v2
+      
+      - name: Install AWS CLI
+        run: pip install --user awscli
+      
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
TheNoodleMoose, line 41: I'm curious if this is needed. I'm attempting to just run the sync commands with ```env:``` and then including the access keys that way. If it doesn't work for me though guess I know why haha.

reldredge71: Yeah this is what I saw was recommended, I think this is fine as it is, but if we find a way that is cleaner then we can make that change.

.github/workflows/build.yml
```diff
@@ -0,0 +1,96 @@
...
+          node-version: 8
+      
+      - name: Install Dependencies
+        run: yarn install --immutable --immutable-cache
+      
+      - name: Build App
+        run: yarn run build 
+      
+      - name: Upload Codecov Reports
+        uses: codecov/codecov-action@v2
```
frett, line 56: this probably won't do anything, codecov should be run as a step on the test job

frett: codecov is also probably not dependent on `yarn run build`

.github/workflows/build.yml
```diff
@@ -0,0 +1,96 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        
+      - name: Deploy to S3 | Staging
+        if: github.ref == 'refs/heads/staging'
+        run: aws s3 sync ./public/ s3://cru-givestage --delete
+
+      - name: Deploy to S3 | Production
+        if: github.ref == 'refs/heads/master'
+        run: aws s3 sync ./public/ s3://cru-giveprod --delete
```
OzzieOrca, line 88: I think `--delete` could be somewhat risky on production. We don't want to delete files before new ones are deployed or have the deploy fail halfway or something. https://github.com/aws/aws-cli/issues/1417

Keeping it on staging is fine... I always thought we should find a better process to cleanup old files on S3 but there's no automated way currently. We could take the suggestion from that issue and run the sync then the sync again right after with `--delete`. Or we could just migrate to AWS Amplify or something where we don't have to worry about this implementation detail.

OzzieOrca (changes requested):
It's great to see things moving to Github Actions :) Thanks for working on this.

I think there's some value in keeping config files similar across projects. Idk if there's a way to share some of the config across projects. Maybe we could look at making a marketplace action (or something similar) eventually 🤷‍♂️ Although maybe things are different enough, it's not worth it.

Maybe peruse some other repos for prior art and see if there's anything we want to keep similar, reuse, or improve on:
https://github.com/CruGlobal/conf-registration-web/pull/760
https://github.com/CruGlobal/mpdx-react/blob/main/.github/workflows/ci.yml
https://github.com/CruGlobal/missionhub-react-native/blob/develop/.github/workflows/ci-cd.yml

.github/workflows/build.yml
```diff
@@ -0,0 +1,96 @@
...
+        run: yarn install --immutable --immutable-cache
+      
+      - name: Build App
+        run: yarn run build
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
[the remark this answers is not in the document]
Omicron7: Assuming a role is more secure than us creating machine users in AWS to handle this. If GitHub supports it, I think we would gladly transition to it.

I'm not sure how this app was setup previously, but where we can generate individual users/api keys per environment, It increases the security.

Here's an example of multiple GitHub environments with specific env AWS creds and the Terraform used to generate them.
https://github.com/CruGlobal/conf-registration-web/settings/environments
https://github.com/CruGlobal/cru-terraform/blob/master/applications/conf-registration-web/stage/github.tf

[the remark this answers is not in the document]
OzzieOrca: Oh, cool. I was unaware Github had environment secrets. And that Terraform can create them, that's really nice :)

Let me know if you want me to help look into the assume role stuff any more. The Github Action config seems pretty simple once we have a role and OIDC configured (which idk much about for AWS).

.github/workflows/build.yml
frett: > I would just run the Codecov bash script after the tests finish running.

You might have meant the codecov action, but just wanted to clarify.

The official codecov action (which Robert is currently using) is a good alternative to manually running the bash script. I'm using it already in our android/kotlin repos and have had no issues with it.

.github/workflows/build.yml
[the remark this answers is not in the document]
frett: the knowledge gained from figuring out the OIDC assume role stuff would probably be beneficial in other places besides just deploying code to S3

.github/workflows/build.yml
OzzieOrca: I've only used the bash script. Obviously I haven't been following what Codecov has been doing recently as they released a new version of the uploader and are deprecating the bash uploader. It won't function after Feb 1st 2022 with intentional brownouts starting last month.
https://about.codecov.io/blog/introducing-codecovs-new-uploader/
https://docs.codecov.com/docs/about-the-codecov-bash-uploader

Sounds like we should migrate everything to the Codecov Github Action. Thanks for bringing that up :)

.github/workflows/build.yml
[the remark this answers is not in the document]
frett: @reldredge71 connect with @Omicron7 he was working on that yesterday and I believe has a working solution in place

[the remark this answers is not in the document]
Omicron7: Yes, we can and should set it up for give-web, and for any repo using GitHub Actions to connect to AWS. Here's the infrastructure PR that added it to conf-registration-web. https://github.com/CruGlobal/cru-terraform/pull/2022
I'll (or you can make a PR) need to do similar for any app that wants to use assumed roles.

.github/workflows/build.yml
```diff
@@ -69,8 +69,12 @@ jobs:
...
         if: github.ref == 'refs/heads/staging'
         run: aws s3 sync ./public/ s3://cru-givestage --delete
-
-      - name: Deploy to S3 | Production
+      
+      - name: First Deploy to S3 | Production
+        if: github.ref == 'refs/heads/master'
+        run: aws s3 sync ./public/ s3://cru-giveprod
+      
+      - name: Second Deploy to S3 | Production
```
OzzieOrca, line 77: This code looks good but idk how self explanatory it is. Maybe adjust the name or add a comment to make it clear that we're making sure the new files are all available before starting to delete any.

.github/workflows/build.yml
```diff
@@ -0,0 +1,87 @@
...
+      
+      - name: Run Tests
+        run: yarn test
+      
+      - name: Upload Codecov Reports
+        uses: codecov/codecov-action@v2 
+   
+  deploy:
+    runs-on: ubuntu-latest
+    needs: [lint, test, codecov]
```
OzzieOrca, line 48:
```suggestion
    needs: [lint, test]
```

.github/workflows/build.yml
OzzieOrca: Also let's do the Cloudfront invalidation before deleting any files on production. Idk if the CLI waits for the invalidation to finish but might be better to start it sooner. We don't want a file cached by Cloudfront referencing an uncached but deleted file.

As you can see this Webpack build/caching strategy could use some cleanup. Normally you cache `index.html` for a short time (say 3-5 mins) and then give every other file a unique hash in the filename and cache them for a year or more cuz if there are changes, they will end up with a new filename. I've setup apps so there's one main js/json file that loads everything else similar to how index.html would but this app hasn't been setup that way yet. Brian improved it at one point but there's still work to do as I don't think it's safe to cache everything for a year yet.

OzzieOrca (changes requested):

Omicron7:
@OzzieOrca Sweet, then I'll remove permissions for that S3 bucket and delete it.

PR #945, opened 2021-10-14
GitHub Actions -- Fixed

reldredge71:
I pushed a GHA implementation yesterday and it broke on staging and production.  I reverted those changes and made some changes here, I think I have the issues resolved.  I'll highlight the major differences here.


wrandall22:
Probably not easy to test rollbar, but can you see if the token is in the stage environment properly?

PR #948, opened 2022-01-27
Migrate From Node-Sass to Dart-Sass

reldredge71:
Node-Sass is becoming deprecated; it is recommended to migrate to Dart-Sass.

These changes are required along with this migration:
https://sass-lang.com/documentation/breaking-changes/extend-compound
https://sass-lang.com/documentation/breaking-changes/slash-div
https://github.com/foundation/foundation-rails/issues/296


reldredge71:
@wrandall22 @OzzieOrca @TheNoodleMoose I'm still getting deprecation warnings for node_modules.  the `quietDeps` option should disable those, but it has not worked yet.  It's possible that there is some issue with using`quietDeps` along with `MiniCssExtractPlugin`: https://github.com/sass/sass/issues/3065#issuecomment-910946898

src/assets/cru-scss/_subnav.scss
```diff
@@ -1,3 +1,5 @@
+@use 'sass:math';
```
OzzieOrca, line 1: You using math here?

src/assets/cru-scss/_supernav.scss
```diff
@@ -1,3 +1,5 @@
+@use 'sass:math';
```
OzzieOrca, line 1: Same here, I don't see math being used. (maybe an unchanged line is using it, I only scrolled through the whole first file).

src/assets/crubrand/generic/_tools.scss
```diff
@@ -17,12 +19,12 @@ $rem-base: 16px !default;
...
 // scss-lint:disable NameFormat 
 @function parseInt($n) { /* 2 */
-    @return $n / ($n * 0 + 1);
+    @return math.div($n, ($n * 0 + 1));
 }
 // scss-lint:enable NameFormat
 
 @function convert-to-rem($value, $base-value: $rem-base) {
-    $value: strip-units($value) / strip-units($base-value) * 1rem;
+    $value: math.div(strip-units($value), strip-units($base-value) * 1rem);
```
OzzieOrca, line 27: I don't think order of operations affects this but I probably would have had the ` * 1rem` after the `math.div` function call instead of as the second parameter.

src/assets/crubrand/generic/_tools.scss
```diff
@@ -75,7 +77,7 @@ $rem-base: 16px !default;
  * Use this em function instead of the other
  */
 @function em($value, $context: $rem-base) {
-    $value: strip-units($value) / strip-units($context) * 1em;
+    $value: math.div(strip-units($value), strip-units($context) * 1em);
```
OzzieOrca, line 80: Same here.

src/assets/crubrand/generic/mixins/_media-query.scss
```diff
@@ -17,13 +19,13 @@ $query-unit: 16px;
             @content;
         }
    
-        @media screen and (#{$query1}-#{$query2}: strip-units($point) / strip-units($query-unit) + em) {
+        @media screen and (#{$query1}-#{$query2}: list.slash(strip-units($point), strip-units($query-unit) + em)) {
```
OzzieOrca, line 22: What does `list.slash` do? This looks like a division...

src/assets/crubrand/generic/mixins/_sizing.scss
```diff
@@ -29,7 +31,7 @@ $sizeValue: strip-units(16px);
 
             @if $unit == "px" {  /* 7 */
                 $px : append($px, $value);
-                $rem: append($rem, ($val / $sizeValue + rem));
+                $rem: append($rem, math.div($val, $sizeValue + rem));
```
OzzieOrca, line 34: I think you have the order of operations here wrong. https://sass-lang.com/documentation/operators#order-of-operations

OzzieOrca (changes requested):
Thanks for updating all those. That was a lot of CSS to go through :)

Idk what's going on with `quietDeps`... The config looks right... https://sass-lang.com/documentation/js-api/interfaces/Options#quietDeps I wonder if somehow the directory config is messed up for this project and it's not identifying `node_modules` right. I could do some digging but I'm not sure what to suggest right now...

src/assets/cru-scss/_subnav.scss
reldredge71: Hmm I must've had a reason for adding it, but I thought I only added it to files that I was removing the `/` from.  Like you said, it was a lot of CSS to go through!  I'll remove it.

src/assets/crubrand/generic/mixins/_media-query.scss
reldredge71: Yes, it should behave the same way as a division.  I got an error saying that `math.div` should only be used for number variables, and I guess it didn't know how to process the results of `strip-units`.  then again, `strip-units` is used in other places and it doesn't seem to cause an issue, I'll have to look into that.

I'm actually noticing another order-of-operations issue here -- I assumed that `+ em` was part of the right side of the `/` operator, but now I realize that it should be it's own thing.

src/assets/crubrand/generic/mixins/_sizing.scss
reldredge71: yeah, I'm realizing now that I kept thinking of the additions as part of the right-side of the `/` operator.  I know order of operations, but I guess I was imagining parentheses around `$sizeValue + rem`.

OzzieOrca (dismissed):

src/assets/crubrand/generic/mixins/_media-query.scss
```diff
@@ -1,3 +1,6 @@
+@use 'sass:list';
```
wrandall22, line 1: This appears to be unused now.

wrandall22:
Make sure to merge to staging and test there before merging to master.

reldredge71:
I wish I could quiet the warnings from node_modules, but at least they don't hold it back from building properly.

I'm noticing one specific layout issue after this migration.  Some items are spilling over the right edge of their containers.  I inspected these elements compared to give-stage2.cru.org, and the only thing that stands out to me is that these elements are missing the CSS property `box-sizing: border-box` post-migration.  I would like to figure out why that is before merging to staging.

Screen Shot 2022-01-28 at 3 35 22 PM

Screen Shot 2022-01-28 at 3 35 00 PM

src/assets/crubrand/generic/bourbon/addons/_prefixer.scss
```diff
@@ -3,6 +3,12 @@
 //************************************************************************//
 // Variables located in /settings/_prefixer.scss
 
+$prefix-for-webkit:    true !default;
+$prefix-for-mozilla:   true !default;
+$prefix-for-microsoft: true !default;
+$prefix-for-opera:     true !default;
+$prefix-for-spec:      true !default;
```
reldredge71, line 10: I'm trying to find the right way to setup these variables.  I think they're supposed to be loaded from src/assets/crubrand/generic/bourbon/settings/_prefixer.scss.  However, I was noticing strange behavior in the layout of some elements, which I traced back to the prefixer mixin in this file, and explicitly declaring these variable defaults in this file fixed the issue.

On a related note, I'm getting deprecation warnings for this file as well, in the disable-prefix-for-all mixin:

Screen Shot 2022-02-04 at 9 51 30 AM

It's complaining about this same list of variables, and about trying to assign them globally.  I can look into this deeper, but if anyone has any insight on it I would appreciate it.

reldredge71: BTW these changes that I made to `_prefixer.scss` fixed the weird `box-sizing` issue I mention above.

wrandall22: Are there adverse effects to removing the `@mixin disable-prefix-for-all() {` block? I'm guessing it was put there for a reason, but I was not part of this codebase when it was being written.

reldredge71: It looks like it gets referenced here: https://github.com/CruGlobal/give-web/blob/master/src/assets/crubrand/generic/bourbon/css3/_keyframes.scss.  That file also seems to try to reference the same variables declared in https://github.com/CruGlobal/give-web/blob/master/src/assets/crubrand/generic/bourbon/settings/_prefixer.scss.

reldredge71: If that's the case, I could try copying the default values there as well, but it would probably be better if they were defined in one place that both files can access.

wrandall22: https://stackoverflow.com/questions/60422903/cannot-make-variables-globally-accessible-in-sass

wrandall22:
Do the updates still have the problems resolved?

reldredge71:
Yes, what you suggested worked great!  Sorry, I should have made that clear.

wrandall22:
You can merge this to staging if you want

reldredge71:
This is looking good on staging to me

wrandall22 (approved):

PR #963, opened 2022-02-15
EP-2274: Convert from needinfo to messages advise

senyor:


src/common/services/api/order.service.js
```diff
@@ -282,17 +282,21 @@ class Order {
     return this.cortexApiService.get({
       path: ['carts', this.cortexApiService.scope, 'default'],
       zoom: {
-        needInfo: 'order:needinfo[]'
+        needInfo: 'order'
```
wrandall22, line 285: Are we not able to do `order:messages[]`? If not, we should label this `order` instead of `needInfo`.

PR #965, opened 2022-02-23
Full nonprod env

wrandall22:
This PR adds the necessary values to run as a parallel environment to staging.


src/common/app.config.js
```diff
@@ -21,7 +21,10 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
         'uatpub.aws.cru.org',
         'uatdisp.aws.cru.org',
         'cru-givestage.s3.amazonaws.com',
-        'give-stage-static.cru.org',
+        'give-stage-static.cru.org'
+      ],
+      nonprod: [
+        'give-stage2-next.cru.org',
         'cru-give-web-assets-nonprod.s3.amazonaws.com',
         'give-nonprod-static.cru.org'
```
reldredge71, line 29: Are these two values still needed in the `staging` environment?

wrandall22: I don't think so. The way I read the docs for this plugin is that these arrays are telling the system which domains go to which environment. Once it figures out which environment is active, it pulls those env vars from the JSON object below.

reldredge71 (approved):

src/common/app.config.js
wrandall22: Exactly.

PR #966, opened 2022-02-23
EP-2332: Block Anonymous Donors from Sending Recipient Comments

reldredge71:
https://jira.cru.org/browse/EP-2332

Temporary fix for the use case where a donor indicates that they wish to be anonymous in the Special Handling comments, yet also wishes to send comments to the recipient.  All messages sent directly to the recipient have the donor's name on it, and so it does not honor the donor's request to be left anonymous.

The fix for now is to disable Recipient comments if there is anything at all in the Special Handling comments.  At the moment we don't have a good way to determine if the Special Handling comments contain anything about wishing to be anonymous, but for now we can simply prevent any message being sent to the recipient if there is anything the donor wishes to communicate to Special Handling first.


src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -292,6 +292,12 @@ class ProductConfigFormController {
     this.updateQueryParam({ key: giveGiftParams.month, value: month })
   }
 
+  changeComments () {
+    if (this.itemConfig['donation-services-comments'] != '') {
+      this.itemConfig['recipient-comments'] = ''
```
reldredge71, line 297: I am a bit concerned about essentially deleting everything the donor has typed already if they start adding donation-services-comments.  But it's important that we remove recipient-comments from itemConfig if there are donation-serivces-comments.  Maybe I can temporarily store recipient-comments locally, so that they can be re-loaded if donation-services-comments are deleted?

wrandall22:
I see a lot of `!= ''` going on here. Should we drop and check for not falsy instead? Or should we do `!== ''`?

src/common/app.config.js
```diff
@@ -106,6 +106,7 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
     SEND_MESSAGE_TO: 'Send a Message to {{ministry}}',
     SPECIAL_INSTRUCTIONS: 'Special Handling Instructions for Processing This Gift',
     MESSAGE_EXAMPLE: 'For example: make this gift anonymous, stop my gift after 18 months, etc.',
+    COMMENTS_DISABLED: 'Comments are disabled when Special Handling Instructions are being sent.',
```
wrandall22, line 109: Should probably put this on both translations (even though we only have English right now).

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -292,6 +292,12 @@ class ProductConfigFormController {
     this.updateQueryParam({ key: giveGiftParams.month, value: month })
   }
 
+  changeComments () {
```
wrandall22, line 295: I think `clearComments` or something similar might be a better name.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: That's an interesting thought. It would certainly be a more complicated solution. Maybe we get this live, then go back and do that? How much effort do you think it would be to have that logic?

wrandall22 (changes requested):

src/app/productConfig/productConfigForm/productConfigForm.component.js
reldredge71: It might not be so bad.... it would probably be handled inside this method here, at some point I would store the value of `itemConfig['recipient-comments']` inside of `this`, and I'll have another conditional here that will bring that value back to `itemConfig['recipient-comments']` once `itemConfig['donation-services-comments']` is empty.  It could get complex making sure the logic is sound, though.

reldredge71:
Right, empty string is a falsy value.  That might be a cleaner way of expressing it.  And yes I'll add the unit test now.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: You can take a quick pass at it if you want. Even if there are edge cases, at least some people would get their comments back. You might want them in session storage vs being in memory, so if they go back after navigating around, the comments will still be there.

wrandall22:
Go ahead and throw this on staging to get QA feedback from Mike.

reldredge71:
@wrandall22 I had to make a minor fix to the logic in the html file.  If this looks good to you, I can make another push to staging and have Mike review it.

wrandall22:
Yep, push to staging 👍

wrandall22 (approved):
I'm fine with merge when it passes QA by Mike.

PR #969, opened 2022-03-09
Convert AddressForm From Angular to React

reldredge71:
Here is what I have so far for converting the AddressForm component into React.  Right now this is being inserted into the CreditCardForm Angular component, but it shouldn't be too difficult to replace the Angular AddressForm throughout the app.
As a follow-up to this work, I would like to also:
- Create a GraphQL proxy of the Geographies API service
- Continue building out other components related to Payment Methods


src/common/components/addressForm/addressForm.react.tsx
```diff
@@ -64,6 +65,47 @@ const AddressForm = ({
...
+    if (values.intAddressLine4 && values.intAddressLine4.length > 100) {
+      errors.intAddressLine4 = 'MAX_LENGTH_ADDRESS_OTHERS_ERROR';
+    }
+    if (!values.locality) {
+      errors.locality = 'CITY_ERROR';
+    } else if (values.locality.length > 50) {
+      errors.locality = 'MAX_LENGTH_CITY_ERROR';
+    }
+    if (loadingRegionsError) {
+      errors.country = 'REGIONS_LOADING_ERROR';
```
wrandall22: `errors.region`

src/common/components/addressForm/addressForm.react.tsx
```diff
@@ -0,0 +1,305 @@
...
+        handleBlur,
+      }) => (
+        <>
+          <FormikAutoSave debounceMs={600} />
+          <div className="row">
+            <div className="col-sm-12">
+              <CountrySelect
+                addressDisabled={addressDisabled}
+                countries={countries.map(country => ({ name: country.name, displayName: country['display-name']}))}             
+                onChange={handleChange}
```
wrandall22, line 170: This should call `refreshRegions` I think. Not seeing how that is being done here.

src/common/components/addressForm/addressForm.react.tsx
```diff
@@ -0,0 +1,305 @@
...
+    if(a['display-name'] < b['display-name']) return -1;
+    if(a['display-name'] > b['display-name']) return 1;
+    return 0;
+  };
+
+  const AddressSchema = Yup.object().shape({
+    country: Yup.string()
+      .required('You must select a country'),
+    streetAddress: Yup.string()
+      .max(200, 'This field cannot be longer than 200 characters')
```
wrandall22, line 76: Can this use the `maxLength` property instead of hard-coded `200`?

wrandall22 (commented):
Looks like translations are still to be figured out

src/common/components/form/selectInput/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          { options.map(optionRow) }
+        </select>
+        { error && (
+          <div role="alert">
+            <div className="help-block">
+              <span>{error}</span>
+              { retry && (
+                <button
+                  id="retryButton"
```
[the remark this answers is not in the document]
reldredge71: Good point, this might have to be more dynamic to account for multiple instances.

src/common/components/form/textInput/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  required = false,
+  maxLength,
+  pattern,
+  disabled,
+  onChange,
+  onBlur,
+  value,
+  error,
+}: TextInputProps) => (
+  <div className={`form-group${required && ' is-required' || ''}${error && ' has-error' || ''}`}>
```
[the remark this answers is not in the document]
reldredge71: I could maybe remove this wrapping `div` from here, and create a wrapper component that will only be used for AddressForm text inputs.

src/common/components/form/textInput/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    { title && <label>{title}</label> }
+    <input
+      type={type}
+      className="form-control form-control-subtle"
```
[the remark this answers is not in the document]
reldredge71: Yes, I could create an optional prop that overrides the input className value.

src/common/components/addressForm/addressForm.react.tsx
reldredge71: I noticed that too, I'll push something that should address that.

src/common/components/addressForm/addressForm.react.tsx
reldredge71: I don't think the `maxLength` property on the input itself is accessible in this scope.

What I've also noticed is that setting `maxLength` on the input field prevents the user from entering any more characters than that limit.  So I don't think this error will even show up at all.  I think the input on its own does a good job of preventing the max length from being exceeded.

PR #972, opened 2022-04-25
EP-2214 - Fix the receipt summary lookup endpoint

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2214)

The endpoint switched from `/items` to `/form` with the upgrade.


senyor (approved):
Nice

PR #973, opened 2022-05-05
HS-756593 - Fix campaign code handling

wrandall22:
Problem: When giving by the Give a Gift button from the search page, the default campaign code is not being sent to the API.

Basically, the code was all there to make this work correctly, but being called in the wrong order when opening the product config modal only.

[Helpscout](https://secure.helpscout.net/conversation/1870905827/756593?folderId=381778)


wrandall22:
Currently, this is introducing duplicated code which I would like to get rid of, but need to test in stage first to see if it's possible.

gyasistory (approved):

PR #974, opened 2022-05-20
Fix add to cart issues

wrandall22:
During the upgrade, the fields for configuring an item for the cart changed from dashes to underscores


aaronlaib (approved):
Looks great, approved from phone :)

PR #975, opened 2022-05-27
Log any errors coming from logout during branded checkout

wrandall22:
This seems to fix things when using local overrides in the Chrome developer tools.

[Helpscout](https://secure.helpscout.net/conversation/1897464967/768063/)


wrandall22:
Doesn't seem to break anything in stage from my testing.

PR #978, opened 2022-06-24
EP-2353 - Do not send extra info to EP

wrandall22:
The origins of this PR was to remove things like `jcr:title` from going to EP and being stored in `tshoppingitemdata` and `torderitemdata`. However; while testing, I found more to do which was kind of related, so went ahead and fixed that too.

- Do not send AEM data to EP
- Do not send data we don't care about to EP (such as amount_with_fees)
- Do not send empty data to EP (yes, it does store it as empty string which we don't want)
- Properly capitalize some fields (EP 8.1 makes all caps) so that they repopulate correctly on editing a gift that is already in the cart

[Jira](https://jira.cru.org/browse/EP-2353)

### DIFF
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -31,10 +31,10 @@ class BrandedCheckoutStep1Controller {
 
   initItemConfig () {
     this.itemConfig = {}
-    this.itemConfig.campaign_code = this.campaignCode
-    if (this.itemConfig.campaign_code &&
-      (this.itemConfig.campaign_code.match(/^[a-z0-9]+$/i) === null || this.itemConfig.campaign_code.length > 30)) {
-      this.itemConfig.campaign_code = ''
+    this.itemConfig.CAMPAIGN_CODE = this.campaignCode
+    if (this.itemConfig.CAMPAIGN_CODE &&
+      (this.itemConfig.CAMPAIGN_CODE.match(/^[a-z0-9]+$/i) === null || this.itemConfig.CAMPAIGN_CODE.length > 30)) {
+      this.itemConfig.CAMPAIGN_CODE = ''
     }
     this.itemConfig['campaign-page'] = this.campaignPage
     this.itemConfig.amount = this.amount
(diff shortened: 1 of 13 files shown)


src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -31,10 +31,10 @@ class BrandedCheckoutStep1Controller {
 
   initItemConfig () {
     this.itemConfig = {}
-    this.itemConfig.campaign_code = this.campaignCode
-    if (this.itemConfig.campaign_code &&
-      (this.itemConfig.campaign_code.match(/^[a-z0-9]+$/i) === null || this.itemConfig.campaign_code.length > 30)) {
-      this.itemConfig.campaign_code = ''
+    this.itemConfig.CAMPAIGN_CODE = this.campaignCode
+    if (this.itemConfig.CAMPAIGN_CODE &&
+      (this.itemConfig.CAMPAIGN_CODE.match(/^[a-z0-9]+$/i) === null || this.itemConfig.CAMPAIGN_CODE.length > 30)) {
```
senyor, line 36: just to clarify if the regular expression should or shouldn't match

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -338,6 +339,16 @@ class ProductConfigFormController {
     })
   }
 
+  omitIrrelevantData (itemConfig) {
+    const data = this.productData.frequency === 'NA'
+      ? omit(itemConfig, ['RECURRING_START_MONTH', 'RECURRING_DAY_OF_MONTH', 'jcr-title', 'AMOUNT_WITH_FEES'])
```
senyor, line 344: is ok 'jcr-title' lower-case?

senyor (commented):
A couple of questions

src/app/branded/step-1/branded-checkout-step-1.component.js
wrandall22: This logic isn't changing, just went from lowercase to uppercase.

src/app/productConfig/productConfigForm/productConfigForm.component.js
wrandall22: Yes, it comes into the object lowercase and never makes it to the API (now) which is what wants uppercase.

PR #980, opened 2022-06-28
EP-2358 - Fix phone number update

wrandall22:
This is the first part of [EP-2358](https://jira.cru.org/browse/EP-2358) which updates the URLs and the property `spouse` to `is-spouse` to send data properly to the API. However, the API needs some work to handle the spouse property correctly when returning phone numbers to the client (see [EP-2360](https://jira.cru.org/browse/EP-2360)) which may require more changes associated with this Jira.


senyor:
GTG

senyor (approved):

PR #984, opened 2022-08-08
EP-2360: Use only main list of phones with provided boolean 'is-spous…

senyor:
…e' value


senyor:
I haven't tested this change yet.

wrandall22 (approved):

wrandall22:
I was able to test successfully.

senyor:
Also tested it successfully. Slight discrepancy with code coverage is ok.

PR #987, opened 2022-10-05
I have no idea if this would work.

EJTison:
I've only just begun looking at the front end, so this is all new to me.  I'm trying to disable the submit button if the amount is invalid, same as it currently gets disabled if a payment method is invalid.

### DIFF
--- a/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
+++ b/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
@@ -41,6 +41,10 @@ class EditRecurringGiftsController {
   allPaymentMethodsValid () {
     return every(this.recurringGifts, gift => gift.paymentMethod)
   }
+
+  allAmountsValid () {
+    return every(this.recurringGifts, gift => gift.amount)
+  }
 }
 
 export default angular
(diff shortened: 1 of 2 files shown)


EJTison:
@wrandall22  Definitely need some feedback on this.  I'm flying by the seat of my pants, haha

wrandall22:
This may not be something we want. I'll need to look at the code again, but it may be treating empty values as unchanged currently, which this would hinder.

EJTison:
> but it may be treating empty values as unchanged currently,

Yes, I believe you're right.

> which this would hinder.

From my understanding, this wouldn't have an impact on that?  This new method only gets called in the context of checking whether the form submit button should be disabled or not.  Theoretically this code shouldn't affect the actual processing of the data, later?

wrandall22:
There seems to be validation in place already that may not be firing when setting up a new gift [here](https://github.com/CruGlobal/give-web/blob/master/src/common/components/giftViews/giftUpdateView/giftUpdateView.tpl.html#L4-L20).

EJTison:
> There seems to be validation in place already that may not be firing when setting up a new gift [here](https://github.com/CruGlobal/give-web/blob/master/src/common/components/giftViews/giftUpdateView/giftUpdateView.tpl.html#L4-L20).

Yep, I looked at that.  That is working correctly, however it does nothing to actually disable the submit button.  So even though an error message displays and there are checks in place to ensure invalid strings are not entered, you can still just hit the "submit" button anyway.  That is what I am trying to remedy.

EJTison:
Tested in staging-- This change works perfectly for UPDATING gifts.  Does not seem to work for creating new ones though.  I guess I was mistaken that both actions used the same part of the code.  I’ll figure out a fix.

EJTison:
Lee said I should probably leave this for Bill's team to take over at this point, since this is not in my scope of work.
These last 4 commits are my incomplete attempt to set things up generally how they need to be, though I'm not sure how to finish setting up `configureRecentRecipients.component.js` and it's spec file properly...  Hopefully they are helpful, and don't add additional confusion.

PR #988, opened 2022-10-14
EP-2371: Fix UI does not populate payment methods

senyor:
… instead of payment-method-id for  recurring and historical donations where applicable


wrandall22:
The console.logs should be removed before merge.

wrandall22:
The flow at this point is:
- PR against `ep-upgrade`
- Test in staging by merging branch directly into `ep-upgrade-staging`
- Once PR is approved and tests work well, merge PR

wrandall22 (approved):

PR #989, opened 2022-10-17
Adapt to 8.1 payload

EJTison:


wrandall22 (approved):
Should squash these two commits together.

PR #990, opened 2022-10-18
EP-2379 Fill in Recent Designation Numbers by Adapting to 8.1 payload

EJTison:
https://jira.cru.org/browse/EP-2379 
This has been tested in 8.1 staging, and fixes the problem it is intended to.  I clicked around a little bit, but did not notice any additional bugs that have arisen as a result of this solution.  However, I did not _**exhaustively**_ test to ensure this is completely without side-effects.

### DIFF
--- a/src/common/services/api/donations.service.js
+++ b/src/common/services/api/donations.service.js
@@ -159,7 +159,7 @@ const DonationsService = /* @ngInject */ function (cortexApiService, profileServ
         return map(response.recipients, (recipient) => {
           return {
             'designation-name': recipient.definition['display-name'],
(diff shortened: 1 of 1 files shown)


wrandall22:
Did it test well in stage?

EJTison:
> Did it test well in stage?

Yes, sir!  The fix is working as intended, as far as I can tell.  As I mentioned, I can't guarantee this doesn't cause some OTHER bug, but I wasn't able to find it, if it does.

EJTison:
You're welcome to test it out in staging and ensure it wasn't a fluke :)

wrandall22 (approved):

PR #991, opened 2022-10-25
EP-2383: Fix template placeholder

senyor:
Fix template placeholder by not using 'translate' keyword, will apply proper translation in future sprints and verify for similar issues


wrandall22:
What is this trying to fix? On the edit gift modal I see this in stage:

Screen Shot 2022-10-25 at 4 47 05 PM

senyor:
Fix is merged into stage already.

wrandall22:
Ah, I see.

wrandall22:
This is also a problem in 6.15, so it will be good to have it fixed.

wrandall22 (approved):

PR #994, opened 2022-12-08
Make UI Expectation Match Back-End Expectation

EJTison:
@wrandall22 Would love to see if you agree that this is the most elegant way to solve [this problem](https://jira.cru.org/browse/EP-2393).

### DIFF
--- a/src/common/components/resetPasswordModal/resetPasswordModal.tpl.html
+++ b/src/common/components/resetPasswordModal/resetPasswordModal.tpl.html
@@ -18,12 +18,12 @@
                type="password"
                class="form-control form-control-subtle"
                ng-model="$ctrl.password"
-               minlength="8"
+               minlength="15"
                ng-pattern="/^(?=.*\d)(?=.*[a-zA-Z]).{6,}$/"
                required>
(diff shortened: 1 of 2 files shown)


wrandall22:
There is currently a code freeze on everything related to give site. This does look like a reasonable solution to the problem. This login form is going away when we move to Okta.

wrandall22 (approved):

EJTison:
> There is currently a code freeze on everything related to give site (so we couldn't put it out there in 6.15). 
I was unable to confirm if this issue exists in 6.15 as well, due to different resultant behavior.  I only know for sure that this exists in 8.1.
Is it alright to merge this for 8.1, despite the code freeze?
> This does look like a reasonable solution to the problem. This login form is going away when we move to Okta.
Ah I see... Interesting.  Will end-users to the give site be expected to use an Okta account, at that point?

wrandall22:
Yes, they will use an Okta account (the same as what they are using now) but will be using the Okta login form (new).

wrandall22:
This does exist in 6.15, but I think only certain kinds of users have the 15 character minimum.

wrandall22:
Also note that `ep-upgrade` is not the branch deploying to the staging environment.

PR #996, opened 2022-12-13
Only welcome if we know their name

EJTison:
This helps with https://jira.cru.org/browse/EP-2386

### DIFF
--- a/src/common/components/signInModal/signInModal.tpl.html
+++ b/src/common/components/signInModal/signInModal.tpl.html
@@ -5,7 +5,7 @@
     <hr class="horizontal-divider"/>
     <h4 class="text-center" translate>If you already have a Cru username, use it to sign in so we can link it to your giving</h4>
   </div>
-  <p ng-if="$ctrl.identified" class="text-center">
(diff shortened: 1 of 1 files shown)


EJTison:
Is this check too detailed, or merely sufficient?

src/common/components/signInModal/signInModal.tpl.html
```diff
@@ -5,7 +5,7 @@ <h4 class="text-center" translate>If you don&apos;t have a Cru username, you can
     <hr class="horizontal-divider"/>
     <h4 class="text-center" translate>If you already have a Cru username, use it to sign in so we can link it to your giving</h4>
   </div>
-  <p ng-if="$ctrl.identified" class="text-center">
+  <p ng-if="typeof $ctrl.sessionService.session.first_name != 'undefined' && $ctrl.sessionService.session.first_name && typeof $ctrl.sessionService.session.last_name != 'undefined' && $ctrl.sessionService.session.last_name" class="text-center">
```
wrandall22, line 8:
```suggestion
  <p ng-if="$ctrl.identified && $ctrl.sessionService.session.first_name && $ctrl.sessionService.session.last_name" class="text-center">
```

EJTison: Identified is the only time we will have their name?  What is "Registered" again?

wrandall22: Registered is if they are fully logged in. Identified is the only time they will need to login and we have their name.

wrandall22 (approved):

PR #999, opened 2023-02-10
[EP-2314] Require uploaded files to be images

canac:
One source of the designation editor photo upload errors is lack of proper file validation. Even with `accept="image/*"` browsers don't validate the MIME type when dragging and dropping files, so a user could upload non-image files that way. Also, Safari allows users to upload HEIC files, which causes a server-side exception when attempting to rotate it ([Rollbar example](https://rollbar.com/Cru/give-aem/items/53/)). Other image formats like SVG or WEBP would likely cause similar issues.

This PR reimplements the dead `angular-upload` package to add that additional validation, allowing only PNG and JPEG files to be uploaded.

https://jira.cru.org/browse/EP-2314

### DIFF
--- a/src/common/directives/imageUpload.directive.js
+++ b/src/common/directives/imageUpload.directive.js
@@ -1,0 +1,64 @@
+import angular from 'angular'
+
+const directiveName = 'imageUpload'
+
+// Based on https://github.com/leon/angular-upload
+// Fixes bug where even if accept is "image/*", you could still drag-and-drop non-image files onto the upload button
+const imageUpload = /* @ngInject */ ($http) => ({
+  restrict: 'EA',
+  scope: {
+    url: '@',
+    onUpload: '&',
+    onSuccess: '&',
+    onError: '&',
+    onComplete: '&'
+  },
+  link: (scope, element) => {
+    const validMimeTypes = ['image/jpeg', 'image/png']
+
+    const el = angular.element(element)
+    const fileInput = angular.element(`<input type="file" accept="${validMimeTypes.join(',')}" />`)
+    el.append(fileInput)
+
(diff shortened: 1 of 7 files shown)


wrandall22:
So, one thing to know is that the exception with rotation does not break the flow, it just logs an error out to Rollbar.

src/common/directives/imageUpload.directive.js
```diff
@@ -0,0 +1,64 @@
...
+    el.append(fileInput)
+
+    fileInput.on('change', (event) => {
+      const files = event.target.files
+      const file = files[0]
+      if (!file) {
+        return
+      }
+
+      if (!validMimeTypes.includes(file.type)) {
```
wrandall22, line 30: We should probably write multiple tests for this, 1 for each supported file type.

canac:
@wrandall22 So what image type validation should take place on upload if any? In my mind, we should definitely block non-image files. HEIC files should probably be blocked too because AEM can't serve them (500 error "MIMEType == null!") and desktop browsers can't render them anyway (including Safari on macOS). GIF works but SVG and WEBP are served with a 200 status code but an empty response and a `Content-Length` of 0. So PNG, JPEG, and GIF appear to be the only useable image formats.

wrandall22:
I think JPG and PNG are sufficient. I wonder if it would be useful to have a message spelling out the restrictions.

canac:
@wrandall22 I changed it to display a message when the user tries to upload an invalid file type.

wrandall22:
Looking pretty good. Feel free to throw it onto staging whenever for testing there.

wrandall22 (approved):

PR #1003, opened 2023-02-13
[no-jira] Update Jest to v29

canac:
As part of trying to get tests to work in a previous PR, I tried updating Jest. It didn't turn out to be necessary, but since I had already done the migration work, I'm going ahead and submitting those changes in this separate PR.


.tool-versions
```diff
@@ -1,2 +1,2 @@
-nodejs 16.0.0
+nodejs 16.10.0
```
wrandall22, line 1: It will be good to test and make sure this works in stage in the deployed environment. I don't imagine it will have issues though since it is just an S3 bucket.

PR #1005, opened 2023-02-17
[EP-2405] Don't set cart count cookie in branded checkout

canac:
Prevent branded checkout pages like `secure.cru.org` from setting the cart count cookie. That cookie was causing a (1) badge to show up on the `give.cru.org` cart even though the cart was empty.

https://jira.cru.org/browse/EP-2405

### DIFF
--- a/src/common/services/api/cart.service.js
+++ b/src/common/services/api/cart.service.js
@@ -24,21 +24,27 @@ import { startMonth } from '../giftHelpers/giftDates.service'
 
 const cartTotalCookie = 'giveCartItemCount'
 const cartTotalCookieDomain = 'cru.org'
+const brandedCheckoutDomains = ['secure.cru.org']
 
 const serviceName = 'cartService'
 
 class Cart {
   /* @ngInject */
-  constructor (cortexApiService, commonService, designationsService, sessionService, hateoasHelperService, $cookies) {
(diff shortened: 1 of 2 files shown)


src/common/services/api/cart.service.js
```diff
@@ -24,21 +24,27 @@ import { startMonth } from '../giftHelpers/giftDates.service'
 
 const cartTotalCookie = 'giveCartItemCount'
 const cartTotalCookieDomain = 'cru.org'
+const brandedCheckoutDomains = ['secure.cru.org']
```
wrandall22, line 27: I wonder if non-branded checkout domains would be more maintainable. Theoretically there could be infinitely many branded checkout domains, but a fairly fixed number of non-branded checkout domains.

canac: I wasn't sure whether to go with a whitelist or blacklist. Are there any other staging domains besides `give-stage2.cru.org`?

wrandall22: The EP upgrade is currently on `give-stage2-next.cru.org` which will go away once we get closer to that project finishing.

wrandall22:
Looks good for testing on `staging`

canac:
Works as expected in staging ✅

wrandall22 (approved):

PR #1006, opened 2023-02-21
[EP-2405] Treat missing cru-profile as not logged in

canac:
**Problem:**

When the user is logged in and then visits a branded checkout page, branded checkout will logout the user. If the branded checkout page is on a *.cru.org subdomain, the logout will clear the `cru-session` and `cru-profile` cookies because they are set on the root domain `cru.org`. However, a logout request to `DELETE https://secure.cru.org/cas/logout`, for example, will leave the `give-session`, `cortex-session`, and `cortex-role` cookies set on `give.cru.org` untouched because a request to `secure.cru.org` can't modify cookies set on `give.cru.org`. After the logout, the user will still have `cortex-role` set but no `cru-profile` anymore.

This PR treats users without a `cru-profile` as not logged in, even if they have a valid Cortex session.

Tested in staging ✅

https://jira.cru.org/browse/EP-2405

### DIFF
--- a/src/common/services/session/session.service.spec.js
+++ b/src/common/services/session/session.service.spec.js
@@ -112,6 +112,7 @@ describe('session service', function () {
     describe('with \'IDENTIFIED\' cortex-session', () => {
       beforeEach(() => {
         $cookies.put(Sessions.role, cortexRole.identified)
+        $cookies.put(Sessions.profile, cruProfile)
         // Force digest so scope session watchers pick up changes.
         $rootScope.$digest()
       })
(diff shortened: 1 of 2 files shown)


src/common/services/session/session.service.spec.js
```diff
@@ -288,6 +295,7 @@ describe('session service', function () {
     describe('with \'IDENTIFIED\' role', () => {
       beforeEach(() => {
         $cookies.put(Sessions.role, cortexRole.identified)
+        $cookies.put(Sessions.profile, cortexRole.registered)
```
wrandall22, line 298: This should be set to `cruProfile` I think, same with the below change.

canac: Good catch!

wrandall22 (approved):

PR #1007, opened 2023-02-22
AEM-395 - Mitigate cache

wrandall22:
This basically adds a `?doneEditing` query parameter to the designation page after the person is done editing it. There is a no-cache rule that looks for this query parameter [here](https://github.com/CruGlobal/cru-aem-commons/pull/223)

### DIFF
--- a/src/app/designationEditor/designationEditor.component.js
+++ b/src/app/designationEditor/designationEditor.component.js
@@ -345,14 +345,16 @@ class DesignationEditorController {
 
   getDoneEditingUrl () {
     const vanityPath = this.designationContent['sling:vanityPath']
+    const cacheBust = '?doneEditing'
+
     if (vanityPath) {
       if (isArray(vanityPath)) {
-        return vanityPath[0].replace('/content/give/us/en', '')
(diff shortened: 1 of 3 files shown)


src/app/designationEditor/designationEditor.component.js
```diff
@@ -341,6 +342,21 @@ class DesignationEditorController {
       this.carouselLoaded = true
     }
   }
+
+  getDoneEditingUrl () {
+    const vanityPath = this.designationContent['sling:vanityPath']
+    const cacheBust = '?doneEditing'
+
+    if (vanityPath) {
+      if (isArray(vanityPath)) {
```
canac, line 351: I believe you could use `Array.isArray` here instead of needing a lodash import.

canac (approved):

src/app/designationEditor/designationEditor.component.js
wrandall22: There also seems to be `angular.isArray` in use in other parts of the project.

wrandall22: I mainly chose lodash for consistency, but the project itself is not consistent.

PR #1008, opened 2023-02-27
EP-2406 and fixing signup error handling for invalid passwords

dr-bizz:
## Description
Fixes [EP-2406](https://jira.cru.org/browse/EP-2406) & enhances error handling for invalid passwords on the signup form.

## Changes
- If recurring gift donations are undefined, return an empty array to prevent an error, allowing the rest of the payment methods to render.
- Created error handling for invalid passwords on the signup form. Increased the password min character count to match API's validation settings.
- Also updated a few errors to have proper punctuation. I need to check that this doesn't cause translation issues.

## Testing
I've tested these changes locally. It's hard to ensure my change fixes the task EP-2406, as whenever I change my payment details, Siebel updates immediately with no errors.


wrandall22:
This needs to be based off of the `ep-upgrade` branch. I would suggest cherry picking.

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -74,12 +75,12 @@
...
              class="form-control form-control-subtle"
              ng-model="$ctrl.password"
-             minlength="8"
+             minlength="15"
              ng-pattern="/^(?=.*\d)(?=.*[a-zA-Z]).{6,}$/"
              required>
       <div ng-if="($ctrl.signUpForm.password | showErrors)" ng-messages="$ctrl.signUpForm.password.$error">
         <span class="help-block" ng-message="required" translate>Password is required.</span>
-        <span class="help-block" ng-message="minlength" translate>Must be 8 or more characters.</span>
+        <span class="help-block" ng-message="minlength" translate>Must be 15 or more characters.</span>
```
wrandall22, line 83: This is not a static rule for all user types

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -36,6 +36,9 @@ class SignUpModalController {
         this.onSuccess(data)
       }, (error) => {
         this.hasError = true
+        if (error?.data?.error === 'invalid_password' && error?.data?.description) {
```
wrandall22, line 39: Note this whole commit will be no longer useful when we move to Okta.

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -74,12 +75,12 @@
              type="password"
              class="form-control form-control-subtle"
              ng-model="$ctrl.password"
-             minlength="8"
+             minlength="15"
```
wrandall22, line 78: This is not a universal rule.

src/common/services/api/profile.service.js
```diff
@@ -240,7 +240,7 @@ class Profile {
           if (paymentMethod.address) {
             paymentMethod.address = formatAddressForTemplate(paymentMethod.address)
           }
-          paymentMethod.recurringGifts = flatMap(paymentMethod.recurringgifts.donations, donation => {
+          paymentMethod.recurringGifts = flatMap(paymentMethod?.recurringgifts?.donations || [], donation => {
```
wrandall22, line 243: This change should be in a separate commit.

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -1,12 +1,13 @@
 <div class="col-md-12 col-xs-12">
   <div class="alert alert-danger" role="alert" ng-messages="$ctrl.signUpErrors" ng-if="$ctrl.hasError">
-    <div ng-message="400" translate>An error occurred during sign up.<br />Please Check your Email Address and Password and try again.</div>
-    <div ng-message="403" translate>An error occurred during sign up.<br />This is most likely caused by the use of a reserved domain name in your email address. You might try to
+    <div ng-message="invalid_password_message">{{$ctrl.signUpErrors.invalid_password_message}}</div>
+    <div ng-message="400" translate>An error occurred during sign up. <br />Please Check your Email Address and Password and try again.</div>
+    <div ng-message="403" translate>An error occurred during sign up. <br />This is most likely caused by the use of a reserved domain name in your email address. You might try to
```
wrandall22, line 5: Is there a reason for these whitespace changes?

src/common/components/signUpModal/signUpModal.tpl.html
dr-bizz: Gotcha, so the universal rule for passwords is that it has to be greater than 8 characters, but we've added an extra layer of security to make it at least 15.

As I continue to edit the give-web site, is there a list of universal rules? So I know what I can and can't change.

wrandall22: These rules are defined within Okta itself, they aren't give site rules. I'm not sure how to find that logic to be honest.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I see in this UI Engineering OKR Tactics & Resource Allocation spreadsheet we have an OKR to make it easy for staff members to log in via OKTA, but what about non-staff members? Will they also use OKTA?

src/common/components/signUpModal/signUpModal.tpl.html
dr-bizz: When the error was rendered, it appeared with no whitespace between the period and the next word. 
```An error occurred during sign up.Please Check your Email Address and Password and try again.```
```This is most likely caused by the use of a reserved domain name in your email address.You might try to...`

src/common/components/signUpModal/signUpModal.component.js
wrandall22: Yep, everyone will use Okta when we make the switch, which will get rid of our custom modal taking username/password and instead redirect the user directly to Okta's sign in form.

dr-bizz:
Because my branch was based on `master` I'm going to close this PR and open a new PR to make things easier.

PR #1009, opened 2023-02-28
adding form to getContacts url

dr-bizz:
Added `/form/` to the getContacts url.


wrandall22 (approved):
Make sure to merge into `ep-upgrade-stage` locally and push that up for testing before merging here.

wrandall22 (changes requested):
Looks like the test needs fixing.

PR #1010, opened 2023-03-03
AEM-568 - Turn full urls into vanity urls for nice hover effect

wrandall22:
When we move to the AEM cloud servers we are going to lose some of our vanity urls when it comes to link creation. The dispatcher (web server) still understands these vanity urls so we can recreate the links in the proper format here.

### DIFF
--- a/src/app/searchResults/searchResults.component.js
+++ b/src/app/searchResults/searchResults.component.js
@@ -91,8 +91,8 @@ class SearchResultsController {
    * @returns {*} the vanity url
    */
   buildVanity (originalPath) {
-    const pattern = /\/designations\/([0-9])\/([0-9])\/([0-9])\/([0-9])\/([0-9])\/([0-9]{7})\.html/g
-    return originalPath.replace(pattern, '/$6')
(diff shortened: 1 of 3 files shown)


src/app/searchResults/searchResults.component.js
```diff
@@ -83,6 +83,17 @@ class SearchResultsController {
...
   }
+
+  /**
+   * Go from a full url such as https://give-stage2.cru.org/designations/0/1/2/3/4/0123456.html
+   * to a vanity url of the pattern https://give-stage2.cru.org/0123456
+   * @param originalPath the full url
+   * @returns {*} the vanity url
+   */
+  buildVanity (originalPath) {
+    const pattern = /\/designations\/([0-9])\/([0-9])\/([0-9])\/([0-9])\/([0-9])\/([0-9]{7})\.html/g
```
canac, line 94: RegEx is ugly no matter what you do, but if you wanted to shorten it, I think you could replace `0-9` with `\d` and `([0-9])\/([0-9])\/([0-9])\/([0-9])\/([0-9])\/` with `(\d\/){5}` and update the match index to `$2`.

canac (approved):

PR #1013, opened 2023-03-15
Record give-session sub in Rollbar

canac:
In an attempt to better debug issues like https://secure.helpscout.net/conversation/2184380957/909478/, record the `give-session` `ssoGuid` in Rollbar to see if it's sometimes missing, causing the 422 error when loading `/bin/crugive/image?designationNumber=XXXXXXX`.


canac:
Tested in staging ✅

wrandall22 (approved):

PR #1015, opened 2023-03-16
Aem cloud

wrandall22:
Updates from aem-cloud testing.

### DIFF
--- a/src/common/services/api/designations.service.js
+++ b/src/common/services/api/designations.service.js
@@ -268,7 +268,8 @@ class DesignationsService {
 
   ministriesList (pagePath) {
     const path = pagePath + '/jcr:content/content-parsys/designation_search_r.json'
-    return Observable.from(this.$http.get(this.$location.origin + '/' + path))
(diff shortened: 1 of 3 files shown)


src/common/services/api/designations.service.js
```diff
@@ -268,7 +268,8 @@ class DesignationsService {
 
   ministriesList (pagePath) {
     const path = pagePath + '/jcr:content/content-parsys/designation_search_r.json'
-    return Observable.from(this.$http.get(this.$location.origin + '/' + path))
+    const url = `${this.$location.protocol()}://${this.$location.host()}/${path}`
```
canac, line 271: If the port will ever be different from 80, `$location.host()` doesn't include the port.

canac (approved):

src/common/services/api/designations.service.js
wrandall22: Yeah, I was debating on including port. It will only matter on local machines, which only I have a full local environment set up with a locally running AEM and front-end web at this point.

PR #1018, opened 2023-03-23
AEM-656 - Add cloud hosts to the cookies add check

wrandall22:
This controls whether or not we keep track of the number of items in the mini cart in the navigation.

[Jira](https://jira.cru.org/browse/AEM-656)


dr-bizz (approved):
Tested on staging and code looks solid.

PR #1019, opened 2023-03-23
Improve UI for password check

wrandall22:
Enforce 15 character passwords at the UI layer since it seems universal now.


dr-bizz (commented):
The code looks good, I tested creating an account, but I couldn't test resetting the password functionality. On local, I was met with an error when entering my email while resetting my password (screenshot below). I assume this is just a local issue. The reset password requirements code is the same as the signup, so it looks good.

This isn't required, but I think it's good to add tests to ensure the form doesn't submit when the password requirements aren't met.

Screenshot 2023-03-24 at 8 49 26 AM

wrandall22:
@dr-bizz if this is ready to go, can you approve it please?

dr-bizz (approved):

PR #1023, opened 2023-03-28
[EP-2314] Image upload improvements

canac:
Adds several improvements to the image upload flow.

* Locally caches uploaded images so that they can be displayed while AEM is still replicating them
* Updates the carousel after the selected images change

When testing, inspect the `` tag of newly-uploaded images in the upload modal and on the page. The `src` should start with `blob:` instead of being a normal URL. When you manually refresh the page all the images should be normal URLs.

Tested in staging ✅

### DIFF
--- a/src/app/designationEditor/photoModal/photo.modal.js
+++ b/src/app/designationEditor/photoModal/photo.modal.js
@@ -22,7 +22,10 @@ class ModalInstanceCtrl {
     this.retryService = retryService
 
     this.imgDomain = envService.read('imgDomain')
-    this.numProcessingPhotos = 0
+    // The key is the File, and the value is the blob URL
+    // This can't just be a set because URL.createObjectURL(file) creates a new blob URL every
+    // time, even when called with the same file
+    this.processingPhotos = new Map()
     this.RETRY_DELAY_MS = 3000
     this.MAX_RETRIES = 20
 
@@ -35,20 +38,30 @@ class ModalInstanceCtrl {
     this.maxNumberOfPhotos = 6
   }
 
-  uploadComplete (response) {
+  uploadStart (file) {
+    this.uploading = true
+
+    this.processingPhotos.set(file, URL.createObjectURL(file))
+  }
+
+  uploadComplete (response, file) {
     this.uploading = false
-    ++this.numProcessingPhotos
 
     this.refreshPhotos(new URL(response.headers('location'), this.$window.location).pathname).then((photos) => {
       this.photos = photos
     }, angular.noop).finally(() => {
-      --this.numProcessingPhotos
+      this.processingPhotos.delete(file)
     })
   }
 
-  // ng-repeat needs an array to iterate over, so generate an array with one element per processing photo
+  uploadError (file) {
+    this.uploading = false
+    this.processingPhotos.delete(file)
+  }
+
+  // ng-repeat can only iterate over arrays, not Maps, generate an array of the blob URLs of all currently uploading/processing images
   getProcessingPhotos () {
-    return new Array(this.numProcessingPhotos).fill(undefined)
+    return Array.from(this.processingPhotos.values())
   }
 
   // Attempt to load a recently uploaded photo because newly uploaded photos will 404 for several seconds
--- a/src/app/designationEditor/photoModal/photo.modal.js
+++ b/src/app/designationEditor/photoModal/photo.modal.js
@@ -1,15 +1,30 @@
 import angular from 'angular'
+import 'angular-environment'
 
+import loading from 'common/components/loading/loading.component'
 import imageUploadDirective from 'common/directives/imageUpload.directive'
 import designationEditorService from 'common/services/api/designationEditor.service'
+import imageCacheService from 'common/services/imageCache.service'
+import retryService from 'common/services/retry.service'
 
 const controllerName = 'photoCtrl'
 
 class ModalInstanceCtrl {
   /* @ngInject */
-  constructor ($timeout, designationNumber, campaignPage, photos, photoLocation, selectedPhoto, designationEditorService) {
+  constructor ($timeout, $interval, $http, $q, $window, designationEditorService, envService, imageCacheService, retryService, designationNumber, campaignPage, photos, photoLocation, selectedPhoto) {
     this.$timeout = $timeout
+    this.$interval = $interval
+    this.$http = $http
+    this.$q = $q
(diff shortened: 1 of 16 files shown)


src/common/services/imageCache.service.js
```diff
@@ -0,0 +1,43 @@
...
+    return this.envService.read('imgDomainDesignation') + url
+  }
+
+  // Lookup an image pathname and return its blob URL if it is cached or its absolute URL otherwise
+  get (url) {
+    return this.blobCache.get(url) || this.makeAbsoluteUrl(url)
+  }
+
+  // Cache an image by its pathname and return a promise that resolves to its blob URL
+  cache (url) {
```
wrandall22, line 28: The URL which is created is in the `location` header of the POST response for the image. It may be easier/faster to get it from there? Though, it won't have the image itself, which it looks like maybe we're getting from here.

canac: You're right that it's in the `Location` header. Here's where we extract that header to get the image URL. https://github.com/CruGlobal/give-web/blob/a61a9dd93819cf0fd77984c3f22abdfc1dbf945e/src/app/designationEditor/photoModal/photo.modal.js#L42 `imageCacheService` needs to cache all the variants of the image (cover, thumbnail, etc.), so just knowing the primary image URL isn't enough to start trying to load the cached images without reimplementing the naming logic on the client (i.e. `${imageUrl}.transform/GiveDesignationCoverPhoto/img.jpg`). We could try it that way, but I'm not sure that it will be faster because the slowdown is coming from getting lucky and requesting from a server that has replicated the image.

src/app/designationEditor/photoModal/photo.modal.js
```diff
@@ -20,14 +35,52 @@ class ModalInstanceCtrl {
...
-      this.designationEditorService.getPhotos(this.designationNumber).then((response) => {
-        this.photos = response.data
-        this.uploading = false
-      }, angular.noop)
-    }, 3500)
+  uploadComplete (response) {
+    this.uploading = false
+    ++this.numProcessingPhotos
+
+    this.refreshPhotos(new URL(response.headers('location'), this.$window.location).pathname).then((photos) => {
```
wrandall22, line 42: Don't we want just the protocol and domain for the second parameter (base)?

canac: Either way works. [MDN has examples](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#examples) with full URLs as the base. But I could change it to `this.$window.location.origin` if you want me to.

wrandall22: Oh, I see one now: `new URL("/en-US/docs", "https://developer.mozilla.org/fr-FR/toto");`

src/common/app.config.js
```diff
@@ -38,7 +38,7 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
       development: {
         apiUrl: 'https://give-stage2.cru.org',
         imgDomain: '',
-        imgDomainDesignation: 'https://give-stage2.cru.org',
+        imgDomainDesignation: 'https://localhost.cru.org:9000',
```
wrandall22, line 41: Did you mean to update this? This would require you to run a local AEM (which would actually be at `http://localhost:4502` by default)

canac: I did my testing on `https://localhost.cru.org:9000`, and since the code is now fetching the images, there are CORS errors if they're on a different origin. The localhost:9000 dev server proxies requests to `give-stage2`, which is why this works: https://github.com/CruGlobal/give-web/blob/2109c3c80116b6c7456e8a73b2c851630270f7c5/webpack.config.js#L143-L160

wrandall22:
Overall I think the code is looking good; I still want to test a bit next week after cloud migration stuff dies down. For future PRs that create new files, we should consider using react2angular (which is already setup in this repo) to see if we can make new code React/Typescript (may not always be possible).

wrandall22:
If the retry delay has made it to staging, it isn't working properly. The retries are still instantaneous there.

canac:
> If the retry delay has made it to staging

@wrandall22 Sorry, I forgot to update `staging` with the delay. It's building now.

wrandall22:
Looks like the deploy happened, but the delay is still not happening.

wrandall22:
So, I see the delay happening now. One thing I thought the cache was supposed to do was to show the image while the real one was loading, is that not the case? If it is supposed to be doing that, it is not doing that. Something else I've noticed while testing is that the transformed image tends to always come back, but the non-transformed image gets the 404s.

canac:
@wrandall22 I modified the image uploader to display the uploaded image immediately, cropped and centered. This required some extra styles that I assume will need to be added to AEM when this lands.

For some reason, I still can't reproduce the 404 errors for recently-uploaded images.

src/app/designationEditor/photoModal/photoModal.spec.js
```diff
@@ -53,19 +64,28 @@ describe('Designation Editor Photo', function () {
     }]
     jest.spyOn($ctrl, 'refreshPhotos').mockReturnValueOnce($q.resolve(photos))
 
-    $ctrl.uploadComplete({ headers: () => '/content/photo3.jpg' })
-    expect($ctrl.numProcessingPhotos).toBe(1)
-
+    $ctrl.uploadStart(file)
+    $ctrl.uploadComplete({ headers: () => '/content/photo3.jpg' }, file)
```
wrandall22, line 68: Should there still be a check before this line to see there is 1 in the map?

wrandall22: I guess the `uploadStart` test handles that.

wrandall22:
I haven't seen the case yet where the file finally resolves. Does the blobUrl get replaced with the real url if it is ever found?

canac:
> I haven't seen the case yet where the file finally resolves. Does the blobUrl get replaced with the real url if it is ever found?

You're correct that it uses the blob URL for the lifetime of the page. If the user chooses one of the images the real URL is sent to the server. The blob URL is still displayed because we have no way of knowing whether it will 404 the next time we try to load the image via its real URL.

However, if we can't reproduce the intermittent 404s anymore since the cloud migration, we could potentially remove all of that caching logic and just keep the other upload improvements in this PR.

wrandall22 (approved):

PR #1024, opened 2023-03-29
Add note to image upload modals

wrandall22:
Add a note to each of the image modals to refresh the page if something does not show up right away.


canac (approved):
#1023 should help with this, but your changes look good.

wrandall22:
Yeah, I agree. I just want to do more thorough testing and review on that one before approving it.

PR #1027, opened 2023-03-30
Remove prod-cloud environment

wrandall22:
This is no longer needed, and may be giving us bad redirects.


canac (approved):

wrandall22:
So, it doesn't break the redirects. I might leave this here until we remove that DNS.

PR #1031, opened 2023-04-04
[EP-2410] Branded checkout improvements

canac:
Changes:

* Log error when tokenization or encryption fails
* Scroll to the section of the branded checkout form that matches the "Change" button the user clicked
* Send review_order event after add_payment_info

https://jira.cru.org/browse/EP-2410

### DIFF
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -97,7 +97,7 @@ class BrandedCheckoutController {
           scrollElement = 'product-config-form'
           break
         case 'payment':
-          scrollElement = 'checkout-existing-payment-methods'
+          scrollElement = 'checkout-step-2'
           break
       }
     }
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -79,11 +79,11 @@ class BrandedCheckoutController {
         this.checkoutStep = 'thankYou'
         break
     }
-    this.$element.querySelector('branded-checkout').scrollIntoView({ behavior: 'smooth' })
+    this.$element.scrollIntoView({ behavior: 'smooth' })
   }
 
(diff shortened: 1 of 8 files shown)


src/app/branded/branded-checkout.component.js
```diff
@@ -78,17 +79,43 @@ class BrandedCheckoutController {
...
+          break
+        case 'payment':
+          scrollElement = 'checkout-existing-payment-methods'
+          break
+      }
     }
-    this.$window.document.querySelector('branded-checkout').scrollIntoView({ behavior: 'smooth' })
+
+    // Watch for changes until the element we are scrolling to exists and everything has loaded
+    // because there will be layout shift every time a new component finishes loading
```
wrandall22, line 106: Hmm, my experience has always been that the loading indicator remains until all of the pieces have finished loading. Did you run into other cases in your testing?

src/app/branded/branded-checkout.component.js
```diff
@@ -78,17 +79,43 @@ class BrandedCheckoutController {
...
+    const observer = new this.$window.MutationObserver(() => {
+      // TODO: When Firefox supports :has(), this query can be changed to `.panel :has(${scrollElement})`
```
wrandall22, line 108: Note that give site must support older versions of browsers also, so we need to be careful using anything that is not supported by like 99% of browsers (we had to support IE 8 up until 2016, for example).

canac: Good point. I updated the comment.

wrandall22 (approved):

canac:
@wrandall22 FYI I'm doing some more testing in staging before I merge this and might have a few more tweaks.

canac:
Tested in staging ✅

wrandall22 (approved):

PR #1035, opened 2023-05-09
Update readme new alb

wrandall22:
See https://github.com/CruGlobal/cru-terraform/pull/4116 for the origin of this PR.


wrandall22:
Let us know if that test succeeds, then I'll update the ALB references to the VIP url instead.

PR #1037, opened 2023-05-17
EP-2354 - Enable auto-formatting of credit card numbers as users type them in.

wrandall22:
This is part of a larger group of work that an intern did last summer. This piece has been approved to go forward by development services.

### DIFF
--- a/src/common/directives/creditCardNumber.directive.js
+++ b/src/common/directives/creditCardNumber.directive.js
@@ -1,0 +1,66 @@
+import angular from 'angular'
+import appConfig from 'common/app.config'
+import cruPayments from 'cru-payments'
+
+const directiveName = 'creditCardNumber'
+
+const creditCardNumberDirective = /* @ngInject */ function () {
+  return {
+    restrict: 'A',
+    require: '?ngModel',
+    link: function (scope, element, attrs, ngModelCtrl) {
+      if (!ngModelCtrl) {
+        return
+      }
+
+      ngModelCtrl.$parsers.unshift(function (creditCardNumber) {
+        let trimmedCardNum = creditCardNumber.replace(/\D/g, '')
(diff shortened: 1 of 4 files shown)


src/common/directives/creditCardNumber.directive.js
```diff
@@ -0,0 +1,83 @@
...
+        let trimmedCardNum = creditCardNumber.replace(/\s+/g, '').replace(/\D/gi, '')
+        const cardType = cruPayments.creditCard.card.info.type(creditCardNumber)
+        const numbers = []
+
+        if (cardType === 'Visa' || cardType === 'MasterCard' || cardType === 'Discover') {
+          if (trimmedCardNum.length > 16) {
+            trimmedCardNum = trimmedCardNum.substr(0, 16)
+          }
+          const partitions = [4, 4, 4, 4]
+          let position = 0
```
canac, line 26: It could be nice to split this logic out into a separate function since it is used three times. We could even add tests for that method since it's not trivial logic.

src/common/directives/creditCardNumber.directive.js
```diff
@@ -0,0 +1,83 @@
...
+  return {
+    restrict: 'A',
+    require: '?ngModel',
+    link: function (scope, element, attrs, ngModelCtrl) {
+      if (!ngModelCtrl) {
+        return
+      }
+
+      ngModelCtrl.$parsers.unshift(function (creditCardNumber) {
+        let trimmedCardNum = creditCardNumber.replace(/\s+/g, '').replace(/\D/gi, '')
```
canac, line 17: I believe only `creditCardNumber.replace(/\D/g, '')` is necessary to strip out all non-digits.

canac (approved):

PR #1043, opened 2023-06-30
EP-2402 - Use right form

wrandall22:
We were using the out of the box payment instrument creation form, which did not send the payment instrument to Siebel. [Companion PR](https://github.com/CruGlobal/give-ep-commerce/pull/151)


dr-bizz (approved):
The code looks good. I'm unable to test on local or staging as it doesn't allow me to log in, returning a 504 error or returning a 307 response when requesting profile info.

As I'm unable to test, I don't know if this is an issue. But does this file also need to be adjusted for `selfservicepaymentinstruments`? 
```
src/common/services/api/fixtures/cortex-purchase.fixture.js
```

wrandall22:
Nope, the purchase area is not self-service, so does not use that resource.

PR #1045, opened 2023-07-18
EP-2431 - Fix duplicate phone display

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2431)

The problem was that in certain circumstances (maybe involving cookies), the login box would come up again (`$onInit` called) but the constructor is not called. This leaves the prior loaded phone numbers in the array, then we just push to it again. The fix here just empties the array right before loading the phone numbers from the API. It does not seem to cause issues with adding phone numbers.


wrandall22:
Tests well in stage.

PR #1049, opened 2023-08-21
EP-2439 - Fix name swap

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2439)

When logging in from the top navigation `Sign In` link, the `Sign In` wording changes to the user's name, but when logging in via the self-service page navigation, it does not work. This is because the AEM side is listening for the `giveloaded` event, which is not fired when login is successful. It fires before the login, but not after. This PR is adding an event that AEM will be listening for to do the same wording swap. Strictly speaking, I'm not sure if/when the sessionModal component code is called, it wasn't in my testing. If we want, I can remove that part of this PR.

### DIFF
--- a/src/common/components/signInForm/signInForm.component.js
+++ b/src/common/components/signInForm/signInForm.component.js
@@ -10,8 +10,10 @@ const componentName = 'signInForm'
 
 class SignInFormController {
   /* @ngInject */
-  constructor ($log, sessionService, gettext) {
+  constructor ($log, $document, sessionService, gettext) {
     this.$log = $log
+    this.$document = $document
+    this.$injector = angular.injector()
     this.sessionService = sessionService
     this.gettext = gettext
   }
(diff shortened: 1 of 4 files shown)


src/common/components/signInForm/signInForm.component.js
```diff
@@ -31,6 +33,12 @@ class SignInFormController {
     this.sessionService
       .signIn(this.username, this.password, this.mfa_token, this.trust_device, this.lastPurchaseId)
       .subscribe(() => {
+        const $injector = this.$injector
+        if (!$injector.has('sessionService')) {
+          $injector.loadNewModules(['sessionService'])
```
dr-bizz, line 38: He mentioned here that you don't need to check if it hasn't got the service as it won't add it if it's there. But I guess you knew that already and just wanted to make it more performant.

https://docs.angularjs.org/api/auto/service/$injector#loadNewModules

dr-bizz (approved):

PR #1052, opened 2023-09-08
Filter suggestedAmounts to remove zero amounts or empty objects

dr-bizz:
## Description
Ensuring we do not show zero amounts on giving modal and we show the detail amounts if all suggested amounts are zero.

## Changes
- When saving suggested amounts, filter out zero and empty objects.
- Removing zero amounts on initial load.
- Filer out zero amounts when rendering amounts on a gift modal.


wrandall22:
Sorry, I was looking at the wrong file in dev tools. It is there.

wrandall22 (approved):

PR #1053, opened 2023-09-12
Add DataDog RUM

caleballdrin:
## Description
Add datadog RUM to give-web.
https://jira.cru.org/browse/EP-2441


.env
wrandall22, line 1: I don't think this file should be committed in here. These values (if meant to be public information) would normally go into app.config.js instead. Is datadog.config.js loaded outside the normal load process and skipping app.config.js somehow?

src/common/datadog.config.js
```diff
@@ -0,0 +1,18 @@
+import { datadogRum } from '@datadog/browser-rum';
+
+datadogRum.init({
+    applicationId: '3937053e-386b-4b5b-ab4a-c83217d2f953',
```
wrandall22, line 4: Will this be the same for all environments?

src/common/datadog.config.js
```diff
@@ -0,0 +1,18 @@
...
+    clientToken: process.env.DATADOG_CLIENT_TOKEN,
+    site: 'datadoghq.com',
+    service:'give-web',
+    env: process.env.NODE_ENV,
+    version: process.env.GITHUB_SHA,
+    sessionSampleRate:100,
+    sessionReplaySampleRate: 20,
```
wrandall22, line 11: @mattdrees do you have an opinion on the values of these rates to start off with? This would be 100% of sessions are recorded and 20% of those are set up for the session replay feature.

wrandall22 (dismissed):

src/common/datadog.config.js
caleballdrin: These were the defaults that DataDog suggested. And it is what MPDx React is using.

src/common/datadog.config.js
caleballdrin: I think so

.env
caleballdrin: Yeah, I shouldn't have included the .env. I will remove that.

caleballdrin: @dr-bizz suggested that I add the DATADOG_CLIENT_KEY to terraform. So I will make a PR there.
I created datadog.config.js because I saw that was how it was done with rollbar.config.js.

src/common/datadog.config.js
mattdrees: My opinion would be to start with both of those at 100 for nonproduction. And for production, to start with 10% for sessionSampleRate. For `sessionReplaySampleRate` in prod, I think I'd go with 10% (which means 1% of all users). 

I think the goal here would be to keep browser rum costs for givesite below about $100 / month. So start with those settings (or something similar), and then adjust a month later once we see how costs turn out.

src/common/datadog.config.js
```diff
@@ -0,0 +1,18 @@
...
+
+datadogRum.init({
+    applicationId: '3937053e-386b-4b5b-ab4a-c83217d2f953',
+    clientToken: process.env.DATADOG_CLIENT_TOKEN,
+    site: 'datadoghq.com',
+    service:'give-web',
+    env: process.env.NODE_ENV,
+    version: process.env.GITHUB_SHA,
+    sessionSampleRate: 10,
+    sessionReplaySampleRate: 1,
```
wrandall22, line 11: As per Matt's comment, you'll want these to be environment-specific.

dr-bizz (changes requested):
Can you check that Datadog is being added to the application? I don't think it is as you haven't made any changes to import DataDog into the main Angular file.

A good way to check is by console logging or adding an item to the window object, storing data in local storage. Just something which shows this file is being loaded into each page.

wrandall22:
> We should only have Datadog on Production unless someone has specified we need it on staging too.

We definitely want it on staging also.

dr-bizz:
> > We should only have Datadog on Production unless someone has specified we need it on staging too.
> 
> We definitely want it on staging also.

You probably will need to set up 2 RUM accounts, one for Prod and one for Staging, as we don't want Staging alerts to go to Sev 1

wrandall22:
I don't think RUM will connect to Sev1; not sure if RUM has the built-in idea of environments like the rest of Datadog does or not.

dr-bizz:
> I don't think RUM will connect to Sev1; not sure if RUM has the built-in idea of environments like the rest of Datadog does or not.

Sorry, you are correct. for some reason, I was thinking of synthetic tests. we will only need one then as you can pass in an `env`, which you can filter on DataDog

src/common/app.config.js
```diff
@@ -535,6 +536,23 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
     PREFERRED_RADIO_STATION: 'Preferred Radio Station: '
   })
   $translateProvider.preferredLanguage('en')
+
+  window.datadogRum && window.datadogRum.init({
```
wrandall22, line 540: How is `datadogRum` getting into `window`?

src/common/app.config.js
```diff
@@ -535,6 +536,24 @@ const appConfig = /* @ngInject */ function (envServiceProvider, $compileProvider
     PREFERRED_RADIO_STATION: 'Preferred Radio Station: '
   })
   $translateProvider.preferredLanguage('en')
+
+  window.datadogRum = datadogRum
+  window.datadogRum && window.datadogRum.init({
+    applicationId: '3937053e-386b-4b5b-ab4a-c83217d2f953',
+    clientToken: process.env.DATADOG_RUM_CLIENT_TOKEN,
```
caleballdrin, line 543: @dr-bizz Is this still the correct way to get the client token now that it is on terraform?

dr-bizz: I think you also need to add the variable here: webpack.config.js on line 58. So the front end can read it.

```
DATADOG_RUM_CLIENT_TOKEN: process.env. DATADOG_RUM_CLIENT_TOKEN || ''

Comment (wrandall22) [reply src/common/app.config.js:543]: You'll also need it in `build.yml` for GHA I think.

**Comment by dr-bizz** [comment]
I would separate your datadog file from the app.config.js. This will also give you access to Angular's window object

datadog.config.js
```
import 'angular-environment'
import { datadogRum } from '@datadog/browser-rum'

export default dataDogConfig = /* @ngInject */ function (envServiceProvider, $window) {
  const config = {
    applicationId: 'XXXXXXXXXXXXXXXXXXXXXXXX',
    clientToken: process.env.DATADOG_RUM_CLIENT_TOKEN,
    site: 'datadoghq.com',
    service: 'give-web',
    version: process.env.GITHUB_SHA,
    sessionSampleRate: envServiceProvider.is('development') ? 100 : 10,
    sessionReplaySampleRate: envServiceProvider.is('development') ? 100 : 1,
    trackUserInteractions: true,
    trackResources: true,
    trackLongTasks: true,
    defaultPrivacyLevel: 'mask-user-input'
  }

  $window.datadogRum = datadogRum

  
  $window.datadogRum && $window.datadogRum.init(config)

  $window.datadogRum && $window.datadogRum.startSessionReplayRecording()
}
```

Then, on the app.config.js
```
import dataDogConfig from './datadog.config.js'
...
.config(appConfig)
.config(rollbarConfig)
.config(dataDogConfig)
```

wrandall22:
Looking good; let's wait Monday to deploy.

dr-bizz (approved):
Looks good to me. Great work! 
Before you merge this into Prod, ensure you have pushed the Terraform PR so the`DATADOG_RUM_CLIENT_TOKEN` is an env on PROD.

PR #1060, opened 2023-10-05
EP-2412 - Force Logged-in During Save

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2412)
[Rollbar](https://app.rollbar.com/a/Cru/fix/item/give-web/64805)

Currently, the system allows you to navigate around and attempt to save changes on the designation editor even if your `give-session` cookie has expired (sometimes...). This PR forces a login if needed on save. One side effect I noticed in my testing is that is sends 2 POST requests instead of 1. I'm not sure why at this point, but I think it has to do with how the `sessionEnforcerService` works. In my automated tests, calling the `sign-in` function happened twice as well. That will be out of scope of this PR to improve.

### DIFF
--- a/src/app/designationEditor/designationEditor.component.js
+++ b/src/app/designationEditor/designationEditor.component.js
@@ -303,18 +303,26 @@ class DesignationEditorController {
     this.loadingOverlay = true
     this.saveDesignationError = false
 
-    return this.designationEditorService.save(this.designationContent, this.designationNumber, this.campaignPage).then(() => {
-      this.saveStatus = 'success'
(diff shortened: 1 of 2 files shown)


wrandall22:
To test, I've been following these steps:

1. Login to my profile
2. Go to my designation page
3. Click Edit page
4. Delete the `give-session` cookie via developer tools
5. Edit the text
6. Save

I also find that I need to refresh the page to see my edits (a problem for another day)

caleballdrin (approved):

PR #1062, opened 2023-10-18
Preventing DD from producing errors when running tests

dr-bizz:
## Description
When running tests, DataDog would produce an error/warning after every test as I didn't have the env var `DATADOG_RUM_CLIENT_TOKEN` on my local.
Added tests for DD.

## Changes
- Ensured DD is only initialised if env var is present.
- Creates 2 tests for DD, one without the Var and one with.
- Added new script for debugging tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "build": "webpack -p",
     "build:analyze": "webpack -p --env.analyze",
     "test": "jest",
+    "test:log": "jest --silent=false",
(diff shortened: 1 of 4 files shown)


wrandall22 (approved):

dr-bizz:
@caleballdrin Just a fyi.  This is one of the ways we could have added tests to the dataDog config.

PR #1065, opened 2023-10-25
Strip amount $

wrandall22:
This PR strips out `$` from the gift amount in the product config form.


dr-bizz (approved):
JS looks good. I don't see it on stage to test, but I'll approve it for now and test locally

wrandall22:
Can only test on the parallel stage environment at the moment. It is unfortunately still not right due to how `parseFloat` handles commas.

dr-bizz:
I see

wrandall22:
The mix of this and the other branch (better error message) are available to test on https://give-stage2-next.cru.org/

hlbraddock (approved):

PR #1067, opened 2023-10-25
Better error for bad amount

wrandall22:
Show a nicer error message if someone puts in an invalid amount. This won't be useful until https://github.com/CruGlobal/give-ep-commerce/pull/194 goes live.

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -339,7 +339,7 @@ class ProductConfigFormController {
       if (includes(error.data, 'already in the cart')) {
         this.errorAlreadyInCart = true
         this.onStateChange({ state: 'errorAlreadyInCart' })
-      } else if (error.data.messages[0] && error.data.messages[0].id === 'field.invalid.decimal.format') {
+      } else if (error.data && error.data.messages && error.data.messages[0] && error.data.messages[0].id === 'field.invalid.decimal.format') {
         this.amountFormatError = error.data.messages[0]['debug-message']
         this.onStateChange({ state: 'errorSubmitting' })
       } else if (includes(error.data, 'decimal number')) {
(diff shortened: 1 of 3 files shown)


src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -340,6 +340,12 @@ class ProductConfigFormController {
       if (includes(error.data, 'already in the cart')) {
         this.errorAlreadyInCart = true
         this.onStateChange({ state: 'errorAlreadyInCart' })
+      } else if (error.data && error.data.messages && error.data.messages[0] && error.data.messages[0].id === 'field.invalid.decimal.format') {
+        this.amountFormatError = error.data.messages[0]['debug-message']
```
dr-bizz, line 344: Do you need to restart the `this.amountFormatError = ''` at the start of the `saveGiftToCart()`?

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -340,6 +340,12 @@ class ProductConfigFormController {
       if (includes(error.data, 'already in the cart')) {
         this.errorAlreadyInCart = true
         this.onStateChange({ state: 'errorAlreadyInCart' })
+      } else if (error.data && error.data.messages && error.data.messages[0] && error.data.messages[0].id === 'field.invalid.decimal.format') {
```
dr-bizz, line 343: As you're making `amount` apart of  `this.itemConfigForm` on the PR https://github.com/CruGlobal/give-web/pull/1068.  I don't think it will make it down to this part as the function would have returned on line 310.

wrandall22: Yeah, it would be hard to get down here, it is more of a failsafe. What I did notice when testing just now is that, if it returns on line 310, it gives no feedback to the user. I'll see if I can remedy that as part of this PR.

PR #1068, opened 2023-10-26
Fix amount handling again

wrandall22:
This is what I should have done instead of #1065.

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -197,8 +197,8 @@ class ProductConfigFormController {
   }
 
   waitForFormInitialization () {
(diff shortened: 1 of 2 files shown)


dr-bizz (commented):
I assume `itemConfigForm.amount` is the `AMOUNT` as as a float?

wrandall22:
No, it is an angular form field which we are adding validators to.

dr-bizz:
This might need to be a seperate PR, but on the `saveGiftToCart()` If you enter an invalid amount, can we show an error?

dr-bizz (commented):
This might need to be a separate PR, but on the `saveGiftToCart()` If you enter an invalid amount, can we show an error?

wrandall22:
An invalid amount will be caught by the server and they'll give us back a 400 response, which should be handled nicely in #1067

dr-bizz:
Shoot I missed that PR

dr-bizz (approved):

PR #1069, opened 2023-10-27
itemConfig should be using AMOUNT, not amount

wrandall22:
Another place where the wrong case for `AMOUNT` is used. Need to do a thorough audit of this.


wrandall22:
https://give-stage2.cru.org/branded-checkout.html

wrandall22:
Going to merge as emergency release, you can retro-review. Same with #1067.

PR #1070, opened 2023-10-27
HS-1040983 - Fix amount mapping on branded checkout

wrandall22:
[HS ticket](https://secure.helpscout.net/conversation/2403286620/1040983/)

This PR fixes mapping to suggested amounts on branded checkout.


wrandall22:
https://give-stage2.cru.org/branded-checkout.html

PR #1071, opened 2023-10-30
Post upgrade go live

wrandall22:
Now that EP upgrade and AEM cloud have gone live, we can remove parallel environments.


canac (changes requested):
I think there's a few more to remove at https://github.com/CruGlobal/give-web/blob/a05902a8d7e92099e51cbb787fac657a988148fd/src/common/services/api/cart.service.js#L44

`productViewDetailsEvent` and `ministriesList` tests also use `give-stage-cloud.cru.org` URLs in their mocks, so it might be good to change those too to avoid confusion.

canac (approved):

PR #1072, opened 2023-11-03
Analytics fix EP upgrade

wrandall22:
I know you warned me, but I still missed some things in the analytics file.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -138,7 +138,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
               donationFrequency: frequency,
               siebel: {
                 productType: 'designation',
-                campaignCode: item.config['campaign-code']
+                campaignCode: item.config.CAMPAIGN_CODE
               }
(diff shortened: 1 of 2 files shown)


dr-bizz:
I'm also seeing that on the branded Checkout we have `.amount` instead of `.AMOUNT`, I guess that needs ti change too?

wrandall22:
No, the `item` and the `config` are not the same objects. I went through branded checkout in the debugger to make sure. It basically ignores the config entirely.

dr-bizz:
I think we might also need to update this page:
`src/common/services/api/designations.service.js` - Line 207 & 217
Screenshot 2023-11-03 at 9 39 15 AM

wrandall22:
`campaign-page` is not in the config per se (e.g. it is not stored in ElasticPath) so the case is correct there. Suggested amounts are managed in AEM. I'll look closer at the amount there, but I think it doesn't need to change.

dr-bizz (approved):

wrandall22:
Yeah, the suggested amounts piece is correct.

PR #1073, opened 2023-11-08
Only expose giving links to national staff

canac:
Only allow national staff to edit their giving links.

To test, overwrite the `designationType` of the response for the `GET /bin/crugive/designation.html?designationNumber=0864190` request in DevTools to `"National Staff"` and anything else.

@wrandall22 Let me know if you know of a cleaner way to inject `designationType === 'National Staff'` into that one test case. I tried a couple of approaches that didn't work.

https://secure.helpscout.net/conversation/2406630583/1042488/

### DIFF
--- a/src/app/designationEditor/personalOptionsModal/personalOptions.spec.js
+++ b/src/app/designationEditor/personalOptionsModal/personalOptions.spec.js
@@ -4,9 +4,9 @@ import module from './personalOptions.modal'
 
 describe('Designation Editor Personal Options', function () {
   beforeEach(angular.mock.module(module.name))
-  var $ctrl
+  let $ctrl
 
-  var controllerInjections = {
+  const controllerInjections = {
     designationNumber: '000555',
     designationType: 'Staff',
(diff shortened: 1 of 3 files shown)


canac:
Tested in staging ✅

src/app/designationEditor/personalOptionsModal/personalOptions.spec.js
```diff
@@ -6,19 +6,23 @@ describe('Designation Editor Personal Options', function () {
   beforeEach(angular.mock.module(module.name))
   var $ctrl
 
+  var controllerInjections = {
```
wrandall22, line 9: `const`?

wrandall22 (changes requested):
The other way I could see it working would be to make an init() function which the controller calls (and can then be called by the test) where it does some of that logic. In this case you could just set $ctrl.designationType = 'National Staff' in the test case and call init(). But, not sure how much cleaner that is.

wrandall22 (approved):

PR #1074, opened 2023-11-10
EP-2467 - Add a temporary message to the success of account verification

wrandall22:
There are issues with syncing billing addresses during the account verification flow. This prevents loading payment methods altogether, but should work if they logout and login again. This gives a message to that effect.


canac (approved):

wrandall22:
Going to close this for now. The fix should be getting deployed today.

PR #1078, opened 2023-12-05
[HS-1063275] Fix long opportunity names

canac:
Fix long opportunity names displaying under the thumbnail image by switching from floating to flexbox.

Before:
Screenshot 2023-12-05 at 11 16 52 AM

After:
Screenshot 2023-12-05 at 11 16 36 AM

https://secure.helpscout.net/conversation/2443333491/1063275?folderId=7296729

### DIFF
--- a/src/app/searchResults/searchResults.tpl.html
+++ b/src/app/searchResults/searchResults.tpl.html
@@ -123,7 +123,7 @@
                     </h4>
                   </div>
 
-                  <div class="is-row clearfix" ng-repeat-end ng-repeat="r in hits">
+                  <div class="is-row" ng-repeat-end ng-repeat="r in hits">
                     <a ng-href="{{$ctrl.buildVanity(r.path)}}" class="is-row-thumb">
                       <img desig-src="{{r.designationNumber}}" campaign-page="{{r.campaignPage}}">
                     </a>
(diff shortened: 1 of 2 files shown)


j2trumpet:
@canac Can I see this in action somewhere? I want to check the responsiveness.

canac:
@j2trumpet https://give-stage2.cru.org/search-results.html?type=featured You'll have to manually change the text of the "Testing UAT Campaign Page Population" element in DevTools so that it's longer.

Here's the production page too: https://give.cru.org/search-results.html?type=featured

wrandall22:
Seems like we should push back on them creating such long names in the first place.

j2trumpet (changes requested):
I would get rid of the `margin-right: $gutter*.5;` on line 291, add `align-items: center` and `column-gap: $gutter*.5` to `.is-row`, and remove the "clearfix" class from the HTML output on every `.is-row`.

j2trumpet:
@canac My suggested changes will vertically center the three pieces of content with each other and add space between the text and the buttons. We could have just added margin like the image had, but that's not the flex way to do it.

Screen Shot 2023-12-05 at 10 19 28 AM

j2trumpet (approved):
Staging looks good to me!

dr-bizz (approved):
Looks great. Should we go live with this in the code freeze, or wait until Jan?

PR #1080, opened 2023-12-11
2362 okta

dr-bizz:
## Description
Replacing our login/logout functionality with Okta.


wrandall22:
Doesn't seem to like `import { afterEach } from 'node:test'`

src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -388,6 +389,11 @@ class ProductConfigFormController {
       this.$window.location = url
     }
   }
+
+  shouldShowForcedUserToLogoutError() {
```
wrandall22, line 393: Tell me more about this new function

dr-bizz: On adding an item to the cart, we check to see if the user has a `public` role, is logged into Okta and has no items in their account. If so, we automatically force the user to logout.

This was happening before, but when it was logging out, it didn't require a refresh. Okta requires a refresh since we send the user to the Okta site to clear Okta's session data and then return to the Give site.

We handle this here `src/common/services/api/cart.service.js` - Lines 164 - 181

On the `sessionService.oktaSignOut` we pass a prop which tells session data to store a prop called `forcedUserToLogout`. We also ensure Okta returns the user to the same page.
`src/common/services/session/session.service.js` Lines 317 - 322

On`src/app/productConfig/productConfigForm/productConfigForm.component.js`, we watch for the `forcedUserToLogout` session data and show the warning to the user if present.

wrandall22: Thanks!

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
```diff
@@ -0,0 +1,134 @@
...
+              <p translate style="margin-bottom: 0px">Sign in with Okta to get started.</p>
+              <p>You will be taken to Okta&rsquo;s website, then back to this site.</p>
+            </div>
+          </div>
+
+          <div class="row active-account">
+            <button class="okta-button"
+                    ng-click="$ctrl.signInWithOkta()"
+                    ng-disabled="$ctrl.isSigningIn">
+              <img src="/assets/img/okta.png" alt="Okta logo">
```
wrandall22, line 114:
```suggestion
              <img src="{{$ctrl.imgDomain}}/assets/img/okta.png" alt="Okta logo">
```

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -2,78 +2,32 @@
...
-    translate>Sign In
+<div class="row text-center">
+  <h4 translate>Sign in with Okta to get started.</h4>
+  <p translate>You will be taken to Okta&rsquo;s website, then back to this site.</p>
+</div>
+<div class="row">
+  <button class="okta-button"
+          ng-click="$ctrl.signInWithOkta()"
+          ng-disabled="$ctrl.isSigningIn">
+    <img src="/assets/img/okta.png" alt="Okta logo">
```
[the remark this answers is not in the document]
dr-bizz: Thanks

wrandall22:
the `*cloud` and `nonprod` environments are not in use anymore and can just be deleted, with the exception of `devcloud`.

src/common/services/session/session.service.js
```diff
@@ -317,6 +318,25 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
     }
   }
 
+  function signOutWithoutRedirectToOkta () {
+    const observable = currentRole() === Roles.public
+      ? Observable.of(true)
```
wrandall22, line 323: I don't think we actually want this to be conditional. Logging out a public role will create a new customer in the EP database and ensure that the user has a brand new cart, which we want.

dr-bizz: Gotcha, I'll update it to always log the user out of EP.

wrandall22:
Not sure if the latest has been merged to stage yet or not, but something is still sending me to the home page when hitting branded checkout. You can test by going to https://give-stage2.cru.org/branded-checkout-fl.html

wrandall22:
[Here](https://secure-stage.cru.org/202211-yearend-frontline.html) is a test branded checkout page that is not on give itself, we'll need to make sure the Okta logout works here as well

wrandall22 (changes requested):
Working through the files again to try and find anything we might have missed. Haven't gotten through it all, but want this feedback at least to get through.

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
wrandall22: Not sure why this comment didn't make it into the pending review...

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -115,7 +115,22 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
+          <button
+            type="button"
+            class="btn pull-left add-spouse-btn"
+            ng-class="{ 'btn-default': $ctrl.showSpouseFields, 'btn-primary': !$ctrl.showSpouseFields }"
+            ng-click="$ctrl.toggleSpouseFields()"
+            ng-if="$ctrl.donorDetails['donor-type'] === 'Household'"
+            translate
+          >{{$ctrl.showSpouseFields ? "Remove spouse" : "Add spouse"}}</button>
+        </div>
+      </div>
```
wrandall22, line 131: We should probably check if it is ok to modify branded checkout like this. I'm guessing they will want to only do this for the sign up process, not during checkout and branded checkout which is what is being done here.

src/common/components/registerAccountModal/registerAccountModal.tpl.html
```diff
@@ -1,34 +1,74 @@
...
   <sign-up-modal ng-switch-when="sign-up"
-                 modal-title="$ctrl.modalTitle"
-                 on-state-change="$ctrl.stateChanged(state)"
-                 on-success="$ctrl.onIdentitySuccess()">
-  </sign-up-modal>
-  <forgot-password-modal ng-switch-when="forgot-password"
-                         modal-title="$ctrl.modalTitle"
-                         on-state-change="$ctrl.stateChanged(state)">
-  </forgot-password-modal>
+    on-state-change="$ctrl.stateChanged(state)"
```
wrandall22, line 43: One code style we've used historically, and I'd prefer to continue using, is to line up attributes on multi-line html elements. This would mean either keeping the spacing as it was or moving the `ng-switch-when` down a line and in line with these properties.

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -2,132 +2,268 @@ import angular from 'angular'
...
-          })
-        })
+  describe('loadDonorDetails()', () => {
+    it('should set loadingDonorDetails to false', () => {
+      jest.spyOn($ctrl.orderService, 'getDonorDetails').mockReturnValue(Observable.throw({status: 404}))
+      $ctrl.loadingDonorDetails = true
+      $ctrl.loadDonorDetails()
+      expect($ctrl.loadingDonorDetails).toEqual(false)
+    })
+  });
```
wrandall22, line 173: Remove semicolon

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -2,132 +2,268 @@ import angular from 'angular'
...
-        })
+    it('should not call createAccount', async () => {
+      jest.spyOn($ctrl.sessionService, 'createAccount')
+
+      jest.spyOn($ctrl.orderService, 'getDonorDetails').mockImplementation(() => Observable.from(
+        [signUpFormData]
+      ))
+      $ctrl.$onInit()
+      $ctrl.signUpForm.$valid = false;
+      $ctrl.submitDetails(signUpFormData.email, signUpFormData.name['given-name'], signUpFormData.name['family-name']).then(() =>{
```
wrandall22, line 190: Add space after `=>`. Not sure why the lint checks are not capturing all of these code style issues.

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -2,132 +2,268 @@ import angular from 'angular'
...
+        data: ['Error one', 'Error 2']
+      }))
+      
+
+      jest.spyOn($ctrl.orderService, 'getDonorDetails').mockImplementation(() => Observable.from(
+        [signUpFormData]
+      ))
+      $ctrl.$onInit()
+      $ctrl.signUpForm.$valid = true
+      $ctrl.submitDetails().then(() =>{
```
wrandall22, line 210: Add space

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -2,132 +2,268 @@ import angular from 'angular'
...
+    })
+
+    it("should call createAccount() with the user's data", () => {
+      jest.spyOn($ctrl.orderService, 'getDonorDetails').mockImplementation(() => Observable.from(
+        [signUpFormData]
+      ))
+      jest.spyOn($ctrl.sessionService, 'createAccount').mockImplementation(() => Promise.resolve({
+        status: 'error',
+        data: ['Error one', 'Error 2']
+      }))
```
wrandall22, line 228: Would this not be a `Promise.reject`?

wrandall22 (changes requested):
Got through a few more.

src/common/services/api/cart.service.spec.js
```diff
@@ -259,8 +257,30 @@ describe('cart service', () => {
...
+        })
 
+        it('should delete cookies and addItem to cart', () => {
+          jest.spyOn(self.cartService.sessionService, 'oktaIsUserAuthenticated').mockReturnValue(Observable.from([false]))
           self.cartService.addItem('items/crugive/<some id>', { amount: 50 }).subscribe()
+          expect(self.cartService.sessionService.oktaIsUserAuthenticated).toHaveBeenCalled()
+          expect(self.cartService.sessionService.signOut).not.toHaveBeenCalled()
+        })
+
+        it('should redirect user to login and not run addItem to cart', () => {
```
wrandall22, line 269: This test does not check whether or not `addItem` was run.

src/common/services/api/cart.service.spec.js
```diff
@@ -259,8 +257,30 @@ describe('cart service', () => {
...
+          jest.spyOn(self.cartService.sessionService, 'oktaIsUserAuthenticated').mockReturnValue(Observable.from([true]))
+          self.cartService.addItem('items/crugive/<some id>', { amount: 50 }).subscribe()
+          expect(self.cartService.sessionService.oktaIsUserAuthenticated).toHaveBeenCalled()
+          expect(self.cartService.sessionService.signOut).toHaveBeenCalled()
+        })
+
+        it('should call _addItem()', () => {
```
wrandall22, line 276: In theory we should check for this call in both scenarios

src/common/services/api/profile.service.js
```diff
@@ -81,6 +81,7 @@ class Profile {
       .map((donorDetails) => {
         donorDetails.mailingAddress = formatAddressForTemplate(donorDetails['mailing-address'].address)
         delete donorDetails['mailing-address']
+        donorDetails['registration-state'] = 'NEW'
```
wrandall22, line 84: I'm not sure I understand why this was added. The registration state should come from the API typically. Anyone coming to the profile page should have a registration state (post login steps).

src/common/services/session/session.service.js
```diff
@@ -26,18 +28,35 @@ export const Sessions = {
...
-const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout, envService) {
+const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout, $window, $location, envService) {
   const session = {}
   const sessionSubject = new BehaviorSubject(session)
   let sessionTimeout
   const maximumTimeout = 30 * 1000
+  const authClient = new OktaAuth({
+    issuer: envService.read('oktaUrl'),
+    clientId: envService.read('oktaClientId'),
+    redirectUri: `${window.location.origin}/sign-in.html`,
```
wrandall22, line 50: Does this client get created on branded checkout? Does the non-existence of this page on BCO affect anything?

src/common/services/session/session.service.js
```diff
@@ -46,95 +65,287 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+        data: [`Already logged in to Okta${email ? ` with email: ${email}` : ''}. You will be redirected to the Sign In page in a few seconds.`, 'Another Error'],
+        redirectToSignIn: true
+      }
+    }
+
+    const data = { }
+
+    if (angular.isDefined(email)) data.email = email
+    if (angular.isDefined(firstName)) data.first_name = firstName
+    if (angular.isDefined(lastName)) data.last_name = lastName
```
wrandall22, line 163: Add curly brackets.

src/common/services/session/session.service.js
```diff
@@ -46,95 +65,287 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+        ? errorObject.data.error.split(',')
+            .filter((str) => str.includes(':errorSummary=>') && !str.includes('Api validation failed: login'))
+            .map((str) => str.match(/"([^"]+)"/)[1].replace(/["]/g, ''))
+        : errorObject
+
+      let checkIfAccountIsPending = false
+
+      const formattedErrors = errors.map((error) => {
+        switch (error) {
+          case 'login: An object with this field already exists in the current organization':
```
wrandall22, line 206: Could we do a quick test with a different browser language setting? Wondering if there is any chance that Okta will translate these error messages in that case and bypass our nice error handling.

src/common/services/session/session.service.js
```diff
@@ -46,95 +65,287 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+      })
+    } catch {
+      // closeSession errors out due to CORS. to fix this temporarily, I've added the logout in a catch just in case.
+      if (!redirectHome) {
+        $window.sessionStorage.setItem('forcedUserToLogout', true)
+      }
+      return authClient.signOut({
+        postLogoutRedirectUri: redirectHome ? null : $window.location.href
+      }).catch(() => {
+        $window.location = `https://signon.okta.com/login/signout?fromURI=${envService.read('oktaReferrer')}`
```
wrandall22, line 323: The base url should also be an environment variable.

src/common/services/session/session.service.js
```diff
@@ -229,9 +489,46 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+          dataAsString,
+          {
+            path: '/',
+            domain: isTest ? '' : cookieDomain,
+            expires: moment().add(20, 'minutes').toISOString()
+          }
+        )
+      } else {
+        const dataAsString = $cookies.get(checkoutSavedDataCookieName)
+        if (dataAsString) session.checkoutSavedData = JSON.parse(dataAsString)
```
wrandall22, line 513: Add curly brackets

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+        .signOut(true)
+        .subscribe(() => {
+          expect(sessionService.authClient.revokeAccessToken).toHaveBeenCalled()
+          expect(sessionService.authClient.revokeRefreshToken).toHaveBeenCalled()
+          expect(sessionService.authClient.closeSession).toHaveBeenCalled()
+          expect(sessionService.authClient.signOut).toHaveBeenCalledWith({
+            postLogoutRedirectUri: null
+          })
         })
+    });
```
wrandall22, line 224: Remove semicolon

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+      jest.spyOn(sessionService.authClient, 'revokeAccessToken').mockImplementationOnce(() => Promise.reject())
+      sessionService
+      .signOut()
+      .subscribe(() => {
+        expect(sessionService.authClient.revokeAccessToken).toHaveBeenCalled()
+        expect(sessionService.authClient.revokeRefreshToken).not.toHaveBeenCalled()
+        expect(sessionService.authClient.closeSession).not.toHaveBeenCalled()
+        expect(sessionService.authClient.signOut).toHaveBeenCalled()
+      })
+    });
```
wrandall22, line 236: Remove semicolon

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+
+    it('should add forcedUserToLogout session data', () => {
+      sessionService
+      .signOut(false)
+      .subscribe(() => {
+        expect($window.sessionStorage.getItem('forcedUserToLogout')).toEqual('true')
+      })
+    });
```
wrandall22, line 244: Remove semicolon

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+
+    it('should add forcedUserToLogout if error', () => {
+      jest.spyOn(sessionService.authClient, 'revokeAccessToken').mockImplementationOnce(() => Promise.reject())
+      sessionService
+      .signOut(false)
+      .subscribe(() => {
+        expect($window.sessionStorage.getItem('forcedUserToLogout')).toEqual('true')
+      })
+    });
```
wrandall22, line 253: Remove semicolon

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+
+    it('should redirect the user to okta if all else fails', () => {
+      jest.spyOn(sessionService.authClient, 'revokeAccessToken').mockImplementationOnce(() => Promise.reject())
+      jest.spyOn(sessionService.authClient, 'signOut').mockImplementationOnce(() => Promise.reject())
+      sessionService
+      .signOut()
+      .subscribe(() => {
+        expect($window.location).toEqual(`https://signon.okta.com/login/signout?fromURI=${envService.read('oktaReferrer')}`)
+      })
+    });
```
wrandall22, line 263: Remove semicolon

src/common/services/session/session.service.spec.js
```diff
@@ -153,133 +170,280 @@ describe('session service', function () {
...
+
+    it('should redirect the user to okta if all else fails', () => {
+      jest.spyOn(sessionService.authClient, 'revokeAccessToken').mockImplementationOnce(() => Promise.reject())
+      jest.spyOn(sessionService.authClient, 'signOut').mockImplementationOnce(() => Promise.reject())
+      sessionService
+      .signOut()
+      .subscribe(() => {
+        expect($window.location).toEqual(`https://signon.okta.com/login/signout?fromURI=${envService.read('oktaReferrer')}`)
```
wrandall22, line 261: This should update to env variable when the session service does.

src/common/services/session/session.service.spec.js
```diff
@@ -424,4 +588,292 @@ describe('session service', function () {
...
+
+
+
+  describe('CheckoutSavedData', () => {
+    describe('updateCheckoutSavedData', () => {
+      it('saves the checkout data to a cookie', () => {
+        // isTest set to TRUE
+        sessionService.updateCheckoutSavedData(checkoutData, true)
+        expect(sessionService.session.checkoutSavedData).toBeDefined()
+        const dataStoredInLocalStoarage = $cookies.get(checkoutSavedDataCookieName)
```
wrandall22, line 857:
```suggestion
        const dataStoredInLocalStorage = $cookies.get(checkoutSavedDataCookieName)
```

src/common/services/session/session.service.spec.js
```diff
@@ -424,4 +588,292 @@ describe('session service', function () {
...
+        expect(dataStoredInLocalStoarage).toEqual(JSON.stringify(checkoutData))
+      })
+      it('copies checkout data cookie and stores it on session', () => {
+        delete sessionService.session.checkoutSavedData;
```
wrandall22, line 861: Remove semicolon

src/common/services/session/sessionModal.service.js
```diff
@@ -66,17 +63,9 @@ const SessionModalService = /* @ngInject */ function ($uibModal, $log, modalStat
...
       dismissAnalyticsEvent: 'ga-sign-in-exit'
     }).result,
-    signUp: () => openModal('sign-up').result,
-    forgotPassword: () => openModal('forgot-password').result,
-    resetPassword: () => openModal('reset-password', { backdrop: 'static' }).result,
-    userMatch: () => openModal('user-match', {
-      backdrop: 'static',
-      openAnalyticsEvent: 'ga-registration-match-is-this-you',
-      dismissAnalyticsEvent: 'ga-registration-exit'
-    }).result,
```
wrandall22, line 76: Are we losing these analytics events? Not seeing this anywhere but the tests now.

src/common/services/session/sessionModal.service.spec.js
wrandall22, line 1: Should also remove the test `send analytics event for user-match` since that no longer happens. Or, we need to reinstate it elsewhere.

wrandall22 (changes requested):
Got through all the files, hopefully caught all the interesting things.

src/app/thankYou/accountBenefits/accountBenefits.component.js
```diff
@@ -32,25 +36,15 @@ class AccountBenefitsController {
...
-      this.sessionModalService.signIn(this.getLastPurchaseId()).then(() => {
-        this.sessionModalService.userMatch().then(() => {
-          // Hide component after successful user match
-          this.isVisible = false
-        }, angular.noop)
-      }, angular.noop)
-    }
+    this.sessionModalService.registerAccount().then(() => {
+      this.isVisible = false
+    }, angular.noop)
```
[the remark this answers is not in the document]
dr-bizz: I think you're right, it shouldn't have been removed.

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
```diff
@@ -0,0 +1,134 @@
...
+            </loading>
+          </div>
+        </div>
+
+        <hr />
+
+        <div class="row bottom-links">
+          <div class="col-xs-6">Powered by Okta</div>
+          <div class="col-xs-6">
+            <a href="https://www.cru.org/us/en/about/privacy.html" target="_blank" rel="noopener" class="pull-right">Privacy Policy</a>
```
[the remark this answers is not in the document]
dr-bizz: I've added the `publicCru` variable

src/common/services/session/session.service.js
```diff
@@ -229,9 +499,52 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+          path: '/',
+          domain: isTest ? '' : cookieDomain
+
+        }
+      )
+      return session.checkoutSavedData
+    } catch { }
+  }
+
+  function clearregisterForSiebelLocalKey () {
```
wrandall22, line 546:
```suggestion
  function clearRegisterForSiebelLocalKey () {
```

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -2,78 +2,32 @@
...
   <loading type="overlay" ng-if="$ctrl.isSigningIn">
-    <translate>Signing you in...</translate>
+    <translate translate>Signing you in...</translate>
   </loading>
-</form>
+</div>
+<div class="row text-center">
+  <h4 translate>Don&rsquo;t have an account?</h4>
+  <a class="okta-signup-button" ng-click="$ctrl.onSignUpWithOkta()" style="cursor: pointer">Sign up for Okta</a>
+  <p translate>With Okta, you can sign in to multiple services with one login. You may learn more about Okta <a href="" target="_blank">here</a>.</p>
```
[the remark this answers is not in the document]
dr-bizz: I've added twhe Okta website `https://www.okta.com/`

src/app/thankYou/accountBenefits/accountBenefits.component.spec.js
```diff
@@ -1,10 +1,11 @@
 import angular from 'angular'
 import 'angular-mocks'
 import module from './accountBenefits.component'
-
+import { Observable } from 'rxjs/Observable'
+import 'rxjs/add/observable/from'
 import { Roles } from 'common/services/session/session.service'
 
-describe('thank you', function () {
+fdescribe('thank you', function () {
```
wrandall22, line 8: Intentional change?

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -1,133 +1,151 @@
...
+
+          <div class="row">
+            <div class="col-xs-12">
+              <p>* indicates required field</p>
+            </div>
+          </div>
+
+          <div class="row buttons">
+            <div class="col-xs-6">
+              <button
```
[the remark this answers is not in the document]
dr-bizz: I've updated the button to be a submit button which will fix this issue.

src/common/components/signUpModal/signUpModal.component.spec.js
dr-bizz: This is because `sessionService.createAccount` returns this function`sessionService.catchCreateAccountErrors` when it handles errors, which returns a resolved promise.

src/common/components/userMatchModal/userMatchModal.component.js
```diff
@@ -97,7 +116,8 @@ class UserMatchModalController {
     this.selectContactError = false
     if (angular.isDefined(contact)) {
       this.verificationService.selectContact(contact).subscribe(() => {
-        this.changeMatchState('activate')
+        this.onActivate()
+        this.firstName = contact.name.split(' ')[0]
```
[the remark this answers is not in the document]
dr-bizz: We can split the name at the last space in the name so `Anna Belinda lastname` would become `Anna Belinda`

src/common/components/userMatchModal/userMatchModal.component.spec.js
```diff
@@ -231,19 +255,17 @@ describe('userMatchModal', function () {
...
-      jest.spyOn($ctrl.verificationService, 'getQuestions').mockReturnValue(Observable.of(['a', 'b', 'c']))
+      jest.spyOn($ctrl.verificationService, 'getQuestions').mockReturnValue(Observable.of([{ key: 'a' }, { key: 'b' }, { key: 'c' }]))
       jest.spyOn($ctrl, 'changeMatchState').mockImplementation(() => {})
 
       $ctrl.onActivate()
 
       expect($ctrl.setLoading).toHaveBeenCalledWith({ loading: true })
-      expect($ctrl.answers).toEqual([])
-      expect($ctrl.questions).toEqual(['b', 'c'])
+      expect($ctrl.questions).toEqual([{ key: 'a' }, { key: 'b' }, { key: 'c' }])
```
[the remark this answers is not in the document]
dr-bizz: I don't think I understand your comment?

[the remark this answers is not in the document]
wrandall22: Sorry, think I was experience PR review fatigue. It looks more like we gained a question `a`. We went from
```
expect($ctrl.questions).toEqual(['b', 'c'])
```
to
```
expect($ctrl.questions).toEqual([{ key: 'a' }, { key: 'b' }, { key: 'c' }])
```

Looking in the full context, too, it seems maybe the old way had `a` split out to `question` and had the rest (b, c) in `questions` array.

src/app/thankYou/accountBenefits/accountBenefits.component.spec.js
dr-bizz: I'ved removed this.

src/common/services/session/sessionModal.service.js
dr-bizz: I've added this back in.

src/common/services/session/sessionModal.service.spec.js
dr-bizz: I've added user match back in.

src/common/services/session/session.service.js
dr-bizz: The errors will always be in English as the service account that created the API token has the display language setting set to English.

https://support.okta.com/help/s/article/How-to-change-the-display-language-of-the-error-that-you-receive-trough-API-calls?language=en_US

src/common/services/session/session.service.js
dr-bizz: I don't believe BCO has a way for users to log in via Okta, so I don't think this will be an issue as this is the URI Okta returns the user to after a successful login.

src/common/services/api/profile.service.js
dr-bizz: I agree. I'm unsure why I added it so I'll remove it and test to ensure everything is working as expected.

dr-bizz (commented):
I left a few comments and recommendations.

We should also add a index.page.ts which forwards `/settings` -> `/settings/preferences/`

src/app/designationEditor/designationEditor.component.js
```diff
@@ -65,6 +66,11 @@ class DesignationEditorController {
       this.$window.location = '/'
     }
 
+    this.sessionHandleOktaRedirectService.onHandleOktaRedirect()
+    this.sessionHandleOktaRedirectService.errorMessageSubject.subscribe((errorMessage) => {
+      this.errorMessage = errorMessage
+    })
```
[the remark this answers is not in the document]
wrandall22: I'm not seeing this in the `master` branch. If it fails, it redirects to the home page. However, I do see `$ctrl.errorMessage` being used on the sign in form. Would that show this error message?

src/app/profile/payment-methods/payment-methods.component.js
```diff
@@ -42,6 +44,11 @@ class PaymentMethodsController {
   }
 
   $onInit () {
+    this.sessionHandleOktaRedirectService.onHandleOktaRedirect()
+    this.sessionHandleOktaRedirectService.errorMessageSubject.subscribe((errorMessage) => {
+      this.errorMessage = errorMessage
+    })
```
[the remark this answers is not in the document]
wrandall22: I would think the signInForm would inherit this error message

src/common/components/signInForm/signInButton/signInButton.component.js
```diff
@@ -0,0 +1,84 @@
...
+      this.onSuccess()
+    }, error => {
+      this.isSigningIn = false
+      if (error && error.config && error.config.data && error.config.data.password) {
+        delete error.config.data.password
+      }
+
+      if (error && error.data && error.data.code && error.data.code === 'SIEB-DOWN') {
+        this.$log.error('Siebel is down', error)
+        this.errorMessage = error.data.message
```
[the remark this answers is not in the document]
wrandall22: This is not about sending user data to Siebel, it is reacting to an error when syncing data from Siebel during login on the API side. Siebel is required to be up in order to successfully login, so we need to keep this functionality.

src/common/components/signUpActivationModal/signUpActivationModal.component.js
```diff
@@ -0,0 +1,135 @@
...
+    this.$window = $window
+    this.sessionService = sessionService
+    this.orderService = orderService
+    this.cartService = cartService
+    this.imgDomain = envService.read('imgDomain')
+    this.publicCru = envService.read('publicCru')
+  }
+
+  $onInit () {
+    if (includes([Roles.identified, Roles.registered], this.sessionService.getRole())) {
```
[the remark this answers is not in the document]
wrandall22: This is probably fine since we aren't planning to add more roles anytime soon I think.

wrandall22 (changes requested):

wrandall22:
I noticed the last commit didn't make it to stage, maybe due to being out of sync

src/common/components/accountBenefitsModal/accountBenefitsModal.component.js
```diff
@@ -2,34 +2,43 @@ import angular from 'angular'
...
+  constructor ($location, gettext, sessionService) {
+    this.$location = $location
     this.gettext = gettext
     this.sessionService = sessionService
   }
 
   $onInit () {
     this.modalTitle = this.gettext('Register Your Account for Online Access')
+    const shouldShowRegisterAccountmodal = !!this.$location.search()?.code && !!this.$location.search()?.state
+    if (shouldShowRegisterAccountmodal) this.onStateChange({ state: 'register-account' })
```
wrandall22, line 20: Curly braces

src/common/components/userMatchModal/userMatchModal.component.js
```diff
@@ -157,6 +177,26 @@ class UserMatchModalController {
...
+
+  continueCheckout () {
+    this.$window.location = '/checkout.html'
+  }
+
+  goToOpportunities () {
+    this.$window.location = '/'
+  }
+
+  goToDashboard () {
```
wrandall22, line 197: Might be worth renaming this to `goToGivingDashboard`. The profile page or the receipt page could be considered dashboards as well.

wrandall22 (changes requested):
What are your thoughts about the `register-account` vs `sign-up` language differences throughout the code?

src/app/main/main.component.js
```diff
@@ -46,6 +47,10 @@ const routingConfig = /* @ngInject */ function ($stateProvider, $locationProvide
       url: '/sign-in.html',
       template: '<sign-in></sign-in>'
     })
+    .state('sign-out', {
```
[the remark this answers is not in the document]
dr-bizz: I assume I'll need to create a page on AEM and link it to the signOut page?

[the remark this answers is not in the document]
wrandall22: I believe that is true. We'll also need to create a `sign-out` component in the `giveaem` repo besides creating a page.

dr-bizz:
> `register-account` vs `sign-up`
`sign-up` is the Okta sign-up modal
`register-account` is the Modal to take users through sign-up for Okta and a donor account. This modal has the sign-up form as a step in the process.

I can see why there could be confusion. I see `sign-up` (to Okta) as a step in the process of registering an account. The other steps are creating a donor account and the user match forms.

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
```diff
@@ -89,7 +89,7 @@ <h4 translate>Your email has been activated!</h4>
             </div>
           </div>
 
-          <div class="row active-account">
+          <div class="row active-account" style="margin-bottom: 20px">
```
wrandall22, line 92: Seems this margin should be added to the styles for `active-account` instead of being inline.

dr-bizz: Good catch. I've fixed it.

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -44,6 +44,6 @@ <h4 translate>Don&rsquo;t have an account?</h4>
   </div>
   <div class="need-help--content">
     <p><a ng-href="{{$ctrl.getOktaUrl()}}/signin/forgot-password" target="_blank" translate>Forgot your password?</a></p>
-    <p>Call (888) 278-7233 opt. 1, then 1.<br />9:00 a.m. to 5:00 p.m. ET, Monday-Friday,<br /> or email us at <a href="mailto:eGift@cru.org">eGift@cru.org</a>.</p>
+    <p>Call <a href="tel:+18882787233">(888) 278-7233</a> opt. 1, then 1.<br />9:00 a.m. to 5:00 p.m. ET, Monday-Friday,<br /> or email us at <a href="mailto:eGift@cru.org">eGift@cru.org</a>.</p>
```
dr-bizz, line 47: They asked me to remove the link for the phone number a few weeks ago, have they asked for it to be added back again?

src/assets/scss/components/_signUpActivationModal.scss
```diff
@@ -84,4 +84,9 @@ sign-up-activation-modal {
       max-width: 300px;
     }
   }
+
+  hr,
+  .bottom-links {
+    display: none;
+  }
```
dr-bizz, line 91: I wonder if you could remove the <hr> or button links from the HTML instead of using CSS to remove them.

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
```diff
@@ -22,12 +22,32 @@ <h3 translate>Register Your Account</h3>
...
-              <h4 translate>Activate Account</h4>
-              <p translate style="margin-bottom: 0px">You will need to go to your <b>email</b> and open an email with the subject</p>
-              <strong translate style="display: block; margin-bottom: 8px">Activate Account for Giving to Cru.</strong>
-              <p translate>Verify your account by opening a link that we will send in that email.</p>
+              <h4 translate>Please verify your email address.</h4>
+              <ul>
+                <li translate>You will need to check your email inbox for a message from us.</li>
+                <li>Open the email and locate the verification link provided.</li>
+                <li>Click on the verification link. This will typically be a button or a hyperlink.</li>
+                <li>You will then be redirected to Okta to create your account password.</li>
```
wrandall22, line 32: These should also get `translate` properties in case they get translated at some point.

src/common/services/session/session.service.spec.js
```diff
@@ -357,40 +316,108 @@ describe('session service', function () {
...
         expect(sessionService.authClient.token.parseFromUrl).not.toHaveBeenCalled()
         expect(data).toEqual(false)
         done()
       })
     })
+
+    describe('successful handleOktaRedirect()', () => {
+      const accessToken = 'accessToken'
+      let isAuthenticatedSpy
+      beforeEach(() => {
```
wrandall22, line 333: You might want to add a spy for the `signIn` function since that is tested elsewhere.

wrandall22 (changes requested):
I've noticed that the changes we made for the sign in widget to accept address/etc also impacted the redirect flow sign in (the user has to put in all that data there at login).

dr-bizz:
> I've noticed that the changes we made for the sign-in the widget to accept address/etc. also impacted the redirect flow sign-in (the user has to put in all that data there at login).

There is a setting for not asking users to enter this info in, we can make sure we set it to do not ask.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,17 +1,20 @@
 import angular from 'angular'
 import includes from 'lodash/includes'
+import OktaSignIn from '@okta/okta-signin-widget'
 import sessionService, { Roles } from 'common/services/session/session.service'
 import orderService from 'common/services/api/order.service'
 import template from './signUpModal.tpl.html'
 import cartService from 'common/services/api/cart.service'
+require('assets/okta-sign-in/css/okta-sign-in.min.css')
```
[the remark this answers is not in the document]
dr-bizz: We can add it to `scss` folder, but this would mean each time we update the package, we would also need to update the `scss` file manually. - I prefer not to have any manual work of updating `scss` files when updating the Okta package but instead allow Webpack to manage to update it.

However, I might not see something, so feel free to disagree.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -24,6 +27,79 @@ class SignUpModalController {
     if (includes([Roles.identified, Roles.registered], this.sessionService.getRole())) {
       this.onStateChange({ state: 'sign-in' })
     }
+    window.currentStep = '1' // Default to step 1
```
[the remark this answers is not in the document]
dr-bizz: That is fine.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -24,6 +27,79 @@ class SignUpModalController {
...
+    const rootScope = this.$rootScope // Capture the $rootScope
+    const scope = this.$scope // Capture the $scope
```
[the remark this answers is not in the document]
dr-bizz: I'm unsure, I copied this from when I did it quickly that one time, so most likely it's a mistake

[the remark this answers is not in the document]
dr-bizz: I'll use the normal `rootScope` and `scope`.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -24,6 +27,79 @@ class SignUpModalController {
...
+        baseUrl: `${window.location.origin}/assets/okta-sign-in/`
+      },
+      flow: 'signup',
+      registration: {
+        parseSchema: (schema, onSuccess) => {
+          const step = window.currentStep || '1'
+          const steps = {
+            1: [schema[0], schema[1], schema[2]],
+            2: [schema[3], schema[4], schema[5], schema[6], schema[7]],
+            3: [schema[8]]
```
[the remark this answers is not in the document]
dr-bizz: Sorry, I removed them. I will add them back

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -15,121 +15,10 @@ <h3 translate>Create Your Account</h3>
...
-              <input type="text"
-                    class="form-control  form-control-subtle"
-                    name="nameFamilyName"
-                    ng-model="$ctrl.donorDetails['name']['family-name']"
-                    required
-                    ng-maxlength="50"
-                    ng-disabled="$ctrl.nameFieldsDisabled">
-              <div role="alert" ng-messages="$ctrl.signUpForm.nameFamilyName.$error" ng-if="($ctrl.signUpForm.nameFamilyName | showErrors)">
-                <div class="help-block" ng-message="required" translate>{{'LAST_NAME_ERROR'}}</div>
-                <div class="help-block" ng-message="maxlength" translate>{{'MAX_LENGTH_LAST_NAME_ERROR'}}</div>
```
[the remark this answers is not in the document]
dr-bizz: The first and last names have errors that show when they are not filled out or when their values are more than 50 characters.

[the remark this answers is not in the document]
dr-bizz: Screenshot 2024-12-16 at 4 25 29 PM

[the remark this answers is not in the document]
wrandall22: Ok, where are those limits enforced? We need to ensure they are not going to be bigger than what Siebel can allow.

src/common/components/signUpModal/signUpModal.component.js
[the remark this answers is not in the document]
wrandall22: I think it might warrant a comment in the code. Theoretically it could be made part of the build process, but probably not worth the effort.

src/common/components/signUpModal/signUpModal.tpl.html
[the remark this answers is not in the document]
dr-bizz: I will investigate, but it looks like it's coming from Okta

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,328 @@
...
+        parseSchema: (schema, onSuccess) => {
+          const step = this.$window.currentStep || 1
+          // Split the form into multiple steps for better user experience.
+          // Retain the values entered by the user in each step.
+          // Pre-populate the form fields with existing user details to save time.
+          const steps = {
+            // Step 1: Name and email
+            1: [
+              {
+                ...schema[0],
```
canac, line 73: What do you think about referencing fields by their name instead of their index?

```suggestion
                ...schema.find(field => field.name === 'userProfile.firstName'),
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,328 @@
...
+  }
+
+  async setUpSignUpWidget () {
+    const donorData = await this.loadDonorDetails()
+    this.$window.currentStep = 1 // Default to step 1
+
+    this.oktaSignInWidget = new OktaSignIn({
+      ...this.sessionService.oktaSignInWidgetDefaultOptions,
+      assets: {
+        baseUrl: `${this.$window.location.origin}/assets/okta-sign-in/`
```
canac, line 60:
```suggestion
        baseUrl: '/assets/okta-sign-in/'
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,328 @@
...
+            3: [schema[8]]
+          }
+          onSuccess(steps[step])
+        },
+        preSubmit: (postData, onSuccess) => {
+          const step = this.$window.currentStep
+          const userProfile = postData.userProfile
+
+          if (step === 1) {
+            Object.assign(this.$rootScope, {
```
canac, line 121: I don't think this will trigger a digest for `$rootScope`. It's fine though unless we have watchers on `$rootScope`.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,328 @@
...
+    this.submitting = false
+  }
+
+  $onDestroy () {
+    this.oktaSignInWidget.remove()
+  }
+
+  async setUpSignUpWidget () {
+    const donorData = await this.loadDonorDetails()
+    this.$window.currentStep = 1 // Default to step 1
```
canac, line 55: Is there a reason this couldn't be a controller instance variable?

canac (commented):
Just noting a few things I saw while reading through this code

package.json
```diff
@@ -59,6 +60,7 @@
     "css-loader": "^3.2.0",
     "html-loader": "^0.5.5",
     "jest": "^29.7.0",
+    "jest-canvas-mock": "^2.5.2",
```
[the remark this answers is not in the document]
canac: Probably being used here: https://github.com/CruGlobal/give-web/blob/800873736954e0bc948cd4af2b2494271715d3a7/jest.config.js#L15

src/common/components/signUpActivationModal/signUpActivationModal.component.js
```diff
@@ -0,0 +1,131 @@
...
+      this.getTotalQuantitySubscription = this.cartService.getTotalQuantity().subscribe(count => {
+        this.cartCount = count
+      }, () => {
+        this.cartCount = 0
+      })
+    }
+    this.loadingAccountErrorCount = 0
+    this.initialLoading = true
+    this.getUnverifiedAccount(false)
+    this.getUnverifiedAccountTimoutId = this.$window.setInterval(() => {
```
canac, line 39: I think we should be using `$interval` here and `$timeout` below

canac: I'll do this

src/common/services/session/session.service.js
```diff
@@ -46,95 +76,297 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
       })
   }
 
-  function signOut () {
-    // https://github.com/CruGlobal/cortex_gateway/wiki/Logout
-    return Observable
-      .from($http({
-        method: 'DELETE',
-        url: casApiUrl('/logout'),
+  async function internalSignIn (lastPurchaseId) {
```
canac, line 121: A lot of the new code in this file uses async/await. Native JS promises don't trigger a AngularJS digest cycle in their `then`/`catch` handlers, so we should use observables instead.

src/common/services/session/session.service.js
```diff
@@ -46,95 +76,297 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+    } catch (err) {
+      try {
+        if (err.status === 401) {
+          throw new Error()
         }
-      }))
-      .map((response) => response.data)
+        return {
+          status: 'error',
+          data: err?.data?.error ?? 'SOMETHING_WENT_WRONG'
```
canac, line 291: Did you mean this?

```suggestion
          data: [err?.data?.error ?? 'SOMETHING_WENT_WRONG']
```

src/common/services/session/session.service.js
```diff
@@ -46,95 +76,297 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
-        method: 'POST',
-        url: casApiUrl('/send_forgot_password_email'),
-        withCredentials: true,
-        data: {
-          email: email,
-          passwordResetUrl: passwordResetUrl
+  async function checkCreateAccountStatus (email) {
+    const isAuthenticated = await authClient.isAuthenticated()
+    if (currentRole() !== Roles.public || isAuthenticated) {
+      return 'Already logged in.'
```
canac, line 272: Should we return an object like the other branches?

```suggestion
      return { status: 'error', data: ['ALREADY_LOGGED_IN'] }
```

src/common/services/session/session.service.js
```diff
@@ -46,95 +76,297 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+        withCredentials: true
+      })
+      return {
+        status: 'success',
+        data: createAccountStatus.data
+      }
+    } catch (err) {
+      try {
+        if (err.status === 401) {
+          throw new Error()
```
canac, line 287: Should we return an error instead of throwing?

```suggestion
          return { status: 'error', data: ['FORBIDDEN'] }
```

package.json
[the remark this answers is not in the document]
wrandall22: Right, but I don't see us mocking canvas anywhere.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: This should work

package.json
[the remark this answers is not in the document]
canac: It appears that this package adds some canvas-related APIs to window that the Okta sign in widget needs. If you comment it out, tests fail with `TypeError: Cannot read properties of null (reading 'backingStorePixelRatio')`

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: Here, I don't think I should be using $rootscope. I've switched it to $scope as other components aren't relying on this information.

src/common/components/signUpActivationModal/signUpActivationModal.component.js
dr-bizz: signUpActivationModal.component.js is one of the components that need to be deleted since we aren't handling that anymore.

I'll go through the code and remove any components or code we no longer need.

dr-bizz: This is now complete

src/common/services/session/session.service.js
dr-bizz: This code is now removed

src/common/services/session/session.service.js
dr-bizz: This code no longer exists

src/common/services/session/session.service.js
dr-bizz: I have recreated the functions to use observables instead of promises.

Thank you for spotting this; it allowed me to improve the code.

src/common/components/contactInfo/contactInfo.tpl.html
wrandall22: Going to do a bump of this since the checkout process is negatively impacted by this block of code.

Screenshot 2025-01-10 at 9 32 24 AM

dr-bizz: Since we are no longer using the contactInfo during sign up, I wonder if we can remove this code all together

wrandall22: I think so, yeah. Anything to do with showSpouseFields could probably go away.

src/common/components/signUpModal/signUpModal.component.js
canac: That makes sense. We can leave it as-is.

src/common/components/contactInfo/contactInfo.tpl.html
dr-bizz: This has now been reverted

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -236,30 +277,30 @@ class SignUpModalController {
...
-  async loadDonorDetails () {
+  loadDonorDetails () {
     this.loadingDonorDetails = true
-    return new Promise((resolve) => {
-      this.orderService.getDonorDetails().subscribe(
-        (data) => {
-          this.loadingDonorDetails = false
-
+    return new Observable(observer => {
+      this.orderService.getDonorDetails().subscribe({
```
canac, line 284: Instead of making a new observable, I think you can use `map` to transform the donor details and `do` to inject error handling:

```js
return this.orderService.getDonorDetails().map((data) => {
 // Update donorData as needed
 return donorData;
}.do({
  error: () => {
    // Do error handling
  },
  finally: () => {
    this.loadingDonorDetails = false
  }
})
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,326 @@
...
-    this.setPristine()
-  }
+    if (includes([Roles.identified, Roles.registered], this.sessionService.getRole())) {
+      this.onStateChange({ state: 'sign-in' })
+    }
+    this.currentStep = 1
+    this.donorDetails = {}
+    this.loadDonorDetails().finally(() => {
+      this.setUpSignUpWidget()
+    }).subscribe()
```
canac, line 38:
```suggestion
    this.loadDonorDetails().subscribe({
      finally: () => this.setUpSignUpWidget()
    })
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -63,49 +76,75 @@ class SignUpModalController {
...
+                type: 'select',
+                options: {
+                  household: this.giveAsIndividualTxt,
+                  organization: this.giveAsOrganizationTxt,
+                },
+                'label-top': true,
+                label: 'Account Type',
+                required: true,
+                wide: true,
+                value: this.$scope.accountType ?? 'household',
```
canac: Is there a reason why we're not pulling this from `donorData['donor-type']` if it is set there?

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -129,7 +169,8 @@ class SignUpModalController {
               state: userProfile.state,
               zipCode: userProfile.zipCode,
               countryCode: userProfile.countryCode,
-              primaryPhone: userProfile.primaryPhone
+              primaryPhone: userProfile.primaryPhone,
+              organizationName: postData.organizationName,
```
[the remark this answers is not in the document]
dr-bizz: We don't need to send the organization Name and accountType to Okta, as that data is meant for the Cortex API.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: Good catch

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I don't think RxJS 5 supports using `finally ` inside `subscribe()`.

Are you wanting to do something like this? If so, `complete` only runs when the Observable successfully completes, so if it errors it will not fire.
```
this.loadDonorDetails().subscribe({
  complete: () => this.setUpSignUpWidget(),
});
```

canac: I'm sorry, you're right. I was thinking of `complete`, but you're right that it doesn't fire when the observable errors.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,322 @@
...
+                type: 'select',
+                options: {
+                  household: this.giveAsIndividualTxt,
+                  organization: this.giveAsOrganizationTxt
+                },
+                'label-top': true,
+                label: 'Account Type',
+                required: true,
+                wide: true,
+                value: this.$scope.accountType ?? donorData['donor-type'] ?? 'household'
```
canac, line 119: Cortex uses capitalized `'Household'` for the `donor-type`.

```suggestion
                options: {
                  Household: this.giveAsIndividualTxt,
                  Organization: this.giveAsOrganizationTxt
                },
                'label-top': true,
                label: 'Account Type',
                required: true,
                wide: true,
                value: this.$scope.accountType ?? donorData['donor-type'] ?? 'Household'
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -1,81 +1,322 @@
...
+                ...schema[7],
+                value: this.$scope.countryCode ?? donorData.mailingAddress?.country ?? ''
+              },
+              {
+                ...schema[8],
+                value: this.$scope.primaryPhone ?? donorData['phone-number'] ?? ''
+              }
+            ],
+            // Step 3: Password
+            3: [schema[8]]
```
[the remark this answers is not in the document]
dr-bizz: Yes, I put this here until we can show the password

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I spotted this, this morning. I will push my code up

src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -165,6 +165,8 @@ class RegisterAccountModalController {
         } else {
           // Proceed to Step 3
           this.stateChanged('contact-info')
+          // Ensure modal is large enough to display contact info form
+          this.setModalSize('md')
```
canac: Is this necessary? I think `this.stateChanged` is supposed to handle this. If it doesn't, we should probably update the logic there.

dr-bizz: I removed this logic with my new changes

src/common/services/session/session.service.js
```diff
@@ -46,95 +75,204 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
+          observer.complete()
+        },
+        error: err => {
+          observer.error(err)
         }
-      }))
-      .map((response) => response.data)
+      })
+    })
   }
```
canac, line 208: Unless I'm missing something, the following is equivalent. If that is the case, we probably don't need separate `signOut` and `internalSignOut` functions.

```suggestion
  function signOut (redirectHome = true) {
    return internalSignOut(redirectHome)
  }
```

dr-bizz:
Recaptcha is breaking lint

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -1221,28 +1223,6 @@ describe('signUpForm', function () {
...
-      }
-
-      $ctrl.signIn().then(() => {
-        expect(handleLoginRedirect).not.toHaveBeenCalled()
-        expect($ctrl.reRenderWidget).toHaveBeenCalledTimes(1)
-        expect($ctrl.$log.error).toHaveBeenCalledTimes(1)
-        expect($ctrl.$log.error).toHaveBeenCalledWith('Error showing Okta sign in widget.', error)
-        done()
-      })
-    })
```
dr-bizz: I also looked into this this morning. I'm unsure what has changed to prevent it from working. The signIn function caches the first response, which returns the token.

If we can get this text working again, it would be helpful, as we don't test whether a sign-in error is logged elsewhere.

dr-bizz: I tried that, but it seemed like the test was caching the first good response, even though I was telling it to error. Maybe you will have better luck than I did.

canac: I added it back in c7e5fa1dfe2cbaa2def56d98396613efc8136a1a

src/common/services/api/cart.service.js
```diff
@@ -167,8 +167,15 @@ class Cart {
     if (!disableSessionRestart && this.sessionService.getRole() === Roles.public) {
       return this.getTotalQuantity().mergeMap((total) => {
         if (total <= 0) {
-          return this.sessionService.signOut().mergeMap(() => {
-            return this._addItem(uri, data)
+          return this.sessionService.oktaIsUserAuthenticated().mergeMap((isAuthenticated) => {
+            if (!isAuthenticated) {
+              return this._addItem(uri, data)
+            }
+            // SignOut() will redirect user to Okta to clear session,
```
canac, line 174: @dr-bizz What does this code do? Why are we signing out of Okta when the user is authenticated?

src/common/services/session/sessionModal.service.js
```diff
@@ -91,9 +96,6 @@ export default angular
   .factory(serviceName, SessionModalService)
   .config(function (modalStateServiceProvider) {
     modalStateServiceProvider.registerModal(
-      'reset-password',
       /* @ngInject */
-      function (sessionModalService) {
-        sessionModalService.resetPassword()
-      })
+      function (sessionModalService) {})
   })
```
canac, line 101: @dr-bizz Can we remove this `config` block? It doesn't seem like it's doing anything.

canac (approved):
This looks really good! I read through every line to look for obvious issues, but I skimmed sections that I know we've reviewed and tested thoroughly in the past.

src/common/services/session/sessionModal.service.js
dr-bizz: Yes, we can remove it. It was ensuring the forgot-password modal appeared whenever reset-password was triggered. But as we no longer use that modal, we can remove it.

src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -64,45 +104,42 @@ class RegisterAccountModalController {
...
-    this.verificationService.postDonorMatches().subscribe(() => {
-      // Donor match success, Proceed to step 5.
-      this.stateChanged('user-match')
-    }, () => {
-      // Donor Match failed, Register Account workflow failed
-      this.onCancel()
+    if (angular.isDefined(this.verificationServiceSubscription)) this.verificationServiceSubscription.unsubscribe()
+    this.verificationServiceSubscription = this.verificationService.postDonorMatches().subscribe({
+      next: () => { this.stateChanged('user-match') }, // Donor match success, Proceed to step 5.
+      error: () => { this.onCancel() } // Donor Match failed, Register Account workflow failed
```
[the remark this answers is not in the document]
dr-bizz: I have noted this as a Phase 2 task

src/common/services/api/cart.service.spec.js
dr-bizz: This has been completed

src/app/profile/payment-methods/payment-methods.component.js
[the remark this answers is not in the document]
dr-bizz: Same here, we no longer need this. It was added by me a long time ago.

src/common/services/session/sessionHandleOktaRedirect.service.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    it('should emit and log error when handleOktaRedirect errors', () => {
+      jest.spyOn(sessionService, 'handleOktaRedirect').mockImplementation(() => Observable.throw('ERROR_MESSAGE'))
+      
+      try {
+        sessionHandleOktaRedirectService.onHandleOktaRedirect()
+      } catch {}
+
+      expect($log.error).toHaveBeenCalledWith('Failed to redirect from Okta', 'ERROR_MESSAGE')
+      sessionHandleOktaRedirectService.errorMessageSubject.subscribe((errorMessage) => {
+        expect(errorMessage).toEqual('generic')
```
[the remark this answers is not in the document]
dr-bizz: page removed

src/common/components/userMatchModal/userMatchModal.component.js
```diff
@@ -117,7 +117,9 @@ class UserMatchModalController {
     if (angular.isDefined(contact)) {
       this.verificationService.selectContact(contact).subscribe(() => {
         this.onActivate()
-        this.firstName = contact.name.split(' ')[0]
+        this.firstName = contact.name.includes(' ') 
+          ? contact.name.substring(0,contact.name.lastIndexOf(' '))
```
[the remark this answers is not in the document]
dr-bizz: We have since rewritten this component. I remember us redoing this part because of your comment.

src/common/components/signUpActivationModal/signUpActivationModal.tpl.html
dr-bizz: No longer using them

src/common/services/session/session.service.spec.js
```diff
@@ -357,40 +316,108 @@ describe('session service', function () {
...
+          }
+        })
+        jest.spyOn(sessionService.authClient.tokenManager, 'setTokens').mockResolvedValue(true)
+        jest.spyOn(sessionService.authClient.tokenManager, 'getTokens').mockResolvedValueOnce({
+          accessToken: {
+            accessToken
+          }
+        })
+        jest.spyOn(sessionService.authClient.token, 'getWithRedirect').mockResolvedValue(true)
+      });
```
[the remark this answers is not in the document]
dr-bizz: We have semicolons dotted all around the code. Instead of removing this one, I would suggest that we make it into an ESLint rule in a future PR.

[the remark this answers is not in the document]
dr-bizz: I've added it to the Give Okta spreadsheet - https://docs.google.com/spreadsheets/d/1e6umeVoVAIlpgHQUk_ybPX359RWnBA0FIsqP-rvvZ-g/edit?gid=0#gid=0

src/common/services/session/session.service.spec.js
```diff
@@ -284,71 +284,30 @@ describe('session service', function () {
...
-        $httpBackend.flush()
+      // Observable.finally is fired after the test, this defers until it's called.
+      // eslint-disable-next-line angular/timeout-service
+      setTimeout(() => {
+        expect($rootScope.$broadcast).toHaveBeenCalledWith(SignInEvent)
         done()
       })
     })
 
     it('should handle a failed login', done => {
```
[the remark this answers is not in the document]
dr-bizz: I think we still need it, as we're testing that `handleOktaRedirect` handles the user not being logged in correctly.

Also, 5 months ago, I added the tests `should trigger BodySignInEvent event` & `'should return SignInEvent broadcast when authenticated` to add more tests to the login process.

src/common/services/session/session.service.spec.js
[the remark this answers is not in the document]
wrandall22: It is already a lint rule, but the lint rules are ignoring tests.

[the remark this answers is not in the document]
dr-bizz: I have removed them

src/common/components/accountBenefitsModal/accountBenefitsModal.component.js
```diff
@@ -21,22 +22,27 @@ class AccountBenefitsModalController {
       // No need to sign in if we already are
       this.onSuccess()
     } else {
-      this.onStateChange({ state: 'sign-in' })
+      this.onRegister()
     }
   }
+
+  onFailure () {
```
dr-bizz, line 29: I think this can be removed

src/common/services/session/session.service.js
```diff
@@ -46,95 +79,185 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
   return {
     session: session,
     sessionSubject: sessionSubject,
+    authClient: authClient, // Exposed for tests only
```
dr-bizz, line 82: Is there a way we can not expose this?

src/common/services/session/sessionEnforcer.service.js
```diff
@@ -21,7 +21,7 @@ export const EnforcerModes = {
   donor: 'donor'
 }
 
-const SessionEnforcerService = /* @ngInject */ function (orderService, sessionService, sessionModalService) {
+const SessionEnforcerService = /* @ngInject */ function ($window, orderService, sessionService, sessionModalService) {
```
dr-bizz, line 24: We can remove `$window`

src/common/app.config.js
```diff
@@ -99,15 +116,21 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
         publicCru: 'https://stage.cru.org',
         publicGive: 'https://give-stage2-next.cru.org',
         acsUrl: 'https://cru-mkt-stage1.adobe-campaign.com/lp/LP63?_uuid=f1938f90-38ea-41a6-baad-9ac133f6d2ec&service=%404k83N_C5RZnLNvwz7waA2SwyzIuP6ATcN8vJjmT5km0iZPYKUUYk54sthkZjj-hltAuOKDYocuEi5Pxv8BSICoA4uppcvU_STKCzjv9RzLpE4hqj&pkey=',
-        isBrandedCheckout: false
+        isBrandedCheckout: false,
+        oktaUrl: 'https://signon.okta.com',
+        oktaClientId: '0oa1p8kb6gr1IGGdc0h8',
```
dr-bizz, line 121:
```suggestion
        oktaUrl: 'https://cru.oktapreview.com
        oktaClientId: '0oa26fp9l9iFxuKJf0h8
```

src/common/services/api/order.service.js
```diff
@@ -445,6 +445,20 @@ class Order {
     }
   }
 
+  saveDonorDataForRegistration (donorDetails) {
```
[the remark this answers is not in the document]
dr-bizz: checked and it's working

PR #1083, opened 2023-12-20
[HS-1071657] Migrate cart and view product data layer events to GA4

canac:
Migrate add_to_cart, remove_from_cart, and view_item events to GA4.

https://secure.helpscout.net/conversation/2457822430/1071657?viewId=7296729

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -30,9 +30,8 @@ function testingTransactionName (item) {
 // Generate a datalayer product object
 const generateProduct = suppressErrors(function (item, additionalData = {}) {
   const sessionStorageTestName = testingTransactionName(item)
-  const testingTransaction = sessionStorageTestName
-    ? window.sessionStorage.getItem(sessionStorageTestName) || undefined
-    : undefined
+  const testingTransaction = Boolean(sessionStorageTestName &&
+    window.sessionStorage.getItem(sessionStorageTestName) === 'true').toString()
   const price = additionalData?.price || item.amount
   const category = additionalData?.category || item.designationType
   const name = additionalData?.name || item.displayName || undefined
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -216,15 +216,14 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
       // Send GTM Advance Ecommerce event
       $window.dataLayer = $window.dataLayer || []
       $window.dataLayer.push({
-        event: 'add-to-cart',
+        event: 'add_to_cart',
         ecommerce: {
           currencyCode: 'USD',
-          add: {
-            products: [generateProduct(productData, {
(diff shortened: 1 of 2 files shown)


src/app/analytics/analytics.factory.js
```diff
@@ -251,12 +250,11 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
...
       $window.dataLayer = $window.dataLayer || []
       $window.dataLayer.push({
-        event: 'remove-from-cart',
+        event: 'remove_from_cart',
         ecommerce: {
           currencyCode: 'USD',
-          remove: {
-            products: [generateProduct(item)]
-          }
+          value: item.amount.toFixed(2),
```
wrandall22, line 256: I would be curious how this is working, and not needing to be `AMOUNT`. I know the old code is doing it, but maybe it is broken?

wrandall22: I guess it wouldn't be broken, otherwise the cart would not show the data since the `cart.tpl.html` is using lowercase.

wrandall22: I see where the transformation is, in `cart.service.js` `handleCartResponse()`

wrandall22:
I'd prefer the first 2 commits to be squashed together before merging.

canac:
Will requested that `testing_transaction` be explicitly set to `false` instead of being missing.

We should consider sending a boolean instead of the strings `"true"` and `"false"` like branded checkout does, but I don't want to change that yet until we can prove that it won't break anything.

wrandall22 (approved):

canac:
According to Will, this change is not high urgency and wait until after the code freeze to be merged.

src/app/analytics/analytics.factory.js
```diff
@@ -448,12 +445,12 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
...
       $window.dataLayer = $window.dataLayer || []
       $window.dataLayer.push({
-        event: 'give-gift-modal',
+        event: 'view_item',
         ecommerce: {
           currencyCode: 'USD',
-          detail: {
-            products: [generateProduct(modifiedProductData)]
-          }
+          // value is unavailable until the user selects a gift amount
```
dr-bizz, line 451: Will there ever be a `view_item` which has a value? If no, should we have it this object, or do the analytics guys need it?

dr-bizz (commented):
Everything looks great! Thanks for improving my code with the `testingTransaction`

src/app/analytics/analytics.factory.js
canac: Good question. `view_item` won't ever have a `value` amount at this stage because the user hasn't selected the gift amount or frequency yet. When I implemented the GA4 spec for branded checkout, the spec docs said, "If no value exists for an attribute in the dataLayer, please pass JavaScript `undefined` as the value." That's why I'm passing `undefined` instead of omitting that field.

PR #1084, opened 2023-12-20
[EP-2471] Customize payment method loading error

canac:
Show a custom error message for 500 errors when loading payment methods until the underlying issue is fixed on the server.

https://jira.cru.org/browse/EP-2471


canac:
@wrandall22 Do we want to merge this immediately despite the code freeze?

wrandall22:
Yes, but we need to be prepared to rollback if needed.

PR #1085, opened 2023-12-21
[no-Jira] Fix momentjs invalid date console warning

canac:
Fix momentjs console warning (at least in tests) about parsing a date in an unexpected format introduced in #1054. Only the day/month/year of `recurringDate` is used, so I removed the time to fix the warning.

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -204,7 +204,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
         cart.item[0].attributes.donationType = 'one-time donation'
       } else {
         cart.item[0].attributes.donationType = 'recurring donation'
-        recurringDate = moment(`${moment().year()}-${itemConfig.RECURRING_START_MONTH}-${itemConfig.RECURRING_DAY_OF_MONTH} ${moment().format('h:mm:ss a')}`)
+        recurringDate = moment(`${moment().year()}-${itemConfig.RECURRING_START_MONTH}-${itemConfig.RECURRING_DAY_OF_MONTH}`, 'YYYY-MM-DD')
       }
 
       // Set donation frequency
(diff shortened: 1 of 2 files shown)


dr-bizz:
I remeber adding this due to time differences. As someone might be in NZ and are 12 hours ahead. I couldn't reproduce the error I was trying to fix when switching between my version and your version. Can you test it to ensure it's the correct date even if the user's is 16 hours ahead of EST?

dr-bizz (commented):
I remeber adding this due to time differences. As someone might be in NZ and are 12 hours ahead. I couldn't reproduce the error I was trying to fix when switching between my version and your version. Can you test it to ensure it's the correct date even if the user's is 16 hours ahead of EST?

canac:
> I remeber adding this due to time differences. As someone might be in NZ and are 12 hours ahead. I couldn't reproduce the error I was trying to fix when switching between my version and your version. Can you test it to ensure it's the correct date even if the user's is 16 hours ahead of EST?

@dr-bizz I tried a variety of times of day in the New Zealand timezone and was not able to reproduce that issue. According to [the docs](https://momentjs.com/guides/#/parsing/local-utc-zone/), when `moment()` is given a date without an offset, it will be interpreted as local time, which is what we want. Maybe at one point you were parsing an ISO date string with an offset, which would be converted to local time, and that was causing the issues?

canac:
`giftStartDate` is always a `moment` object, so I updated the test data to reflect that. That also fixed an issue where tests were sometimes failing because the mocked start date wasn't being treated as local time because it included an offset.

canac:
To see the issue that I fixed in 696bf383832d60b03261be03b50b255e31d5bf35, run the tests in a timezone west of Eastern time, i.e. `TZ=US/Central yarn test`.

dr-bizz (approved):
This is great.

When switching my local time via dev tools to New Zealand, It seems the recurring day on the add to cart form doesn't update, so it must be getting the date day from somewhere else. So it's not necessary to include  `${moment().format('h:mm:ss a')}` on the analytics. That might be an issue to fix later, or this might not be an issue for people in New Zealand at all. I'll check with my brother.

You're good to move forward as your code is working.

wrandall22 (approved):

PR #1086, opened 2024-01-12
[EP-2483] Include next draw date when changing frequency

canac:
The API server needs the next draw date information when the frequency changes to something other than monthly.

https://jira.cru.org/browse/EP-2483

### DIFF
--- a/src/common/models/recurringGift.model.js
+++ b/src/common/models/recurringGift.model.js
@@ -201,7 +201,15 @@ export default class RecurringGiftModel {
   }
 
   get toObject () {
-    return this.gift
+    if (this.gift['updated-start-month'] === '') {
+      return this.gift
+    }
+
+    // When the start month changes the server needs the recurring day of month even if it didn't change
+    return {
+      ...this.gift,
+      'updated-recurring-day-of-month': this.gift['updated-recurring-day-of-month'] || this.parentDonation['recurring-day-of-month']
+    }
   }
 
   clone () {
--- a/src/common/models/recurringGift.model.spec.js
+++ b/src/common/models/recurringGift.model.spec.js
@@ -564,6 +564,27 @@ describe('recurringGift model', () => {
     it('should return the object to send to the api', () => {
       expect(giftModel.toObject).toEqual(giftModel.gift)
     })
+
+    it('should include the recurring day of month when the month changes', () => {
+      giftModel.startMonth = '6'
+      expect(giftModel.toObject).toEqual({
+        ...giftModel.gift,
+        'updated-recurring-day-of-month': '15',
+        'updated-start-month': '06',
+        'updated-start-year': '2015',
+      })
+    })
+
+    it('should not override the modified recurring day of month when the month changes', () => {
+      giftModel.transactionDay = '20'
+      giftModel.startMonth = '6'
+      expect(giftModel.toObject).toEqual({
+        ...giftModel.gift,
+        'updated-recurring-day-of-month': '20',
+        'updated-start-month': '06',
+        'updated-start-year': '2015',
+      })
+    })
   })
 
   describe('clone', () => {


src/common/models/recurringGift.model.js
```diff
@@ -201,7 +201,17 @@ export default class RecurringGiftModel {
...
+    const interval = this.gift['updated-rate'].recurrence.interval
+    if (interval === '' || interval === 'Monthly') {
+      return this.gift
+    }
+
+    return {
+      ...this.gift,
+      'updated-recurring-day-of-month': this.gift['updated-recurring-day-of-month'] || this.parentDonation['recurring-day-of-month'],
+      'updated-start-month': this.gift['updated-start-month'] || this.nextGiftDate.format('MM'),
+      'updated-start-year': this.gift['updated-start-year'] || this.nextGiftDate.format('YYYY')
```
wrandall22, line 213: So, I don't think this is quite what we want. The idea is for `this.parentDonation['recurring-day-of-month']` to be sent on the PUT if it did not change. This is always setting the updated values (telling the server that something DID change).

canac: @wrandall22 Are you saying we need to send the `recurring-day-of-month` if `updated-start-month` isn't empty?

What about when the frequency doesn't change but the start month does? Does the server need the `recurring-day-of-month` in that case?

I'm probably misunderstanding something because the server still 500s with this payload:

```json
{
  "donations": [{
    "donation-lines": [{
      "amount": 50,
      "designation-active": true,
      "designation-name": "Caleb and Amy Cox",
      "designation-number": "0864190",
      "donation-line-row-id": "1-59D6WA5",
      "donation-line-status": "Standard",
      "payment-instrument-id": "mm3gmnzxgrrdollfmjrtqljumy4gkljymnrteljzgjstszrqmy4dimjyme=",
      "updated-amount": "",
      "updated-designation-number": "",
      "updated-donation-line-status": "",
      "updated-payment-instrument-id": "",
      "updated-rate": { "recurrence": { "interval": "Quarterly" } },
      "updated-recurring-day-of-month": "",
      "updated-start-month": "03",
      "updated-start-year": "2024",
      "recurring-day-of-month": "12"
    }],
    "donation-status": "Active",
    "effective-status": "Active",
    "rate": { "recurrence": { "interval": "Monthly" } },
    "donation-row-id": "1-59D6WA1"
  }]
}
```

wrandall22: > Are you saying we need to send the recurring-day-of-month if updated-start-month isn't empty?

Yes (or we could put it into the updated day of month since the update is already happening).

> What about when the frequency doesn't change but the start month does?

The frequency doesn't matter in the equation, if the start month changes we need to send either a day of month or updated day of month. I'm seeing also that the updated-start-year is required as well when the start month changes (maybe already handling that).

> I'm probably misunderstanding something because the server still 500s with this payload

In that payload the `recurring-day-of-month` is in the wrong place, it would need to be on the parent object.

canac: > Yes (or we could put it into the updated day of month since the update is already happening).

Putting it in the updated day of month seems simpler. Does that mean that this is a valid payload when the month is changing and the day is staying the same?

```json
{
  "donations": [{
    "donation-lines": [{
      ...
      "updated-rate": { "recurrence": { "interval": "" } },
      "updated-recurring-day-of-month": "12",
      "updated-start-month": "01",
      "updated-start-year": "2024"
    }],
    "donation-status": "Active",
    "effective-status": "Active",
    "rate": { "recurrence": { "interval": "Quarterly" } },
    "donation-row-id": "1-59D6WA1"
  }]
}
```

While sending a payload like this when nothing about the dates changes?

```json
{
  "donations": [{
    "donation-lines": [{
      ...
      "updated-rate": { "recurrence": { "interval": "" } },
      "updated-recurring-day-of-month": "",
      "updated-start-month": "",
      "updated-start-year": ""
    }],
    "donation-status": "Active",
    "effective-status": "Active",
    "rate": { "recurrence": { "interval": "Quarterly" } },
    "donation-row-id": "1-59D6WA1"
  }]
}
```

If not, I can put it on the `donations` entry:


```json
{
  "donations": [{
    "donation-lines": [{
      ...
      "updated-rate": { "recurrence": { "interval": "" } },
      "updated-recurring-day-of-month": "",
      "updated-start-month": "01",
      "updated-start-year": "2024"
    }],
    "recurring-day-of-month": "12",
    "donation-status": "Active",
    "effective-status": "Active",
    "rate": { "recurrence": { "interval": "Quarterly" } },
    "donation-row-id": "1-59D6WA1"
  }]
}
```

> The frequency doesn't matter in the equation, if the start month changes we need to send either a day of month or updated day of month. I'm seeing also that the updated-start-year is required as well when the start month changes (maybe already handling that).

Yes it already handles that. Whenever `RecurringGiftModel` changes `updated-start-month`, it also changes `updated-start-year`.

wrandall22: Both of those payloads look good to me.

src/common/models/recurringGift.model.spec.js
```diff
@@ -564,6 +564,27 @@ describe('recurringGift model', () => {
...
+      expect(giftModel.toObject).toEqual({
+        ...giftModel.gift,
+        'updated-recurring-day-of-month': '15',
+        'updated-start-month': '06',
+        'updated-start-year': '2015',
+      })
+    })
+
+    it('should not override the modified recurring day of month when the month changes', () => {
+      giftModel.transactionDay = '20'
```
wrandall22, line 579: Does this call `set transactionDay (value)` under the hood?

wrandall22:
Code is looking good; I haven't tested the new code on staging yet.

dr-bizz (commented):
The code looks good. When I tested it locally, it worked. I'll let Bill approve this request as he knows more about this than I do.

wrandall22 (approved):

wrandall22:
Tests well on staging.

PR #1087, opened 2024-01-18
[no-Jira] Fix missing add_to_cart event

canac:
# Context

The `add_to_cart` datalayer event was not being fired when the user input a custom amount. This is because the amount would be a string instead of a number and `itemConfig.AMOUNT.toFixed(2)` was throwing an exception that `suppressErrors` was silently hiding.

# Fix

Parse the custom amount string into a number. Also, log caught exceptions to make it easier to diagnose issues like this in the future.

Tested in staging ✅

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -275,8 +275,9 @@ class ProductConfigFormController {
   }
 
   changeCustomAmount (amount, retainCoverFees) {
-    this.checkAmountChanged(amount)
-    this.itemConfig.AMOUNT = amount
+    const amountAsNumber = parseFloat(amount)
+    this.checkAmountChanged(amountAsNumber)
+    this.itemConfig.AMOUNT = amountAsNumber
     this.customAmount = amount
     this.customInputActive = true
     if (!retainCoverFees && this.amountChanged) {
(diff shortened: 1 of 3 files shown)


src/app/productConfig/productConfigForm/productConfigForm.component.js
```diff
@@ -274,7 +274,9 @@ class ProductConfigFormController {
     this.updateQueryParam({ key: giveGiftParams.amount, value: amount })
   }
 
-  changeCustomAmount (amount, retainCoverFees) {
+  changeCustomAmount (stringAmount, retainCoverFees) {
```
wrandall22, line 277: I forget if this is called before or after the custom validators.

wrandall22: Looks like this change does alter the way the custom amount field works in that it no longer allows `$` or `,` to be added. This is not necessarily a bad change, but it is a change of functionality on the form itself vs just fixing the analytics.

canac: Good catch. I'm trying a different approach that should preserve the the punctuation so that `itemConfig.AMOUNT` can consistently be a number throughout the application. But if that doesn't work, I'll just do the parsing in the datalayer.

wrandall22: Note, too, that `itemConfig.AMOUNT` is also coming from the API, probably as a string, but I'm not 100% sure.

canac:
@wrandall22 `$` and `,` appear to be preserved now.

wrandall22 (approved):

PR #1088, opened 2024-02-01
[EP-2485] Prevent duplicate purchase analytics events in branded checkout

canac:
Previously, a `purchase` event was being fired by the regular checkout code and by branded checkout on the confirmation page, resulting in duplicate `purchase` events in branded checkout. This PR makes sure that only one `purchase` event is fired in branded checkout.

https://jira.cru.org/browse/EP-2485

### DIFF
--- a/src/app/thankYou/summary/thankYouSummary.component.js
+++ b/src/app/thankYou/summary/thankYouSummary.component.js
@@ -21,9 +21,10 @@ const componentName = 'thankYouSummary'
 
 class ThankYouSummaryController {
   /* @ngInject */
-  constructor ($rootScope, $window, analyticsFactory, orderService, profileService, sessionModalService, designationsService, $log) {
+  constructor ($rootScope, $window, analyticsFactory, envService, orderService, profileService, sessionModalService, designationsService, $log) {
     this.$rootScope = $rootScope
     this.$window = $window
+    this.envService = envService
     this.orderService = orderService
     this.profileService = profileService
     this.sessionModalService = sessionModalService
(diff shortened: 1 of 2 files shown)


canac:
Tested in staging ✅

src/app/thankYou/summary/thankYouSummary.component.js
```diff
@@ -73,7 +74,9 @@ class ThankYouSummaryController {
 
         this.analyticsFactory.pageLoaded()
         this.analyticsFactory.setPurchaseNumber(data.rawData['purchase-number'])
-        this.analyticsFactory.transactionEvent(this.purchase)
+        if (!this.envService.read('isBrandedCheckout')) {
+          this.analyticsFactory.transactionEvent(this.purchase)
```
wrandall22, line 78: I wonder how this change impacts the two thank you scenarios in branded checkout:
1) The default thank you page
2) A custom thank you page

Is there any easy way to confirm this?

canac: `this.brandedAnalyticsFactory.purchase()` is still always called by https://github.com/CruGlobal/give-web/blob/baa8293d78c595cee8d9eaafcd77696488c29ad2/src/app/thankYou/summary/thankYouSummary.component.js#L71-L73 This just changes it to stop emitting another `purchase` event from `analyticsFactory`. I tested the before and after on JPF, which has a custom thank you page. Let me know if I'm missing something though.

canac: Good point. Just tested on josh.org, which doesn't have a custom thank you page, and there is only one `purchase` event, as expected.

PR #1089, opened 2024-03-01
[HS-1112780] Force sign-in after 422 response in designation editor

canac:
Would this fix the issue? If loading the content responds with a 422, we ask the user to sign in again and try again to load the content.

I wasn't able to test in staging because of conflicts with Daniel's Okta work. But I did test it locally.

https://secure.helpscout.net/conversation/2525923620/1112780?viewId=7296729

### DIFF
--- a/src/app/designationEditor/designationEditor.component.js
+++ b/src/app/designationEditor/designationEditor.component.js
@@ -10,6 +10,7 @@ import sessionEnforcerService, {
   EnforcerCallbacks,
   EnforcerModes
 } from 'common/services/session/sessionEnforcer.service'
+import sessionModalService from 'common/services/session/sessionModal.service'
 import sessionService, { Roles } from 'common/services/session/session.service'
 import designationEditorService from 'common/services/api/designationEditor.service'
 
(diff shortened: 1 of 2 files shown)


src/app/designationEditor/designationEditor.spec.js
```diff
@@ -176,6 +176,24 @@ describe('Designation Editor', function () {
...
+      $httpBackend.expectGET(designationConstants.designationEndpoint + '?designationNumber=' + designationSecurityResponse.designationNumber)
+        .respond(422, null)
+      $httpBackend.expectGET(designationConstants.designationImagesEndpoint + '?designationNumber=' + designationSecurityResponse.designationNumber)
+        .respond(422, null)
+
+      $ctrl.getDesignationContent().then(() => {
+        expect($ctrl.sessionModalService.open).toHaveBeenCalled()
+        expect($ctrl.designationContent).toBeDefined()
+        expect($ctrl.contentLoaded).toEqual(true)
+        expect($ctrl.loadingContentError).toEqual(false)
```
wrandall22, line 192: I wonder if these expectations are being called or if we need to add `done()` to this test since it is asynchronous. Did you test to find out for sure if these are being called?

wrandall22:
This certainly could fix the issue. The edge case in my mind is what happens if the login doesn't properly fix the cookies and they keep getting 422 responses. I suppose we can fix that bridge if it comes up.

canac:
> The edge case in my mind is what happens if the login doesn't properly fix the cookies and they keep getting 422 responses.

Maybe I should add logic to make sure that the login+retry only happens once.

src/app/designationEditor/designationEditor.component.js
```diff
@@ -93,17 +95,28 @@ class DesignationEditorController {
       // get designation photos
       this.designationEditorService.getPhotos(this.designationNumber)
     ]).then(responses => {
-      this.contentLoaded = true
```
canac, line 96: I removed this because `this.updateCarousel` synchronously sets `this.contentLoaded` to `false`. It made tests confusing because in the test you would expect `this.contentLoaded === true` unless you check the body of `this.updateCarousel`. But maybe it would be better to remove the `this.updateCarousel` here and rely on the fact that `this.contentLoaded` is changed in this `.then()` handler to refresh the carousels. @wrandall22 Do you have thoughts on which approach is clearer?

wrandall22: I think calling the function is clearer to me

canac:
@wrandall22 Do you want to give this another review?

canac:
@wrandall22 When I tried it locally, a successful login does not trigger a page refresh.

PR #1090, opened 2024-03-05
[no-Jira] Update analytics purchase event

canac:
Improve the `purchase` analytics event as requested by Will Gualtieri

* Set `job_id` to the `utm_term` query parameter
* Change `designation` from the string `'designation'` to the designation number
* Add the campaign code to branded checkout

### DIFF
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -555,7 +555,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $location, $timeout,
 
         // Set the transactionId in localStorage to be the one that is passed in
         sessionStorage.setItem('transactionId', currentTransactionId)
-        const jobId = new URLSearchParams($location.search()).get('utm_term') || undefined
+        const jobId = $location.search().utm_term
         const cartObject = transactionCart.items.map((cartItem) => {
           const { amount, amountWithFees } = cartItem
           purchaseTotal += amount
--- a/src/app/analytics/analytics.factory.js
+++ b/src/app/analytics/analytics.factory.js
@@ -34,9 +34,9 @@ const generateProduct = suppressErrors(function (item, additionalData = {}) {
   const sessionStorageTestName = testingTransactionName(item)
   const testingTransaction = Boolean(sessionStorageTestName &&
     window.sessionStorage.getItem(sessionStorageTestName) === 'true').toString()
-  const price = additionalData?.price || item.amount
-  const category = additionalData?.category || item.designationType
-  const name = additionalData?.name || item.displayName || undefined
+  const price = additionalData.price || item.amount
+  const category = additionalData.category || item.designationType
+  const name = additionalData.name || item.displayName || undefined
   const recurringDate = additionalData.recurringDate
     ? additionalData.recurringDate.format('MMMM D, YYYY')
     : item.giftStartDate
@@ -59,7 +59,7 @@ const generateProduct = suppressErrors(function (item, additionalData = {}) {
   }
 })
 
-const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService, sessionService) {
+const analyticsFactory = /* @ngInject */ function ($window, $location, $timeout, envService, sessionService) {
   return {
     checkoutFieldError: suppressErrors((field, error) => {
       $window.dataLayer = $window.dataLayer || []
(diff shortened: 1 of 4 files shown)


src/app/analytics/analytics.factory.js
```diff
@@ -555,6 +555,7 @@ const analyticsFactory = /* @ngInject */ function ($window, $timeout, envService
 
         // Set the transactionId in localStorage to be the one that is passed in
         sessionStorage.setItem('transactionId', currentTransactionId)
+        const jobId = new URLSearchParams($location.search()).get('utm_term')
```
wrandall22, line 558: Any sort of fallback value for this? Most of the time, `utm_term` won't be in the URL.

src/app/branded/analytics/branded-analytics.factory.spec.js
```diff
@@ -11,14 +11,18 @@ const productData = {
...
   beforeEach(angular.mock.module(module.name))
 
   const self = {}
-  beforeEach(inject((brandedAnalyticsFactory, $window) => {
+  beforeEach(inject((brandedAnalyticsFactory, $window, $location) => {
     self.brandedAnalyticsFactory = brandedAnalyticsFactory
     self.$window = $window
     self.$window.dataLayer = []
+
+    jest.spyOn($location, 'search').mockReturnValue(`?utm_term=${utmTerm}`)
```
wrandall22, line 28: May be worth having a test without this query parameter as well.

src/app/branded/analytics/branded-analytics.factory.js
```diff
@@ -37,6 +37,7 @@ function generateEcommerce (siebelTransactionId, jobId) {
       price: amountPaid,
       quantity: '1',
       recurring_date: item.giftStartDate ? item.giftStartDate.format('MMMM D, YYYY') : undefined,
+      campaign_code: item.config.CAMPAIGN_CODE || undefined,
```
wrandall22, line 40: Is `item.config` defined for branded checkout? I'm not sure.

canac: Yes https://github.com/CruGlobal/give-web/blob/d30faafd08b4a1f40463a5bb4fcc1a65eb6167f6/src/app/branded/step-1/branded-checkout-step-1.component.js#L36

src/app/analytics/analytics.factory.js
canac: It will be `undefined` if it is missing. At least when I was working with the Adswerve people on branded checkout, they said to set unavailable attributes to `undefined`.

wrandall22: According to [these docs](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get) it will return `null` rather than `undefined` when calling `get('missing_parameter)`

dr-bizz: That's correct, just confirmed it in a window

src/app/analytics/analytics.factory.spec.js
```diff
@@ -659,10 +663,11 @@ describe('analytics factory', () => {
             payment_type: 'credit card',
             purchase_number: '23032',
             campaign_code: undefined,
-            designation: 'designation',
+            designation: '0643021',
             item_brand: 'STAFF',
             processingFee: undefined,
             testing_transaction: 'false',
+            job_id: utmTerm,
```
dr-bizz, line 670: Can you add a test for when `utmTerm` isn't present?

src/app/analytics/analytics.factory.js
canac: You're right. That's fixed now.

src/app/analytics/analytics.factory.spec.js
canac: Just did. Thanks

src/app/analytics/analytics.factory.js
wrandall22: Looks like it was fixed on branded, but not standard checkout analytics file.

dr-bizz (approved):

wrandall22 (approved):

canac:
It turns out that the campaign code and the job id don't actually need to be explicitly added to the data layer because they are automatically tracked by GA.

PR #1091, opened 2024-03-08
[EP-2486] Prevent invalid payment methods from being chosen

canac:
Prevent invalid payment methods from being selected by default on step 2 of checkout.

When testing locally, I used the following change to simulate invalid payment methods:

```patch
diff --git a/src/common/services/api/order.service.js b/src/common/services/api/order.service.js
index fcbb2ad4..475cce6a 100644
--- a/src/common/services/api/order.service.js
+++ b/src/common/services/api/order.service.js
@@ -234,6 +234,7 @@ class Order {
           paymentMethods.unshift(selector.chosen)
         }
         return map(paymentMethods, paymentMethod => {
+          paymentMethod.description['payment-instrument-identification-attributes']['expiry-year'] = '2020'
           paymentMethod.description.selectAction = paymentMethod.self.uri
           if (paymentMethod.description['payment-instrument-identification-attributes']['street-address']) {
             paymentMethod.description.address =
```

https://jira.cru.org/browse/EP-2486

### DIFF
--- a/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
+++ b/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
@@ -71,13 +71,14 @@ class ExistingPaymentMethodsController {
   }
 
   selectDefaultPaymentMethod () {
-    const chosenPaymentMethod = find(this.paymentMethods, { chosen: true })
+    const paymentMethods = this.paymentMethods.filter(paymentMethod => this.validPaymentMethod(paymentMethod))
+    const chosenPaymentMethod = find(paymentMethods, { chosen: true })
     if (chosenPaymentMethod) {
       // Select the payment method previously chosen for the order
       this.selectedPaymentMethod = chosenPaymentMethod
     } else {
       // Select the first payment method
-      this.selectedPaymentMethod = this.paymentMethods[0]
+      this.selectedPaymentMethod = paymentMethods[0]
     }
     this.switchPayment()
   }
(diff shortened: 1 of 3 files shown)


wrandall22 (approved):

dr-bizz (approved):
The code looks good. Great work. Can you merge it onto staging so I can test it there?

dr-bizz (commented):
Might be good to add a test to check `selectedPaymentMethod` is `null` when all cards are invalid.

canac:
@dr-bizz I added that test and merged into staging.

wrandall22:
So this will prevent default selection; just wanted to make sure it will also prevent selecting a payment method with an invalid date? They should be forced to update the expiration date if they choose a payment method with an expired date.

canac:
@wrandall22 Correct, invalid payment methods have always been disabled https://github.com/CruGlobal/give-web/blob/27de8baafe7080a1211c5efeccecac07cd30e49f/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html#L7-L13 (note the `ng-init="expired = !$ctrl.validPaymentMethod(paymentMethod)"` on line 7 and the `ng-disabled="expired"` on line 9) The bug was that the previous code could initially set `$ctrl.selectedPaymentMethod` to an invalid payment method.

dr-bizz:
Looks good in staging

dr-bizz:
We should have done more testing with this PR. Part of that is on me. Anything with checkout should be tripled-checked before rolling into production.

PR #1093, opened 2024-03-13
[EP-2486] Prevent invalid payment methods from being chosen

canac:
Prevent invalid payment methods from being selected by default on step 2 of checkout. This follow-up PR to #1091 fixes the bug that caused the Continue button on step 2 to always be disabled.

I tested a regular checkout flow in staging to ensure that the Continue button is enabled, but I wasn't able to test with expired credit cards in staging. I tested expired credit cards locally though.

https://jira.cru.org/browse/EP-2486

### DIFF
--- a/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
+++ b/src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
@@ -129,12 +129,10 @@ class ExistingPaymentMethodsController {
   }
 
   switchPayment () {
-    if (this.selectedPaymentMethod) {
-      this.onPaymentChange({ selectedPaymentMethod: this.selectedPaymentMethod })
-      if (this.selectedPaymentMethod['bank-name']) {
-        // This is an EFT payment method so we need to remove any fee coverage
-        this.orderService.storeCoverFeeDecision(false)
-      }
+    this.onPaymentChange({ selectedPaymentMethod: this.selectedPaymentMethod })
+    if (this.selectedPaymentMethod?.['bank-name']) {
+      // This is an EFT payment method so we need to remove any fee coverage
+      this.orderService.storeCoverFeeDecision(false)
     }
   }
 }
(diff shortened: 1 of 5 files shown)


wrandall22:
Were you able to test the branded checkout flow in staging?

wrandall22 (changes requested):
Something still seems wrong with the EFT flow as the Continue button in staging is still disabled.

canac:
@wrandall22 I'm struggling with this one 😬. Thanks for the second pair of eyes on it. Branded checkout was working but you're right that continue button was disabled when there were no saved payment methods. I fixed it and the synthetic test passes now. Clearly, that logic is easy to get wrong and since we can't do component testing, I think I should move the disabled check into the controller and add a bunch of tests for it.

canac:
@wrandall22 And I somehow committed the wrong thing so now it's always enabled. You can wait a bit before reviewing again.

canac:
@wrandall22 This is now ready for review again. I tested in staging with branded checkout and in regular checkout with no payment methods, all invalid payment methods, one invalid payment method, and one invalid payment method edited to become valid (all by manually overriding the payment methods response).

wrandall22:
I noticed that the synthetic test was failing a short bit ago, and then nothing seems to have changed, but now it is working. Did you find the cause of the failing tests?

canac:
> I noticed that the synthetic test was failing a short bit ago, and then nothing seems to have changed, but now it is working. Did you find the cause of the failing tests?

I did not. It was failing to save the payment method. I was getting the same 500 error for a while for bank accounts (credit cards could save successfully). Then it went away locally and when I reran the synthetic test, it passed as well. I don't know why it was failing.

wrandall22:
Looks like it might be related to the DNS issues in the sev1 channel. It was failing to hit CCP.

wrandall22 (approved):

dr-bizz (commented):
This looks good to me, but Is there a way to test this with an expired Credit Card?

Also, one of the users who reported the error had saved cards, but the continue button was still disabled.

image (5)

canac:
@dr-bizz In the previous PR, the step 2 component was disabling the button if `$ctrl.selectedPaymentMethod` wasn't set. However, `$ctrl.selectedPaymentMethod` was only set in the existing payment methods component, not the step 2 component. So `$ctrl.selectedPaymentMethod` was always `undefined` and the continue button was always disabled. The commits after the first one fix that by setting `selectedPaymentMethod` in the step 2 component whenever the payment method changes.

dr-bizz (approved):
We should roll this out and have devs like Jeff clay (who had issues before) test it on production immediately.

PR #1095, opened 2024-03-22
Use same donation history endpoint

wrandall22:
This PR has to do with BigQuery costs. The way the current year/month historical donation endpoint works, it is hitting OBIEE once for each donation, creating several queries when there are several donations for a date range. In BigQuery, that kind of querying is cost prohibitive. The recipients historical donation endpoint queries all the data at once, returning a list of donations instead of a list of ids. The data that the current year/month view uses exists within the recipient response (though  it does not narrow down to a specific month, but only by year or by `recent` which is the last 3 months).

I believe this will be a slower call from the front-end because we are not zooming the payment method and recurring donation endpoints anymore (though these are actually currently broken in production), and instead making separate http calls for this data. This does allow us not to have to create new API endpoints on the back-end.

A benefit of this PR is that it fixes bugs that are currently in production where we are missing the payment and recurring donation information on this view. I will make a separate PR to fix the recipient donation view, which is also broken.


wrandall22:
@hlbraddock FYI, I found some extra time to start working on this.

hlbraddock:
> @hlbraddock FYI, I found some extra time to start working on this.

Very nice.

Note that, if the BI Engine cached partitioned donor table includes the donations in question, there would be no added $ cost, even with the current solution (though there is performance cost).

So, this solution comes into play, in terms of reducing $ costs, when the donations are not BI Engine cached.

src/app/profile/yourGiving/historicalView/historicalView.component.js
```diff
@@ -34,11 +40,56 @@ class HistoricalView {
       this.loadingGiftsError = true
     })
   }
+
+  isRecent (year, month) {
+    const today = moment()
+    return today.subtract(3, 'months').isBefore(moment(`${year}/${month}/01`, 'YYYY/MM/DD'))
```
canac, line 46: Do you think something like this is more intuitive? (Double-check that I got the condition right).

```suggestion
    return today.diff(moment(`${year}/${month}/01`, 'YYYY/MM/DD'), 'months') <= 3;
```

src/app/profile/yourGiving/historicalView/historicalView.component.js
```diff
@@ -34,11 +40,56 @@ class HistoricalView {
...
+  }
+
+  parseHistoricalGifts (historicalGifts, year, month) {
+    if (!historicalGifts.length || !historicalGifts[0].donations) {
+      return historicalGifts
+    }
+
+    const filteredList = historicalGifts.filter((historicalGift) => {
```
canac, line 54:
```suggestion
    const filteredGifts = historicalGifts.filter((historicalGift) => {
```

src/app/profile/yourGiving/historicalView/historicalView.component.js
```diff
@@ -34,11 +40,56 @@ class HistoricalView {
...
+      historicalGift.donations = historicalGift.donations.filter((donation) => {
```
canac, line 55: This mutates the `donations` attributes on `historicalGifts` items. But that should be OK because it's not used again after being passed to `parseHistoricalGifts`, right?

src/app/profile/yourGiving/historicalView/historicalView.component.js
```diff
@@ -34,11 +40,56 @@ class HistoricalView {
...
+        const transactionDate = donation['historical-donation-line']['transaction-date']
+        const momentDate = moment(transactionDate['display-value'], 'YYYY/MM/DD')
+
+        return momentDate.year() === year && (momentDate.month() + 1) === month
+      })
+      return historicalGift.donations.length > 0
+    })
+
+    return filteredList.flatMap((donationSummary) => {
+      if (donationSummary.donations) {
```
canac, line 65: This condition is always true right? If `donations` isn't set, I believe we'll get an exception on line 55.

canac (commented):
I wasn't able to test it, but everything looks reasonable.

src/app/profile/yourGiving/historicalView/historicalView.component.js
wrandall22: I believe that to be true, though can make it immutable if it is better practice.

src/app/profile/yourGiving/historicalView/historicalView.component.js
wrandall22: Yeah, I think so. Probably this is left over from my TDD work as I had both in place in parallel.

src/app/profile/yourGiving/historicalView/historicalView.component.js
canac: In general, I try to avoid mutating inputs to avoid surprises later on if the caller didn't expect the method to modify the input.

canac: It's not the `filter` that is mutating, it's the `historicalGift.donations = ...` assignment. That's because `historicalGifts[0]` and `filteredGifts[0]` refer to the same JavaScript object. You'll have to do something like `historicalGifts.map(historicalGift => ({ ...historicalGift, donations: historicalGift.donations.filter(...) })).filter(historicalGift => historicalGift.donations.length > 0)` to avoid mutating `historicalGifts`' donations.

wrandall22: [cloneDeep](https://lodash.com/docs/#cloneDeep) seems to do the trick based on JSON outputs. Let me know what you think.

wrandall22:
This is testable in stage now

src/app/profile/yourGiving/historicalView/historicalView.component.js
canac: The second `cloneDeep` (`cloneDeep(historicalGift.donations)`) shouldn't be necessary. Other than that it looks good!

canac (approved):
I had trouble logging into staging with Okta, but it seems to work locally

wrandall22:
Hmm, appears to be an issue with the payment method data, perhaps from the database refresh from prod. [Rollbar](https://app.rollbar.com/a/Cru/fix/item/give-ep/19152#detail)

PR #1096, opened 2024-03-22
EP-2488 - Fix payment link

wrandall22:
When we migrated to EP 8.1, the link got renamed from `payment-method-link` to `payment-instrument-link`, but that change did not get updated on the front-end. This only fixes the recipient view of historical donations, not the monthly view. Either monthly will have to be fixed on the back-end, or #1095 will fix it.

### DIFF
--- a/src/app/profile/yourGiving/recipientView/recipientGift/recipientGift.component.js
+++ b/src/app/profile/yourGiving/recipientView/recipientGift/recipientGift.component.js
@@ -37,7 +37,7 @@ class RecipientGift {
 
       const paymentMethodRequests = []; const paymentMethodUris = []
       angular.forEach(this.recipient.donations, (donation) => {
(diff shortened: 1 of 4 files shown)


wrandall22:
Doesn't appear to be deploying to stage properly

canac (approved):
Code looks good. What differences would you expect to see in the frontend after this fix?

wrandall22:
When it properly deploys (hence #1097), when you go to the recipient view historical donations and drill into a recipient you should see the payment details filled out.

dr-bizz (approved):

wrandall22:
Got the deploy to stage done if you want to look, it looks good to me there.

canac:
Yep, I see the payment info in staging now. Great work with the fix!

PR #1102, opened 2024-08-01
Recaptcha improvements

wrandall22:
The main part of this PR is adding Recaptcha to the checkout summary submit button, but there are some tweaks from the previous PR as well and minor code improvements.

### DIFF
--- a/src/app/checkout/checkout.tpl.html
+++ b/src/app/checkout/checkout.tpl.html
@@ -82,8 +82,8 @@
         <checkout-cart-summary cart-data="$ctrl.cartData"
                                show-submit-btn="$ctrl.checkoutStep === 'review'"
                                enable-submit-btn="$ctrl.enableSubmitOrderBtn"
-                               on-submit="$ctrl.submittedOrder = true"
-                               submitting-order="$ctrl.submittingOrder">
+                               submitting-order="$ctrl.submittingOrder"
+                               component-reference="$ctrl.selfReference">
         </checkout-cart-summary>
       </div>
       <div class="col-md-4 hidden-xs hidden-sm pull-right">
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -204,7 +204,7 @@
               <div class="checkout-cta pull-right">
                 <recaptcha-wrapper
                            ng-if="!$ctrl.isBranded"
-                           action="'submit'"
+                           action="'submit_gift'"
                            on-success="$ctrl.submitOrderInternal"
                            on-failure="$ctrl.handleRecaptchaFailure"
                            component-instance="$ctrl.selfReference"
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -203,7 +203,6 @@
               </div>
               <div class="checkout-cta pull-right">
                 <recaptcha-wrapper
-                           ng-if="!$ctrl.isBranded"
                            action="'submit_gift'"
                            on-success="$ctrl.submitOrderInternal"
                            on-failure="$ctrl.handleRecaptchaFailure"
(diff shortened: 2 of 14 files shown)


wrandall22:
Tests on staging are showing sometimes one button will work while the other will not.

dr-bizz:
I assume this isn't ready for a review as it's a draft, so I will hold off reviewing it.

With your issue about it stopping the bottom submit button. I wonder if you need 2 separate Recaptchas.

wrandall22:
I think what I need to do is rip out this library and add recaptcha to the page myself and have more flexibility when it comes to calling recaptcha. This should also help me resolve the issue on branded checkout of potential conflicting recaptchas.

src/common/components/Recaptcha/RecaptchaWrapper.tsx
```diff
@@ -21,6 +26,17 @@ interface RecaptchaWrapperProps {
   $log: any
 }
 
+export const findExistingScript = (): boolean => {
+  const scripts = document.getElementsByTagName('script')
+
+  for (let script of scripts) {
+    if (script.src.startsWith('https://www.google.com/recaptcha/api.js?render=')) {
```
wrandall22, line 33: May be worth doing a more general lookup to catch things like `//www.google.com/recaptcha/api.js?render=`, but this is working on all the current implementations.

wrandall22:
This is ready for review now.

src/common/components/Recaptcha/Recaptcha.tsx
```diff
@@ -31,63 +36,74 @@ export const Recaptcha = ({
...
+  $log,
+  recaptchaKey
 }: RecaptchaProps): JSX.Element => {
-  const { executeRecaptcha } = useGoogleReCaptcha();
+
+  const [ready, setReady] = useState(false)
+  const [grecaptcha, setGrecaptcha] = useState(window.grecaptcha)
+  let timesIntervalRun = 0
+
+  useEffect(() => {
```
dr-bizz, line 47: Should there be a failsafe, just in case an error happens with recaptcha, and `window.grecaptcha` doesn't get defined?

Like after so many seconds, we set it to ready? But I guess maybe not, as the bots could just wait.

src/common/components/Recaptcha/RecaptchaWrapper.tsx
```diff
@@ -1,18 +1,23 @@
...
+}
+
 interface RecaptchaWrapperProps {
   action: string
   onSuccess: (componentInstance: any) => void
   onFailure: (componentInstance: any) => void
   componentInstance: any
   buttonId: string
-  buttonType?: 'submit' | 'reset' | 'button' | undefined
+  buttonType?: 'submit' | 'reset' | 'button'
```
dr-bizz, line 20: This isn't a biggy, but since the buttonType prop is used in multiple files, it might be good to turn it into a Enum.

src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -213,16 +212,6 @@
                            button-classes="'btn btn-primary btn-lg btn-block'"
                            button-disabled="!$ctrl.canSubmitOrder()"
                            button-label="'SUBMIT_GIFT'"></recaptcha-wrapper>
-                  <button
```
dr-bizz, line 216: I assume branded checkout is now using recaptcha as well?

src/app/checkout/checkout.tpl.html
```diff
@@ -82,8 +82,8 @@ <h3 class="panel-name inline-block" translate>Secure Checkout</h3>
         <checkout-cart-summary cart-data="$ctrl.cartData"
                                show-submit-btn="$ctrl.checkoutStep === 'review'"
                                enable-submit-btn="$ctrl.enableSubmitOrderBtn"
-                               on-submit="$ctrl.submittedOrder = true"
-                               submitting-order="$ctrl.submittingOrder">
+                               submitting-order="$ctrl.submittingOrder"
+                               component-reference="$ctrl.selfReference">
```
dr-bizz, line 86: I get why you are passng down the reference, but I don't understand how we move the user forward through to the next stage in the checkout.

dr-bizz (approved):
This looks good to me. I saw no issues with this PR. I had some questions, but that was more for me to understand.

src/common/components/Recaptcha/Recaptcha.tsx
wrandall22: It could be worth it; so far in my testing I haven't run into an issue with this, but we also don't want to prevent donors from giving. If we do a failsafe, we'd just call `onSuccess` instead of setting `ready` to true.

src/app/checkout/step-3/step-3.tpl.html
wrandall22: It will after this PR, yes.

src/app/checkout/checkout.tpl.html
wrandall22: The `cart-summary` component is emitting an event on the `$rootScope` of `componentReference` and the `step-3` component is listening for that event and calling the submit logic.

PR #1103, opened 2024-08-06
Recaptcha improvements

wrandall22:
This PR makes Recaptcha work on branded checkout. It won't work until https://github.com/CruGlobal/cortex_gateway/pull/148 and https://github.com/CruGlobal/cru-terraform/pull/6413 are fully applied.


caleballdrin (approved):
Looks good!

wrandall22:
Thanks for reviewing. Added 1 more commit if you could take a quick look as well.

PR #1105, opened 2024-10-16
[no-Jira] Upgrade jest and babel

canac:
Fixes [CVE-2023-45133 in @babel/traverse](https://github.com/CruGlobal/give-web/security/dependabot/127)

Briefly scanning through the list of Dependabot, this seems to be the main urgent issue that isn't just a ReDOS or only exploitable against the webpack dev server.


dr-bizz (approved):
Tests and builds are looking great. Testing on staging also.

PR #1106, opened 2024-10-21
Revert "[no-Jira] Upgrade jest and babel"

dr-bizz:
Reverts CruGlobal/give-web#1105

#### Issue:
Friday and over the weekend, we had multiple people say they couldn't log on to give.cru.org. On inspecting, when users log on, the login request comes back as `200`, but the page doesn't react and log you in on the front end. 

As this was the only change we made on the frontend, it makes sense to revert it to see if this caused the issue and then investigate futher


canac:
Interesting. Thanks for reverting. It will be hard to test further in staging since all the Okta stuff is in staging right now.

dr-bizz:
I'm still getting the error on my Chrome, but not on other browsers. I will talk to the people who reported the issue to see if it has fixed it for them. Or if there is another issue.

PR #1109, opened 2024-10-28
Updates hideCvv flag for edit payment methods and new payment methods.

wjames111:
OKR: Implement real-time CVV validation on all transactions for Give App

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,7 @@ Note: For session cookies to work correctly, add the below entries to your hosts
 
 - `yarn test` to run Jest tests
 - `yarn lint` to run StandardJS linter
+- `yarn lint:styles` to run StyleLint linter _(run with `--fix` to fix issues)._
 - `yarn build` to generate minified output files. These files are output to `/dist`. `common.js` must be included before any of the other JS files.
 - `yarn build:analyze` to open a visualization of bundle sizes after building
 
--- a/src/app/designationEditor/designationEditor.scss
+++ b/src/app/designationEditor/designationEditor.scss
@@ -1,3 +1,7 @@
 .designation-editor-paragraph-text{
   min-height: 90px;
 }
+
+.ta-editor {
+  height: auto;
+}
--- a/src/assets/scss/styles.scss
+++ b/src/assets/scss/styles.scss
@@ -32,3 +32,6 @@
 @import 'nav-cart';
 @import 'carousel';
 @import 'directives/creditCardCvv';
+
+// TODO: Remove this manual import when we figure out why styles imported in components are not being added to the page
+@import '../../app/designationEditor/designationEditor';


wrandall22:
So, there are a few things.

1) You should probably make the changes in `existingPaymentMethods` instead of in `paymentMethodDisplay` since the first is only used in checkout and the latter is used in several places we don't want to capture cvv.
2) The validation is happening, but is not preventing the user from moving forward, which means we need to hook into the continue button on checkout step 2 and ensure that the validation passes there before moving forward.
3) The Angular way of re-using the cvv template/controller code is to create a [directive](https://docs.angularjs.org/guide/directive) and put it in both places. We have several directives in the codebase already which you can reference (even `creditCardNumber` has one). This would be a way to not copy/paste code and maintain it in two places.

wrandall22:
The CVV is the only thing being entered, so is the only thing being validated in this situation. That validation can live in the directive.

dr-bizz:
Great work on this PR!

Just a note: before merging with staging, can you run `yarn lint:write`? This will ensure the lint issues are fixed or bring attention to code that needs to be changed. This way, your changes on staging will be deployed to the staging site.

dr-bizz:
It seems like it wasn't just you, there were some other changes from other PRs that caused staging to fail being deployed.

wrandall22:
You can turn it into a draft PR until it's ready if you want.

dr-bizz:
It's all good. TThis is where amplify previews would come in handy

wrandall22:
Hey Will, thanks for working on this. I was more thinking of putting this part of the HTML in the directive:

```
<div class="form-group" ng-class="{'has-error': ($ctrl.creditCardPaymentForm.securityCode | showErrors), 'is-required': !$ctrl.paymentMethod}">
          <label translate>{{'SEC_CODE'}}</label>
          <input type="text"
                 name="securityCode"
                 autocomplete="cc-csc"
                 class="form-control form-control-subtle"
                 ng-model="$ctrl.creditCardPayment.securityCode"
                 ng-required="!$ctrl.paymentMethod || $ctrl.creditCardPayment.cardNumber"
                 ng-attr-placeholder="{{$ctrl.paymentMethod && !$ctrl.creditCardPayment.cardNumber ? '***' : ''}}">
          <div role="alert" ng-messages="$ctrl.creditCardPaymentForm.securityCode.$error" ng-if="($ctrl.creditCardPaymentForm.securityCode | showErrors)">
            <div class="help-block" ng-message="required" translate>{{'CARD_SEC_CODE_ERROR'}}</div>
            <div class="help-block" ng-message="minLength" translate>{{'MIN_LENGTH_CARD_SEC_CODE'}}</div>
            <div class="help-block" ng-message="maxLength" translate>{{'MAX_LENGTH_CARD_SEC_CODE'}}</div>
            <div class="help-block" ng-message="cardTypeLength" ng-init="isAmex = $ctrl.cardInfo.type($ctrl.creditCardPayment.cardNumber) === 'American Express'">
              <translate ng-if="!isAmex">{{'LOCATION_OF_CODE_OTHER'}}</translate>
              <translate ng-if="isAmex">{{'LOCATION_OF_CODE_AMEX'}}</translate>
            </div>
          </div>
        </div>
```

Then using it in both `creditCardForm` and `existingPaymentMethods`

wrandall22:
Yep, this is the direction I was thinking.

src/app/checkout/step-2/step-2.component.js
```diff
@@ -117,6 +117,9 @@ class Step2Controller {
   }
 
   getContinueDisabled () {
+    if (typeof this.isCvvValid !== 'undefined' && !this.isCvvValid) {
+      return true
+    }
```
wrandall22, line 122: This shouldn't be done all the time, only when we're actually using a credit card.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -52,6 +55,14 @@ class ExistingPaymentMethodsController {
     }
   }
 
+  addCustomValidators () {
+    this.$scope.$watch('$ctrl.creditCardPaymentForm.securityCode.$viewValue', (number) => {
+      this.creditCardPaymentForm.securityCode.$validators.minLength = cruPayments.creditCard.cvv.validate.minLength /* eslint-disable-line no-mixed-operators */
```
wrandall22, line 60: I don't think you need this comment here

src/common/directives/creditCardCvv.directive.js
```diff
@@ -0,0 +1,35 @@
...
+    <div class="help-block" ng-message="minLength" translate>{{'MIN_LENGTH_CARD_SEC_CODE'}}</div>
+    <div class="help-block" ng-message="maxLength" translate>{{'MAX_LENGTH_CARD_SEC_CODE'}}</div>
+    <div class="help-block" ng-message="cardTypeLength" ng-init="isAmex = $ctrl.cardInfo.type($ctrl.creditCardPayment.cardNumber) === 'American Express'">
+      <translate ng-if="!isAmex">{{'LOCATION_OF_CODE_OTHER'}}</translate>
+      <translate ng-if="isAmex">{{'LOCATION_OF_CODE_AMEX'}}</translate>
+    </div>
+  </div>
+</div>`
+
+const creditCardCvv = /* @ngInject */ function () {
```
wrandall22, line 25: Let's use an arrow function here

wrandall22 (changes requested):
I got part way through the review, this is a good start. I'm going to play around a little bit with the directive locally as well.

branded-checkout.html
```diff
@@ -12,7 +12,7 @@
 <body>
 
 <branded-checkout designation-number="2294554" show-cover-fees="true" ng-cloak></branded-checkout>
-<script src="branded-checkout.v2.js"></script>
+<script src="dev.v2.js"></script>
```
wrandall22, line 15: We'll want to change this back before merge.

src/app/checkout/step-2/step-2.component.spec.js
```diff
@@ -308,6 +308,53 @@ describe('checkout', () => {
...
+        jest.spyOn(self.controller.brandedAnalyticsFactory, 'savePaymentType')
+        self.controller.handlePaymentChange({'card-type': 'visa'})
+        self.controller.isCvvValid = false
+        expect(self.controller.getContinueDisabled()).toBe(true)
+      })
+
+      it('should return false when cvv is invalid and new payment type is bankAccount', () => {
+        jest.spyOn(self.controller.brandedAnalyticsFactory, 'savePaymentType')
+        self.controller.handlePaymentChange({'account-type': 'checking'})
+        self.controller.isCvvValid = false
```
wrandall22, line 338: Will this be false or undefined? Or null?

src/app/checkout/step-2/step-2.component.spec.js
```diff
@@ -308,6 +308,53 @@ describe('checkout', () => {
...
+
+      it('should return false when cvv is invalid and new payment type is credit card', () => {
+        jest.spyOn(self.controller.brandedAnalyticsFactory, 'savePaymentType')
+        self.controller.handlePaymentChange({'card-type': 'visa'})
+        self.controller.isCvvValid = false
+        expect(self.controller.getContinueDisabled()).toBe(true)
+      })
+
+      it('should return false when cvv is invalid and new payment type is bankAccount', () => {
+        jest.spyOn(self.controller.brandedAnalyticsFactory, 'savePaymentType')
```
wrandall22, line 336: Don't need this spy

src/app/checkout/step-2/step-2.component.spec.js
```diff
@@ -308,6 +308,53 @@ describe('checkout', () => {
...
+      it('should return false when cvv is invalid and card type is not credit card', () => {
+        self.controller.handleExistingPaymentLoading(true, true)
+        self.controller.isCvvValid = false
+        self.controller.handlePaymentChange({'account-type': 'checking'})
+        
+        expect(self.controller.getContinueDisabled()).toBe(false)
+      })
+
+      it('should return false when cvv is invalid and new payment type is credit card', () => {
+        jest.spyOn(self.controller.brandedAnalyticsFactory, 'savePaymentType')
```
wrandall22, line 329: Don't need this spy

src/app/checkout/step-2/step-2.component.js
```diff
@@ -117,6 +117,9 @@ class Step2Controller {
   }
 
   getContinueDisabled () {
+    if (typeof this.isCvvValid !== 'undefined' && !this.isCvvValid) {
```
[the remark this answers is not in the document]
wrandall22: You can double check what the value would be in the dev tools; let me know if you want help seeing how to do that for Angular apps.

src/app/checkout/step-2/step-2.component.spec.js
wrandall22: Gotcha, typically it is best to have a test for each possible path of the conditional, so you might want to add one more for that at least.

wrandall22 (changes requested):
Still digging, but a few more comments in the meantime

wrandall22:
Almost all of the functionality seems to be working well. One edge case I have found is that if you add a credit card on the existing payment methods step, you have to fill in CVV (as expected). Then you have to fill it out again before you can move on because the new field is empty. In this situation, we should be able to pull it from the `sessionStorage`, comparing the payment instrument id to get the right one (in case there are multiple).

src/common/directives/creditCardCvv.directive.js
```diff
@@ -0,0 +1,35 @@
...
+
+const template =
+`<div class="form-group" ng-class="{'has-error': ($ctrl.creditCardPaymentForm.securityCode | showErrors), 'is-required': !$ctrl.paymentMethod}">
+  <label translate>{{'SEC_CODE'}}</label>
+  <input type="text" 
+    name="securityCode"
+    autocomplete="cc-csc"
+    class="form-control form-control-subtle"
+    ng-model="$ctrl.creditCardPayment.securityCode"
+    ng-required="!$ctrl.paymentMethod || $ctrl.creditCardPayment.cardNumber"
```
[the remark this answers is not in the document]
wrandall22: Yes, I think the `ng-if` that you added should be sufficient to not need to change this.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html
```diff
@@ -11,7 +11,14 @@
         <button class="btn btn-xs btn-link" ng-click="$ctrl.openPaymentMethodFormModal(paymentMethod)" ng-if="paymentMethod['card-type']" translate>edit</button>
       </label>
     </div>
-
+    <div ng-if="$ctrl.selectedPaymentMethod['card-type']" class="row">
+      <div class="col-sm-4">
+        <form novalidate name="$ctrl.creditCardPaymentForm">
+          <credit-card-cvv ></credit-card-cvv>
```
wrandall22, line 17: It doesn't seem like the cvv is getting stored in sessionStorage here like it does in other cases, which means it probably isn't getting sent to the server properly. This is due to the unsetting of the cvv in `selectPayment` but never re-setting it anywhere else (e.g. on hitting continue)

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.spec.js
```diff
@@ -329,6 +343,49 @@ describe('checkout', () => {
           expect(self.controller.onPaymentChange).toHaveBeenCalledWith({ selectedPaymentMethod: undefined })
           expect(self.controller.orderService.storeCoverFeeDecision).not.toHaveBeenCalled()
         })
+
+        it('should reset securityCode viewValue on switch payment', () => {
```
wrandall22, line 347:
```suggestion
        it('should reset securityCode viewValue', () => {
```

`on switch payment` is implied based on where it is in the code.

src/app/checkout/step-2/step-2.component.spec.js
```diff
@@ -308,6 +308,77 @@ describe('checkout', () => {
...
+
+      it('should return true when cvv validity is false and card type is credit card', () => {
+        self.controller.handleExistingPaymentLoading(true, true)
+        self.controller.isCvvValid = false
+        self.controller.handlePaymentChange({'card-type': 'visa'})
+        
+        expect(self.controller.getContinueDisabled()).toBe(true)
+      })
+
+      it('should return true when cvv validity is true and card type is credit card', () => {
```
wrandall22, line 320: I think a more natural language on some of these tests would be nicer. Instead of `cvv validity is true`, `cvv is valid` is more natural. This way of phrasing works fine for `undefined`, but I think we should be more natural on the true/false options.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.spec.js
```diff
@@ -329,6 +343,49 @@ describe('checkout', () => {
...
+          self.controller.switchPayment()
+
+          expect(self.controller.creditCardPaymentForm.securityCode.$setViewValue).toHaveBeenCalledWith('')
+          expect(self.controller.creditCardPaymentForm.securityCode.$render).toHaveBeenCalled()
+        })
+      })
+
+      describe('addCvvValidators', () => {
+        it('should add validator functions to creditCardPaymentForm.securityCode', () => {
+          self.controller.addCvvValidators()
```
wrandall22, line 359: Since this function is just adding a watch on the scope, the way this test is built feels a bit flaky. You aren't changing `$ctrl.creditCardPaymentForm.securityCode.$viewValue` as part of this test, which means you are relying on some other test in this file, or possibly the initialization of the `creditCardPaymentForm` to be doing it for you, which may run earlier or later than this test. This test should be fully self-contained. Maybe we need to reset it somehow first then change the value then expect all within this test.

wrandall22 (changes requested):

src/app/checkout/step-2/step-2.component.js
```diff
@@ -117,6 +117,9 @@ class Step2Controller {
   }
 
   getContinueDisabled () {
```
dr-bizz, line 119: I know you didn't name it, but should we change this function name to `isContinueDisabled()`?

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -52,6 +55,14 @@ class ExistingPaymentMethodsController {
     }
   }
 
+  addCvvValidators () {
+    this.$scope.$watch('$ctrl.creditCardPaymentForm.securityCode.$viewValue', (number) => {
+      this.creditCardPaymentForm.securityCode.$validators.minLength = cruPayments.creditCard.cvv.validate.minLength
```
dr-bizz, line 60: This is causing an error on localhost. `securityCode` is undefined. This might work staging, butI have nly tested on local.

src/common/components/paymentMethods/paymentMethodForm/paymentMethodForm.modal.tpl.html
```diff
@@ -49,4 +49,3 @@ <h3 class="text-center">
     </div>
   </div>
 </div>
```
dr-bizz, line 51: Is there much point of changing this file?

src/app/checkout/step-2/step-2.component.spec.js
```diff
@@ -308,6 +308,77 @@ describe('checkout', () => {
...
+        expect(self.controller.getContinueDisabled()).toBe(false)
+      })
+
+      it('should return false when cvv validity is true and card type is not credit card', () => {
+        self.controller.handleExistingPaymentLoading(true, true)
+        self.controller.isCvvValid = true
+        self.controller.handlePaymentChange({'account-type': 'checking'})
+        
+        expect(self.controller.getContinueDisabled()).toBe(false)
+      })
```
dr-bizz, line 358: You could wrap all these in a description `card type is credit card` or `card type is not credit card`

dr-bizz (changes requested):
I found an error while testing locally, but the code is looking great!
I also found some of things, but there are less urgent.

Screenshot 2024-11-25 at 8 31 28 AM

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -130,6 +152,22 @@ class ExistingPaymentMethodsController {
 
   switchPayment () {
     this.onPaymentChange({ selectedPaymentMethod: this.selectedPaymentMethod })
+    if (this.selectedPaymentMethod?.['card-type'] && this.creditCardPaymentForm?.securityCode) {
+      const selectedUri = this.selectedPaymentMethod.self.uri
+      const storage = JSON.parse(this.sessionStorage.getItem('storedCvvs'))
+      const getSelectedCvv = storage ? storage[Object.keys(storage).filter((item) => item === selectedUri)] : false
```
wrandall22, line 158: If you use `''` instead of `false` you don't need the conditional below.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.spec.js
```diff
@@ -329,6 +345,79 @@ describe('checkout', () => {
...
+          self.controller.selectedPaymentMethod = { 'card-type': 'Visa', self: { type: 'cru.creditcards.named-credit-card', uri: 'selected uri' }, selectAction: 'some uri' }
+          self.controller.switchPayment()
+
+          expect(self.controller.creditCardPaymentForm.securityCode.$setViewValue).toHaveBeenCalledWith('')
+          expect(self.controller.creditCardPaymentForm.securityCode.$render).toHaveBeenCalled()
+        })
+
+        it('should add securityCode viewValue from sessionStorage', () => {
+          self.controller.creditCardPaymentForm.securityCode.$viewValue = '123'
+          self.controller.selectedPaymentMethod = { 'card-type': 'Visa', self: { type: 'cru.creditcards.named-credit-card', uri: '/paymentmethods/crugive/giydsnjqgi=' }, selectAction: 'some uri' }
```
wrandall22, line 360: This line is a bit too long, let's break it up to multiple lines.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.tpl.html
wrandall22: bump

dr-bizz (commented):
Thank you for thinking the error on the checkout. I have review the code again and made some comments, but they are only small things.

I did test the security code on the checkout and found it didn't matter what value I added, it always allowed me to continue. Should that be the case, or are you just taking the input on step 2, and then checking it when the user submits the checkout?

src/app/checkout/step-2/step-2.tpl.html
```diff
@@ -31,7 +31,9 @@
       default-payment-type="$ctrl.defaultPaymentType"
       hide-payment-type-options="$ctrl.hidePaymentTypeOptions"
       cart-data="$ctrl.cartData"
-      branded-checkout-item="$ctrl.brandedCheckoutItem">
+      branded-checkout-item="$ctrl.brandedCheckoutItem"
+      enable-continue="$ctrl.enableContinue($event)"
+      >
```
wrandall22, line 36: Let's put this back on the same line like before.

src/app/checkout/step-2/step-2.component.js
```diff
@@ -109,14 +110,19 @@ class Step2Controller {
       this.changeStep({ newStep: 'review' })
       this.onStateChange({ state: 'submitted' })
       this.paymentFormState = 'success'
+    } else if ($event.state === 'submitted') {
+      // this.orderService.storeCardSecurityCode(this.selectedPaymentMethod.cvv, this.selectedPaymentMethod.self.uri)
```
wrandall22, line 114: Did you mean to uncomment this?

wrandall22 (changes requested):
A few more things (I think I finally made it through all of the code at least)

dr-bizz (dismissed):
I am getting this error message when trying to continue to step 3.
```
TypeError: Cannot read properties of undefined (reading 'cvv')
at Step2Controller.onPaymentFormStateChange (step-2.component.js:114:74)
at Step2Controller.submit (step-2.component.js:75:10)
```


I also get an error message when trying to submit the checkout.
```
{"error":"Invalid Data"}
```

In my incognito window, I get an error that prevents me from moving from Step 2 to Step 3. But this could just be my local.
```
Error tokenizing credit card 
```

wrandall22:
`Error tokenizing credit card` is due to you not using the Referer Control extension properly.

`{"error":"Invalid Data"}` is due to the work being done in staging for recaptcha enterprise.

dr-bizz:
> `Error tokenizing credit card` is due to you not using the Referer Control extension properly.
> 
> `{"error":"Invalid Data"}` is due to the work being done in staging for recaptcha enterprise.

This makes sense. Ignore that comment.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.spec.js
```diff
@@ -364,13 +364,13 @@ describe('checkout', () => {
...
               type: 'cru.creditcards.named-credit-card', 
-              uri: '/paymentmethods/crugive/giydsnjqgi=' 
+              uri: '/paymentinstruments/orders/crugive/hfsdoylfhbswk…4tmljug5qtqllcg44wcljvhezgcntcmvtdeojwgy=' 
             }, 
             selectAction: 'some uri' 
           }
           self.$window.sessionStorage.setItem(
             'storedCvvs', 
-              '{"/paymentmethods/crugive/giydsnjqgi=":"456","/paymentmethods/crugive/giydsnjqgy=":"321"}'
+              '{"/paymentinstruments/orders/crugive/hfsdoylfhbswk…4tmljug5qtqllcg44wcljvhezgcntcmvtdeojwgy=":"456"}'
```
wrandall22: Note that both of these show possibilities of reality.

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -130,6 +154,15 @@ class ExistingPaymentMethodsController {
 
   switchPayment () {
     this.onPaymentChange({ selectedPaymentMethod: this.selectedPaymentMethod })
+    if (this.selectedPaymentMethod?.['card-type'] && this.creditCardPaymentForm?.securityCode) {
+      const selectedUri = this.selectedPaymentMethod.self.uri
+      const storage = JSON.parse(this.sessionStorage.getItem('storedCvvs'))
```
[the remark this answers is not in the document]
wrandall22: It seems to change when it stops using the payment method id in the cart and starts using the payment method id itself. So, when it is persisted and we go back, it seems to grab the payment method id. From my testing, `cvv` is always the last one entered, which would resolve the initial case we were trying to solve here (a cvv showing up in the box right after editing a payment method).

wrandall22:
Something seems to have regressed on the display of adding a new payment method and the optional cover fees area.

Prod:
Screenshot 2024-12-12 at 1 22 45 PM

This branch:
Screenshot 2024-12-12 at 1 21 10 PM

wrandall22:
@rguinee I'd be curious to hear your design perspective on a good way of showing the security code here as well. I'm thinking inline label/input instead of stacked and not quite so tall.

rguinee:
> @rguinee I'd be curious to hear your design perspective on a good way of showing the security code here as well. I'm thinking inline label/input instead of stacked and not quite so tall.

Are we requiring they enter it again on a saved card for security purposes?

wrandall22:
Correct

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
```diff
@@ -130,6 +156,13 @@ class ExistingPaymentMethodsController {
 
   switchPayment () {
     this.onPaymentChange({ selectedPaymentMethod: this.selectedPaymentMethod })
+    if (this.selectedPaymentMethod?.['card-type'] && this.creditCardPaymentForm?.securityCode) {
+      // Set cvv from session storage
+      const storage = JSON.parse(this.sessionStorage.getItem('cvv')) || ''
+      this.creditCardPaymentForm.securityCode.$setViewValue(storage)
+      this.creditCardPaymentForm.securityCode.$render()
+      this.sessionStorage.removeItem('cvv')
```
wrandall22, line 164: We don't want to clear this from session storage here, as this is what is sent to the server.

rguinee:
Screenshot 2024-12-12 at 1 57 57 PM
Agree on your instinct @wrandall22. There are several other recommendations I'd have for this like moving "Add a payment method" into the list of Your Payment Methods.

wrandall22:
Do you envision that being below the selected payment method or the full list? I think the former.

rguinee:
Exactly @wrandall22. They could have multiple card payment methods, so we would show below the selected option the sec code verification input. There are other considerations to make here. Is it validated when 3 or 4 digits are entered and can show success or error response? Does it have it be submitted and if incorrect the errors appear after page reloads?

wrandall22:
The validation is there and it shows in real time, not after submit (unless the back-end finds issues with the card verification with TSYS)

src/app/checkout/step-2/existingPaymentMethods/existingPaymentMethods.component.js
wrandall22: In my testing, it cleared right after I edited the existing payment method and entering the CVV there. It got copied to the text box but then removed from session storage.

wrandall22: Ah, I see what you're saying. It does get added back after hitting continue. We need to test that on the single step BCO branch as well to see how that behaves.

wrandall22: Actually, the one step BCO won't have this issue due to the user being logged out and not being able to go back from a review step to edit, the 2 step BCO will, so I can try testing there.

wrandall22: So, it looks like `switchPayment` is called when hitting `Continue` on 2 step branded checkout and when hitting `Submit` on single step branded checkout. So this will definitely need to change in some way.

src/common/directives/creditCardCvv.directive.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.give-modal-content, .branded-checkout {
+	.credit-card-cvv-label {
+		text-transform: uppercase;
+		font-weight: 500;
+		letter-spacing: 1px;
+		font-size: 14px;
+		display: block;
+    margin-bottom: 4px;
```
wrandall22, line 38: Whitespace

src/common/directives/creditCardCvv.directive.scss
wrandall22, line 1: The spacing in scss files should be 2 spaces, not tabs.

src/common/directives/creditCardCvv.directive.scss
```diff
@@ -0,0 +1,48 @@
...
+    .credit-card-cvv-help-block {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.give-modal-content, .branded-checkout {
+  .credit-card-cvv-label {
+    text-transform: uppercase;
+    font-weight: 500;
```
wrandall22, line 34: On my local, this is being overwritten by the bold above.

src/common/directives/creditCardCvv.directive.scss
```diff
@@ -0,0 +1,48 @@
...
+    letter-spacing: 1px;
+    font-size: 14px;
+    display: block;
+    margin-bottom: 4px;
+    &::after {
+      content: '*';
```
wrandall22, line 40: On my local, this is being overwritten by the `:` above.

wrandall22 (changes requested):

wrandall22:
The staging deploy never ran I think after your latest changes. However, after running it, there is still some improper overriding going on:

Screenshot 2025-01-06 at 10 41 04 AM

wrandall22:
That is why local is only good for testing up to a point. This is how the overrides are working in stage:

Screenshot 2025-01-06 at 10 55 29 AM

wrandall22:
Screenshot 2025-01-07 at 1 43 33 PM

[Testing page](https://give-dev-cloud.cru.org/testing-grounds/test-v3.html)

wrandall22:
I found the CSS that is making it happen on the same line, and it is within AEM. It is a flex display on the `row` class. This CSS is only present on dev, not stage, which is why we're not seeing it on the other page.

wrandall22:
Adding a `display: block` to `branded-checkout .row` would override this flex.

src/common/directives/creditCardCvv.directive.scss
```diff
@@ -46,3 +46,9 @@
     }
   }
 }
+
+.branded-checkout {
```
wrandall22, line 50:
```suggestion
branded-checkout
```
I think this might go better in the `branded-checkout.scss` file.

src/assets/scss/branded-checkout.scss
```diff
@@ -16,6 +16,9 @@ branded-checkout{
   line-height: $line-height-base;
   color: $text-color;
   background-color: $body-bg;
-
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+  
+  .existing-payment-method .row {
```
wrandall22, line 21:
```suggestion
  .row {
```

wrandall22:
I think this is good for UAT and some git commit cleanup for that last bit of styling. When Mike signs off on it, I'm good with it merging.

branded-checkout.html
wrandall22: Bump

wrandall22:
You can run `yarn lint:styles --fix` to apply the new SCSS linting.

src/assets/scss/_cvv.scss
dr-bizz, line 1: To keep the folder structure consistent, what are your thoughts about moving this file to the below, so there isn't any guessing which component or directive it belongs to?

```
src/assets/scss/directives/creditCardDvv.scss
```

PR #1110, opened 2024-10-29
[no-Jira] Upgrade angular to final version

canac:
The lockfile contained two versions of angular. I'm not sure if version 1.7.9 was actually used, but I manually deleted the `angular@*` section from the lockfile and ran `yarn` to update the lockfile. Now we only have angular version 1.8.3 in the lockfile. Version 1.8.3 is the final version for AngularJS. This resolves two XSS security vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@babel/runtime-corejs2": "^7.25.9",
     "@cruglobal/cru-payments": "^1.2.4",
     "@datadog/browser-rum": "^5.16.0",
-    "angular": "^1.8.2",
+    "angular": "^1.8.3",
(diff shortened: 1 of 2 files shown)


wrandall22 (approved):
Looks like the only difference between 1.8.2 and 1.8.3 is README files in npm (see [Changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md))

canac:
> Looks like the only difference between 1.8.2 and 1.8.3 is README files in npm

@wrandall22 Good to know. Then the only noticeable change will be removing the v1.7.x version from the lockfile.

PR #1111, opened 2024-10-29
[no-Jira] Upgrade all instances of babel-traverse

canac:
Somehow #1108 still ended up with a vulnerable version of `@babel/traverse` in the lockfile. This PR ensures that we're only using one version of babel-traverse and that that version resolves https://github.com/CruGlobal/give-web/security/dependabot/127.


dr-bizz (approved):
Thank you for fixing this

wrandall22 (approved):

PR #1114, opened 2024-10-29
[no-Jira] Improve ARIA labels and tabbable inputs

canac:
This PR implements some low-hanging accessibility improvements for the give site. Most of them are related to improving screen reader labels and making sure that inputs are tabbable.

The biggest change is that inputs are supposed to go inside of the `<label>` so that the label applies to the input. I fixed this for most of the forms and inputs across the application. I had to change some CSS rules after changing the DOM structure.

The remaining changes should be well-described by their commit messages.

### DIFF
--- a/src/app/cart/cart.tpl.html
+++ b/src/app/cart/cart.tpl.html
@@ -9,7 +9,7 @@
             <div class="panel-body">
               <div class="row">
                 <div class="col-sm-6">
-                  <h3 class="panel-name" translate>Your Gift Cart</h3>
+                  <h3 id="cart-header" class="panel-name" translate>Your Gift Cart</h3>
                 </div>
                 <div class="col-sm-6 text-right">
                   <loading inline="true" class="text-right" ng-if="$ctrl.updating">
@@ -28,7 +28,7 @@
                 Your cart is empty
               </p>
               <form ng-if="$ctrl.cartData.items.length">
-                <table class="table giftsum-table">
+                <table class="table giftsum-table" aria-labelledby="cart-header">
                   <thead>
                   <tr>
                     <th class="th-title-gift" translate>Gift</th>
--- a/src/app/cart/cart.tpl.html
+++ b/src/app/cart/cart.tpl.html
@@ -40,7 +40,7 @@
                   <tr class="giftsum-gift-row" ng-repeat="i in $ctrl.cartData.items">
                     <td class="td-gift">
                       <img desig-src="{{i.designationNumber}}" class="giftsum-profile pull-left" width="90" height="51">
-                      <span class="giftsum-person giftsum-title"><a ng-href="/{{i.designationNumber}}">{{i.displayName}}</a></span>
+                      <span id="cart-{{$index}}" class="giftsum-person giftsum-title"><a ng-href="/{{i.designationNumber}}">{{i.displayName}}</a></span>
                       <span class="giftsum-accountnum giftsum-detail">#{{i.designationNumber}}</span>
                     </td>
                     <td class="td-frequency">
@@ -57,9 +57,9 @@
                       <span class="giftsum-detail"><span class=" visible-xs" translate>Gift Amount</span></span>
                     </td>
                     <td class="td-actions giftsum-actions text-sm-right" ng-if="!i.removing">
-                      <a href="" class="btn btn-link btn-giftsum-action" ng-click="$ctrl.editItem(i)" translate>Edit</a>
+                      <a href="" class="btn btn-link btn-giftsum-action" ng-click="$ctrl.editItem(i)" aria-describedby="cart-{{$index}}" translate>Edit</a>
                       <span class="sep"> | </span>
-                      <a href="" class="btn btn-link btn-giftsum-action" ng-click="$ctrl.removeItem(i)" translate>Remove</a>
+                      <a href="" class="btn btn-link btn-giftsum-action" ng-click="$ctrl.removeItem(i)" aria-describedby="cart-{{$index}}" translate>Remove</a>
                       <div class="text-danger" role="alert" ng-if="i.removingError">
                         <span translate>Error removing</span>
                         <span class="hidden-xs hidden-sm" translate>gift</span>
--- a/src/app/designationEditor/titleModal/titleModal.tpl.html
+++ b/src/app/designationEditor/titleModal/titleModal.tpl.html
@@ -1,15 +1,19 @@
 <div class="modal-body">
   <div class="form-group">
-    <label>Receipt Title</label>
-    <input name="receiptTitle" type="text" class="form-control" ng-model="$ctrl.receiptTitle" disabled>
-    <span class="help-block">
+    <label>
+      Receipt Title
+      <input name="receiptTitle" type="text" class="form-control" aria-describedby="receipt-help" ng-model="$ctrl.receiptTitle" disabled>
+    </label>
+    <span id="receipt-help" class="help-block">
       <strong>Email <a href="mailto:fds@cru.org">fds@cru.org</a> to update/change.</strong>
     </span>
   </div>
 
   <div class="form-group">
-    <label translate>Give.Cru.org Title</label>
-    <input name="giveTitle" type="text" class="form-control" ng-model="$ctrl.giveTitle">
+    <label>
+      <span translate>Give.Cru.org Title</span>
+      <input name="giveTitle" type="text" class="form-control" ng-model="$ctrl.giveTitle">
+    </label>
   </div>
 </div>
 <div class="modal-footer">
--- a/src/app/designationEditor/titleModal/titleModal.tpl.html
+++ b/src/app/designationEditor/titleModal/titleModal.tpl.html
@@ -5,7 +5,7 @@
       <input name="receiptTitle" type="text" class="form-control" aria-describedby="receipt-help" ng-model="$ctrl.receiptTitle" disabled>
     </label>
     <span id="receipt-help" class="help-block">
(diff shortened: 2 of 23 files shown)


dr-bizz (approved):
This looks great! I tested it using VoiceOver and it works well.

src/app/productConfig/productConfig.component.js
```diff
@@ -46,7 +46,6 @@ export default angular
       campaignCode: '@',
       campaignPage: '@',
       productCode: '@',
-      buttonLabel: '@',
```
wrandall22, line 49: This will break the AEM configuration.

src/app/searchResults/searchResults.tpl.html
```diff
@@ -129,14 +129,14 @@ <h4 class="is-results-divider" ng-if="facet === 'featured' && $ctrl.featuredGrou
                     </a>
                     <div class="is-row-meta">
                       <a ng-href="{{$ctrl.buildVanity(r.path)}}">
-                        <span class="is-row-title">{{r.name}}</span>
+                        <span id="result-{{r.designationNumber}}" class="is-row-title">{{r.name}}</span>
```
wrandall22, line 132: There is a chance this will produce duplicates

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -180,26 +180,24 @@ <h4 class="panel-title border-bottom-small mt0 pt-" translate>
...
-                <label translate>{{'DAY'}}</label>
-                <div class="form-group">
-                  <select class="form-control form-control-subtle"
-                          ng-model="$ctrl.itemConfig['RECURRING_DAY_OF_MONTH']"
-                          ng-options="o as (o | ordinal) for o in $ctrl.possibleTransactionDays($ctrl.itemConfig['RECURRING_START_MONTH'], $ctrl.nextDrawDate)"
-                          ng-change="$ctrl.changeStartDay($ctrl.itemConfig['RECURRING_DAY_OF_MONTH'], $ctrl.itemConfig['RECURRING_START_MONTH'])">
-                  </select>
-                </div>
+                <span translate>{{'DAY'}}</span>
+                <select class="form-control form-control-subtle"
```
wrandall22, line 196: I wonder why you chose to wrap the month with a label but remove the label from day.

src/common/components/addressForm/addressForm.tpl.html
```diff
@@ -1,37 +1,39 @@
...
-            <button id="retryButton1" type="button" class="btn btn-default btn-sm" ng-click="$ctrl.loadCountries()" translate>{{'RETRY'}}</button></div>
-        </div>
-        <div role="alert" ng-messages="$ctrl.parentForm.addressCountry.$error" ng-if="($ctrl.parentForm.addressCountry | showErrors)">
-          <div class="help-block" ng-message="required" translate>{{'COUNTRY_SELECT_ERROR'}}</div>
-        </div>
+      </label>
+      <div role="alert" ng-if="$ctrl.loadingCountriesError">
+        <div class="help-block"><translate>{{'COUNTRY_LIST_ERROR'}}</translate>
+          <button id="retryButton1" type="button" class="btn btn-default btn-sm" ng-click="$ctrl.loadCountries()" translate>{{'RETRY'}}</button></div>
+      </div>
```
wrandall22, line 17: I'd prefer a bit nicer structure of this HTML:

```
      <div role="alert" ng-if="$ctrl.loadingCountriesError">
        <div class="help-block">
          <translate>{{'COUNTRY_LIST_ERROR'}}</translate>
          <button id="retryButton1" type="button" class="btn btn-default btn-sm" ng-click="$ctrl.loadCountries()" translate>{{'RETRY'}}</button>
        </div>
      </div>
```
Not sure if moving translate to its own line will break the spacing, if that is the case, the closing div should still move to its own line.

src/app/productConfig/productConfig.component.js
canac: Thanks for noticing that. I removed that commit.

src/app/searchResults/searchResults.tpl.html
canac: Good point, I'll link them by id instead of designation number.

src/app/designationEditor/titleModal/titleModal.tpl.html
```diff
@@ -1,15 +1,19 @@
...
   <div class="form-group">
-    <label>Receipt Title</label>
-    <input name="receiptTitle" type="text" class="form-control" ng-model="$ctrl.receiptTitle" disabled>
-    <span class="help-block">
+    <label>
+      Receipt Title
+      <input name="receiptTitle" type="text" class="form-control" aria-describedby="receipt-help" ng-model="$ctrl.receiptTitle" disabled>
+    </label>
+    <span id="receipt-help" class="help-block">
       <strong>Email <a href="mailto:fds@cru.org">fds@cru.org</a> to update/change.</strong>
```
wrandall22, line 8: While we're in here, we might want to update this to `dev.services@cru.org`

src/app/designationEditor/websiteModal/websiteModal.tpl.html
```diff
@@ -1,8 +1,10 @@
 <div class="modal-body">
   <div class="form-group">
-    <label>Website Address</label>
-    <input name="website" type="text" class="form-control" placeholder="http://" ng-model="$ctrl.website">
-    <span class="help-block">Fill in this field to add a website link.</span>
+    <label>
+      Website Address
+      <input name="website" type="text" class="form-control" placeholder="https://" aria-describedby="website-help" ng-model="$ctrl.website">
```
wrandall22, line 5: Unrelated to this PR, but someday we might want to add the protocol to the address they put in here if it is missing, it is a common source of mistakes on these pages.

src/app/searchResults/searchResults.tpl.html
```diff
@@ -129,14 +129,14 @@ <h4 class="is-results-divider" ng-if="facet === 'featured' && $ctrl.featuredGrou
...
+                        <span id="result-{{$index}}" class="is-row-title">{{r.name}}</span>
                         <span class="is-row-num">#{{r.designationNumber}}</span>
                       </a>
                     </div>
                     <div class="is-row-actions hidden-xs">
-                      <product-config product-code="{{r.designationNumber}}" campaign-page="{{r.campaignPage}}" button-label="Give a Gift" button-size="sm"></product-config>
+                      <product-config product-code="{{r.designationNumber}}" campaign-page="{{r.campaignPage}}" described-by="result-{{$index}}" button-label="Give a Gift" button-size="sm"></product-config>
                       &nbsp;
-                      <a ng-href="{{$ctrl.buildVanity(r.path)}}" class="btn btn-sm btn-subtle" ng-click="$ctrl.productViewDetailsAnalyticsEvent(r)" translate>Details</a>
+                      <a ng-href="{{$ctrl.buildVanity(r.path)}}" class="btn btn-sm btn-subtle" ng-click="$ctrl.productViewDetailsAnalyticsEvent(r)" aria-describedby="result-{{r.designationNumber}}" translate>Details</a>
```
wrandall22, line 139: Does this need to be updated to `index` as well?

wrandall22 (commented):
Great work on this so far, thanks for making this PR! I left a few more comments throughout.

Have you been able to test the style differences in mobile/desktop and standard/branded checkouts/self-service?

canac:
@wrandall22 Thanks for the feedback. I fixed your issues and also a styling issue with checkbox labels showing up on a different line from the checkbox. I have tried to test everything at least in desktop mode and the main site/branded checkout to make sure everything looks the same. I'm sure it could use another pass, and I should check mobile also. What's the difference between self-service and standard?

wrandall22:
Standard checkout has forms on the page directly, while self-service sometimes has them in modals, for example. I'm thinking about the payment methods, for example.

wrandall22:
I think this might have caused the [DD test](https://app.datadoghq.com/synthetics/details/v9x-sn6-c2e/result/1710298305061973873) to fail

canac:
> I think this might have caused the [DD test](https://app.datadoghq.com/synthetics/details/v9x-sn6-c2e/result/1710298305061973873) to fail

@wrandall22 Thanks for pointing that out. It must have been a bad merge with staging. I reset staging back several days and merged master and this back in. The DataDog synthetic test passes now.

src/app/cart/cart.tpl.html
```diff
@@ -40,7 +40,7 @@ <h3 class="panel-name" translate>Your Gift Cart</h3>
                   <tr class="giftsum-gift-row" ng-repeat="i in $ctrl.cartData.items">
                     <td class="td-gift">
                       <img desig-src="{{i.designationNumber}}" class="giftsum-profile pull-left" width="90" height="51">
-                      <span class="giftsum-person giftsum-title"><a ng-href="/{{i.designationNumber}}">{{i.displayName}}</a></span>
+                      <span id="cart-{{$index}}" class="giftsum-person giftsum-title"><a ng-href="/{{i.designationNumber}}">{{i.displayName}}</a></span>
```
wrandall22, line 43: If there is a desire to have something meaningful rather than an index, in the context of a cart, the designation number/frequency combination is unique.

wrandall22 (approved):
If all the testing goes well, I think this is good to go. My main thought is making sure the UI styling is unchanged.

canac:
@wrandall22 I have gone through multiple times including just now looking at the name, address, and payment forms in checkout, self-service, and branded checkout and can't find any problems with the forms. Like you mentioned, I was especially looking for styling differences/errors. I feel really good about where it stands now.

PR #1115, opened 2024-10-29
[no-Jira] Dependabot upgrades

canac:
This is the last of the dependabot warnings that look relevant to me. The rest are related to dev dependencies (mostly pulled in by `webpack-dev-server`, which isn't used in production), a couple of ReDOS vulnerabilities in Angular, and possible XSS in Internet Explorer.

The crypto-js warning isn't in a function we're using, but I'm upgrading it anyway. I'm not sure if the lodash vulnerabilities are exploitable in our code, but I'm upgrading it anyway also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "angular-ui-router": "^1.0.30",
     "bootstrap-sass": "^3.4.1",
     "change-case-object": "^2.0.0",
-    "crypto-js": "^3.1.9-1",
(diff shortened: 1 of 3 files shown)


dr-bizz (approved):
This looks good to me. I tested the `sha` function locally and it was working. There wasn't any large changes in `lodash` so I didn't test that, but it looked like it was only bug fixes and security patches.

PR #1116, opened 2024-11-19
EP-2521 Standard Buttons instead of Radio Buttons for suggested amounts

rguinee:
[EP-2521](https://jira.cru.org/secure/RapidBoard.jspa?rapidView=126&projectKey=EP&view=detail&selectedIssue=EP-2521&sprint=1285#)

Use Buttons instead of Radio Buttons for suggested amounts. 

Adjusted the ProductConfigForm file to be buttons instead of inputs. Rearranged the amount and label controllers to suit the Figma designs. Added CSS to style the buttons and add interactivity when tapping/clicking an option.

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
+++ b/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
@@ -66,10 +66,65 @@
             <p translate>{{'GIFT_IN_CART_ERROR'}}</p>
           </div>
 
-          <h4 class="panel-title border-bottom-small" translate>
-            {{'GIFT_AMOUNT'}}
-          </h4>
-            <div ng-if="$ctrl.useSuggestedAmounts">
+        <h4 class="panel-title border-bottom-small" translate>
+          {{'GIFT_AMOUNT'}}
+        </h4>
+        <div class="panel-body pt0">
+          
+          <div ng-if="$ctrl.useSuggestedAmounts">
+            <div ng-if="$ctrl.useV3 === 'true'">
+              <fieldset aria-label="Suggested Gift Amounts. Tab for Custom Amount">
+                <div class="button-group">
+                  <div ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
+                    <input 
+                      type="radio"
+                      name="suggestedAmount"
+                      id="sa-{{$index}}"
+                      ng-click="$ctrl.changeAmount(suggested.amount)"
+                      ng-class=" {'active': !$ctrl.customInputActive && $ctrl.itemConfig.AMOUNT === suggested.amount}"
+                      ng-checked="$first"
+                    />
+                    <label for="sa-{{$index}}">
+                      <span class="amount-box">{{$ctrl.suggestedAmount(suggested.amount)}}</span>
+                      <span class="amount-description">{{suggested.label}}</span>
+                    </label>
+                  </div>
+                  <div class="custom-amount-button" ng-class="{'has-error': ($ctrl.itemConfigForm.amount | showErrors)}">
+                    <div class="input-prepend">
+                      <span class="input-prepend-text">$</span>
+                    </div>
+                    <label for="ca">
+                      <input 
+                        name="suggestedAmount"
+                        type="radio"
+                        ng-checked="$ctrl.customInputActive"
+                      />
+                      <input
+                        name="amount"
+                        type="text"
+                        id="ca"
+                        step="1"
+                        tabindex="-1"
+                        ng-model="$ctrl.customAmount"
+                        ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
+                        ng-focus="$ctrl.customInputActive = true;"
+                        ng-required="$ctrl.customInputActive"
+                        placeholder="{{'OTHER_PLACEHOLDER' | translate}}" />
+                    </label>
+                  </div>
+                  <div 
+                    role="alert" 
+                    ng-messages="$ctrl.itemConfigForm.amount.$error"
+                    ng-if="($ctrl.itemConfigForm.amount | showErrors)">
+                      <div class="help-block" ng-message="pattern" translate>{{'VALID_DOLLAR_AMOUNT_ERROR'}}</div>
+                      <div class="help-block" ng-message="required" translate>{{'AMOUNT_EMPTY_ERROR'}}</div>
+                      <div class="help-block" ng-message="minimum" translate translate-values="{currencyLimit: (1 | currency)}">{{'AMOUNT_MIN_ERROR'}}</div>
+                      <div class="help-block" ng-message="maximum" translate translate-values="{currencyLimit: (10000000 | currency)}">{{'AMOUNT_MAX_ERROR'}}</div>
+                  </div>
+                </div>            
+              </fieldset>
+            </div>
+            <div ng-if="$ctrl.useV3 !== 'true'">
               <div class="radio radio-custom-amount"
                   ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
                 <label>
@@ -108,7 +163,8 @@
                   </div>
                 </div>
               </div>
-            </div>
+            </div> 
+          </div>
 
             <div data-toggle="buttons"
                 ng-class="{'has-error': ($ctrl.itemConfigForm.amount | showErrors)}"
--- a/src/assets/scss/_give.scss
+++ b/src/assets/scss/_give.scss
@@ -498,6 +498,137 @@ hr.horizontal-divider {
   font-weight: 700;
 }
 
+.button-group {
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+
+  .help-block {
+    padding: 10px;
+    color: #715927;
+    background-color: #f4d89d;
+    border: 2px solid #f9b625;
+    border-radius: 2px;
+  }
+
+  input[type="radio"] {
+    position: absolute;
+    opacity: 0;
+    cursor: pointer;
+  }
+
+  label {
+    display: flex;
+    flex: 1;
+    cursor: pointer;
+    margin: 0;
+  
+    .amount-box {
+      margin: 0;
+      min-width: 50px;
+      text-align: center;
+      align-content: center;
+      aspect-ratio: 1 / 1;
+      background-color: #ddd;
+      color: inherit;
+      border-radius: 2px 0 0 2px;
+      transition: all 0.3s ease;
+    }
+
+    .amount-description {
+      width: 100%;
+      margin: 0;
+      padding: 5px 10px;
+      align-content: center;
+      background-color: #f0f0f0;
+      font-weight: normal;
+    }
+  }
+
+  input[type="radio"]:hover + label {
+    .amount-description {
+      background-color: #e4e4e4;
+    }
+  }
+  input[type="radio"]:focus + label {
+    .amount-box {
+      min-width: 55px;
+      color:#ddd;
+      background-color:#3a3a3a;
+    }
+  }
+  input[type="radio"]:checked + label {
+    .amount-box {
+      min-width: 55px;
+      color: #ddd;
+      background-color: #3a3a3a;
+    }
+  }
+}
+
+.custom-amount-button {
+  
+  display: flex;
+  flex-direction: row;
+  height: 50px;
+  border: 1px solid #ddd;
+
+  .input-prepend {
+    margin: 0;
+    min-width: 50px;
+    text-align: center;
+    align-content: center;
+    min-width: 50px;
+    aspect-ratio: 1 / 1;
+    background-color: #ddd;
+    cursor: default;
+  }
+
+  label {
+    all: unset;
+    width: 100%;
+    height: 100%;
+
+    input[type="radio"] {
+      position: absolute;
+      opacity: 0;
+    }
+
+    input[type="text"] {
+      width: 100%;
+      height: 100%;
+      padding: 0 7px;
+      border: none;
+      border-radius: 2px;
+      transition: all 0.3s ease;
+
+      &::placeholder {
+        color: $colorCru-gray;
+      }
+
+      &:focus {
+        border: 1px solid #3eb1c8;
+        outline: none;
+      }
+    }
+  }
+}
+
+/* Check for custom amount input */
+.custom-amount-button:has(input:checked) {
+  border: 1px solid #3a3a3a;
+  font-weight: 700;
+  color: #3a3a3a;
+}
+
+.custom-amount-button:has(input:checked) .input-prepend {
+  min-width: 55px;
+  aspect-ratio: 1 / 1;
+  background-color: #3a3a3a;
+  color: #fff;
+  transition: all 0.3s ease;
+}
+
 .table-payment-history {
   thead {
     th {


src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -68,48 +68,54 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                  class="suggested-button"
+                  ng-click="$ctrl.changeAmount(suggested.amount)"
+                  ng-class=" {'active': !$ctrl.customInputActive && $ctrl.itemConfig.AMOUNT === suggested.amount}"
+                  ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
+                  <div class="amount-box">
+                    {{$ctrl.suggestedAmount(suggested.amount)}}
+                  </div>
+                  <div class="label-text">
+                    {{suggested.label}}
+                  </div>
```
wrandall22, line 85: How well does this work with a screen reader?

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -68,48 +68,54 @@ <h4 class="panel-title border-bottom-small" translate>
         <h4 class="panel-title border-bottom-small" translate>
           {{'GIFT_AMOUNT'}}
         </h4>
-        <div class="panel-body pt0">
+        <div class="panel-body p0">
```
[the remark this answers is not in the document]
rguinee: Screenshot 2024-11-27 at 11 57 38 AM
Each button and button text is in the Accessibility Tree, however I have not tested the experience myself yet.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
rguinee: Screenshot 2024-11-27 at 11 57 38 AM<br>
Each button and button label text is in the Accessibility Tree, but I have not tested it myself using a screen reader extension.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -68,48 +68,54 @@ <h4 class="panel-title border-bottom-small" translate>
...
-                       ng-click="$ctrl.changeAmount(suggested.amount)"
-                       ng-checked="!$ctrl.customInputActive && $ctrl.itemConfig.AMOUNT === suggested.amount" />
-                {{$ctrl.suggestedAmount(suggested.amount)}} {{suggested.label}}
-              </label>
-            </div>
-            <div class="radio radio-custom-amount form-inline" ng-class="{'has-error': ($ctrl.itemConfigForm.amount | showErrors)}">
-              <div class="form-group">
+            <div class="button-group">
+              <div 
+                  role="button"
```
j2trumpet, line 75: Why did you move away from an actual button element?

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
rguinee: We'll need to change the element type to radio input and I can just style it to look as it does now. Also, the custom amount isn't clearly showing up in the Accessibility Tree, so I'll make sure that works correctly.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
j2trumpet: I think a styled radio input would work fine, but I'm not fully read-in on this task, so I've assumed there was some reason to switch from radio inputs to buttons.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -68,48 +68,54 @@ <h4 class="panel-title border-bottom-small" translate>
...
-                         placeholder="{{'OTHER_PLACEHOLDER' | translate}}" />
+                  <input 
+                    name="suggestedAmount"
+                    type="radio"
+                    ng-checked="$ctrl.customInputActive" />
+                  <input
+                    name="amount"
+                    type="text"
+                    step="1"
+                    tabindex="-1"
```
rguinee, line 100: @wrandall22 was there a reason why we wanted to prevent tab navigation specifically for a custom amount input?

wrandall22: It looks like it has been that way from the [beginning](https://github.com/CruGlobal/give-web/commit/53a627310a50165a10dd2c93da648a03e35e8050#diff-8be251bc10d103c8d03b3b4b880cd93b2fa17b86eb51b65106eb7ea0edfe0ddf). I don't see a reason given in that commit. Perhaps it was defaulting to the textbox instead of the first radio button without it?

rguinee: Initially this seems to have been functioning inside a modal which you wouldn't want to have present in the tabindex. So I can see why the attribute default being set to -1 makes sense. I'm wondering if there was or still could be some JS that is handling the focus trap and it's changing that attribute or setting it "".

wrandall22: This is still in a modal on standard checkout, just not branded checkout.

rguinee: Oh I see. As of now, a user can tab into it and navigate with arrows. The tabindex threw me off because I assumed it would prevent the ability to focus on the form with tab.

j2trumpet:
For the suggested amount inputs, the design changes (amount background gets darker and height increases) when the user clicks on it, but these changes don't happen for the custom amount until the user enters a value. Could these changes happen when the user clicks/focuses on the custom input field for the sake of consistent UX?

rguinee:
Definitely!

rguinee:
> For the suggested amount inputs, the design changes (amount background gets darker and height increases) when the user clicks on it, but these changes don't happen for the custom amount until the user enters a value. Could these changes happen when the user clicks/focuses on the custom input field for the sake of consistent UX?

@j2trumpet I've made this change. See [here](https://give-stage2.cru.org/search-results.html?type=featured&modal=give-gift&d=1997172&c=2170).

j2trumpet:
I don't know what the ultimate goal for this PR/ticket is, but hopefully sticking with the radio inputs and just styling them differently works. @wrandall22 ?

From a UX/UI standpoint, I think we're good.

wrandall22:
Here is what I see on my phone.

wrandall22:
The main purpose is to make it easier to select these amounts on mobile devices, which is accomplished with this for sure. However, we still need to determine from stakeholders if this is going to be tied to v3 and if it will be tied specifically to mobile.

rguinee:
> Please change the target branch to `EP-2517-branded-checkout-improvements` since this will be tied to `v3`. Also, we need to add some conditionals to make sure this only happens on `v3`.

Updated target branch. I'll need assistance on those conditionals.

wrandall22:
You can look at https://github.com/CruGlobal/give-web/pull/1120/files for the `ng-if` statements as a starting point.

rguinee:
Got it, thanks!

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -68,7 +67,94 @@ <h4 class="panel-title border-bottom-small" translate>
         <h4 class="panel-title border-bottom-small" translate>
           {{'GIFT_AMOUNT'}}
         </h4>
-        <div class="panel-body pt0">
+        <div class="panel-body p0" ng-if="$ctrl.useV3 == 'true'">
+          <fieldset aria-label="Suggested Gift Amounts. Tab for Custom Amount">
```
wrandall22, line 71: I think this label and fieldset need to be within the conditional (probably just move the `ng-if` up to this element)

rguinee: You're referring to the <h4>?

wrandall22: No, I'm referring to the `fieldset`. It should not be in the DOM if there are no suggested amounts.

wrandall22:
Let's make sure to clean up this Git history so the commits from the base branch don't show up here. Also, the productConfigForm conflicts need to be resolved.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
rguinee: Oh I see. If I'm understanding the use of ng-if, it won't be in the DOM if the ng-if is in the parent `<div>` above it right?

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -65,51 +65,139 @@ <h4 class="panel-title border-bottom-small" translate>
...
-                <label>
-                  <input name="suggestedAmount"
-                        type="radio"
-                        ng-click="$ctrl.changeAmount(suggested.amount)"
-                        ng-checked="!$ctrl.customInputActive && $ctrl.itemConfig.AMOUNT === suggested.amount" />
-                  {{$ctrl.suggestedAmount(suggested.amount)}} {{suggested.label}}
+        <h4 class="panel-title border-bottom-small" translate>
+          {{'GIFT_AMOUNT'}}
+        </h4>
+        <div class="panel-body p0" ng-if="$ctrl.useV3 == 'true'">
```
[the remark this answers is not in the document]
rguinee: Thank you

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -65,51 +65,139 @@ <h4 class="panel-title border-bottom-small" translate>
...
+            </div>
+            <div class="radio radio-custom-amount form-inline" ng-class="{'has-error': ($ctrl.itemConfigForm.amount | showErrors)}">
+              <div class="form-group">
+                <label>
+                  <input 
+                    name="suggestedAmount"
+                    type="radio"
+                    ng-checked="$ctrl.customInputActive"
+                  />
+                  <input
```
wrandall22, line 178: Was it intentional to lose the `form-control` and `form-control-subtle` classes in this existing version of suggested amounts?

rguinee: Good catch, it had some styles that were a part of the design before v3. I added those classes back in.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -49,67 +49,154 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                    name="amount"
+                    class="form-control form-control-subtle"
+                    type="text"
+                    id="ca"
```
wrandall22: This line and line 178 are removed in the next commit. This is where a squash would come into play.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -169,97 +258,98 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
+              <button type="button"
+                      id="sendHandlingButton"
+                      class="btn btn-link"
                       tabindex="-1"
-                      placeholder="{{'MESSAGE_EXAMPLE' | translate}}"
-                      style="display:block;"></textarea>
-          </li>
-        </ul>
-      </div><!-- // panel-body -->
-    </div><!-- // panel -->
```
wrandall22: This change is undone in the next commit, should be squashed.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -169,97 +258,98 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
-              <div class="form-group">
-                <label>
-                  <span translate>{{'MONTH'}}</span>
-                  <select class="form-control form-control-subtle"
-                          ng-model="$ctrl.itemConfig['RECURRING_START_MONTH']"
-                          ng-options="(m | date:'MM') as (m | date:'MMMM, yyyy') for m in $ctrl.possibleTransactionMonths($ctrl.nextDrawDate)"
-                          ng-change="$ctrl.changeStartDay($ctrl.itemConfig['RECURRING_DAY_OF_MONTH'], $ctrl.itemConfig['RECURRING_START_MONTH'])">
-                  </select>
-                </label>
+      </div>
```
wrandall22, line 260: This change (removal of comment on div) is undone in the next commit, should be squashed.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -146,7 +233,9 @@ <h4 class="panel-title border-bottom-small" translate>
                 </div>
               </div>
             </div>
-        </div><!-- // panel -->
```
wrandall22, line 149: This change (removal of comment on div) should be squashed.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -156,200 +156,197 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
-                    type="text"
-                    id="ca"
-                    step="1"
-                    tabindex="-1"
-                    ng-model="$ctrl.customAmount"
-                    ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
-                    ng-focus="$ctrl.customInputActive = true;"
-                    ng-required="$ctrl.customInputActive"
-                    placeholder="{{'OTHER_PLACEHOLDER' | translate}}" />
+                  <input name="suggestedAmount"
```
wrandall22: These changes aren't related to ng-if and should be squashed into the previous commit.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -156,200 +156,197 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
-                </div>
-              </label>
-              <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error"
-                  ng-if="($ctrl.itemConfigForm.amount | showErrors)">
-                <div class="help-block" ng-message="pattern" translate>{{'VALID_DOLLAR_AMOUNT_ERROR'}}</div>
-                <div class="help-block" ng-message="required" translate>{{'AMOUNT_EMPTY_ERROR'}}</div>
-                <div class="help-block" ng-message="minimum" translate="AMOUNT_MIN_ERROR" translate-values="{currencyLimit: (1 | currency)}"></div>
-                <div class="help-block" ng-message="maximum" translate="AMOUNT_MAX_ERROR" translate-values="{currencyLimit: (10000000 | currency)}">
+                  <div class="help-block" ng-message="maximum" translate translate-values="{currencyLimit: (10000000 | currency)}">
+                    {{'AMOUNT_MAX_ERROR'}}
```
wrandall22, line 192: This code style change should be squashed into the previous commit.

src/assets/scss/_give.scss
```diff
@@ -629,6 +629,32 @@ input[type="radio"] {
 }
 /* Product Config Form Amounts End */
 
+.radio-method,
```
wrandall22, line 632: This seems to be the main change in this commit, so the commit message should reflect what is happening here.

src/app/branded/branded-checkout.component.js
```diff
@@ -187,6 +187,6 @@ export default angular
       onOrderFailed: '&',
       language: '@',
       showCoverFees: '@',
-      useV3: '@'
+      useV3: '@',
```
wrandall22: This commit should just go away since the only change is removed in the next commit.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -195,109 +195,109 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
+                    <span ng-switch-when="NA" translate>{{'SINGLE_GIFT'}}</span>
+                    <span ng-switch-default>{{f.display}}</span>
+                  </span>
+            </label>
+            <span ng-repeat-end></span>
           </div>
+          <loading inline="true" ng-if="$ctrl.changingFrequency" class="mt-- text-center">
+            <translate>{{'CHANGING_FREQUENCY'}}</translate>
+          </loading>
         </div><!-- // panel-body -->
```
wrandall22, line 171: This looks like it should be squashed into the first or second commit.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -195,109 +195,109 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
                 </div>
               </div>
             </div>
-          </div>
-        </div><!-- // panel-body -->
-      </div><!-- // panel -->
+          </div><!-- // panel-body -->
+        </div><!-- // panel -->
+      </div>
     </div>
```
wrandall22, line 302: This looks like it should be squashed into the first or second commit.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -254,52 +257,50 @@ <h4 class="panel-title border-bottom-small mt" translate>
...
               </div>
             </div>
-          </div><!-- // panel-body -->
-        </div><!-- // panel -->
+          </div>
+        </div><!-- // panel-body -->
       </div>
     </div>
-    <div class="panel panel-default give-modal-panel">
+    <div ng-if="$ctrl.useV3 !== 'true'" class="panel panel-default give-modal-panel">
```
wrandall22, line 303: This is unrelated to your PR, so this change should be squashed with the first commit so it will just disappear.

wrandall22 (changes requested):
It looks like there is still quite a bit of cleanup to do in the git history. What I would recommend at this point, since basically all the work is done in the first commit, is to squash everything into one commit, then do a `git reset head^` to uncommit everything, then make individual commits bringing in specific changes that are logical. For this particular PR, probably 2 or 3 commits would be sufficient, something like:

1) Added margin to Gift Amount header (and why)
2) Build out custom amount buttons for branded checkout version 3
3) Style the new layout

wrandall22 (approved):
Great job!

PR #1117, opened 2024-11-20
EP-2517 Branded Checkout Improvements

caleballdrin:
Base branch with use-v3 config option.


branded-checkout.html
```diff
@@ -11,8 +11,8 @@
 </head>
 <body>
 
-<branded-checkout designation-number="2294554" show-cover-fees="true" ng-cloak></branded-checkout>
-<script src="branded-checkout.v2.js"></script>
+<branded-checkout designation-number="2294554" show-cover-fees="true" use-v3="true" ng-cloak></branded-checkout>
+<script src="dev.v2.js"></script>
```
wrandall22, line 15: We'll want to change this back before we merge.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -188,6 +188,7 @@ export default angular
       next: '&',
       onPaymentFailed: '&',
       radioStationApiUrl: '<',
-      radioStationRadius: '<'
+      radioStationRadius: '<',
+      useV3: '<',
```
wrandall22, line 192: This is also needed in `branded-checkout.component.js`

caleballdrin:
@wjames111 You can make your feature branches off of this one.

branded-checkout.html
wrandall22: There is no config file, we use this flag for everything.

wrandall22:
Reminder to fix lint issues

wrandall22:
@wjames111 looks like the order switch of frequency/amount didn't take into account the new fields that show up:

Screenshot 2024-12-13 at 5 20 06 PM

wrandall22:
So, the current HTML order is Amount, Frequency, Dates; the desired order is Frequency, Dates, Amount (I think, though we should confirm this, because FL donate site has it as Frequency, Amount, Dates). So, if you achieve that with CSS or HTML restructure; either is okay, but CSS will be less duplicate code and less messy I think if you are able to figure out how to do it.

src/common/components/contactInfo/contactInfo.component.js
wrandall22, line 1: We need to account for no phone number on the form when adding custom validators on v3.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -65,7 +65,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
             </div>
           </div>
         </div>
-        <div class="col-sm-2">
+        <div ng-if="$ctrl.useV3 !== 'true'" class="col-sm-2">
```
wrandall22, line 68: The CSS should probably be updated to make first/last name look nicer in this case.

Screenshot 2024-12-18 at 2 04 21 PM

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -121,7 +121,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
 
       </div>
 
-      <div class="row" ng-if="$ctrl.donorDetails['donor-type'] === 'Household'">
+      <div class="row" ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3 !== 'true'">
```
wrandall22, line 124: The merge with staging and all the `hideSpouseDetails` config didn't quite go right.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -239,7 +238,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'CONTACT_INFO'
             </div>
           </div>
         </div>
-        <div class="col-sm-6">
+        <div ng-if="$ctrl.useV3 !== 'true'" class="col-sm-6">
```
wrandall22, line 241: The styling should be updated for v3 to look nicer. I think we should do something about the h4 above in v3 as well. Email doesn't really belong with mailing address, but it feels weird all by itself as well under contact info.

Screenshot 2024-12-18 at 1 58 31 PM

wrandall22 (changes requested):

wrandall22:
Merges of BCO changes to staging should only come from this branch, not the various feature branches pointing to this one. However, not all of those branches are ready to migrate into this one yet.

src/common/components/contactInfo/contactInfo.tpl.html
wrandall22: Specifically, they should probably take half the space as other lines do when having 2 fields.

src/common/components/contactInfo/contactInfo.tpl.html
wrandall22: Email address should take the full width like other fields that are by themselves on a line.

src/common/components/contactInfo/contactInfo.tpl.html
wrandall22: I tracked down the issue to something with the Okta work, so will resolve the comment here.

src/common/components/contactInfo/emailField/emailField.tpl.html
wrandall22, line 1: I didn't catch this in the initial review, but we should change this file's indentation to 2 spaces.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -214,7 +269,7 @@ <h4 class="panel-title border-bottom-small mt0 pt-" translate>
         </div><!-- // panel-body -->
       </div><!-- // panel -->
     </div>
-    <div class="panel panel-default give-modal-panel">
+    <div ng-if="$ctrl.useV3 !== 'true'" class="panel panel-default give-modal-panel">
```
wrandall22, line 272: The latest [EP-2525](https://jira.cru.org/browse/EP-2525) requirements involve this coming back (always present, optional).

dr-bizz:
I have added styleLint to the master branch. I will fix your branch after lunch. This will make it easier to merge into staging.

src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
```diff
@@ -39,7 +39,7 @@ describe('product config form component', function () {
       defaultFrequency: 'MON',
       updateQueryParam: jest.fn(),
       onStateChange: jest.fn(),
-      $window: { location: '' }
+      $window: { location: '' },
```
wrandall22, line 42: Let's revert this change.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -118,11 +120,23 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
           </div>
         </div>
-
+        <div ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3 === 'true'">
+          <div class="col-sm-12 col-md-2">
+            <div class="form-group">
+              <label class="mb0 hidden-xs hidden-sm">
+                <span class="invisible mb0"></span>
+              </label>
+              <button type="button" class="btn btn-primary btn-block" ng-click="$ctrl.toggleSpouseDetails()" uib-tooltip="{{'By adding your spouse\'s name, we can ensure that both of your contributions are linked together, providing a clearer and more accurate record of your household\'s giving.' | translate}}">
```
wrandall22, line 129: It doesn't look like these changes have made it to staging yet.

caleballdrin: I can jump on a call to go through merge conflicts if they are related to single-step changes.

wrandall22:
Looks like there is a slight CSS issue on the `Remove Spouse` button on [this test page](https://give-dev-cloud.cru.org/testing-grounds/test-v3.html):

Screenshot 2025-03-19 at 2 51 12 PM

rguinee:
@wjames111 @wrandall22 Do you have info on breakpoints where this is happening?

src/app/branded/branded-checkout.tpl.html
```diff
@@ -16,7 +16,10 @@
           show-cover-fees="$ctrl.showCoverFees"
           next="$ctrl.next()"
           on-payment-failed="$ctrl.onPaymentFailed($event.donorDetails)"
-          radio-station-api-url="$ctrl.radioStationApiUrl">
+          radio-station-api-url="$ctrl.radioStationApiUrl"
+          use-v3="$ctrl.useV3"
```
canac, line 20: I just finished adding some comments about making sure to say `$ctrl.useV3 === 'true'` instead of just checking `$ctrl.useV3`. However, if we make this change here, then we know that `useV3` is a bool instead of a string in `branded-checkout-step-1`, which will simplify all of those checks. Then it will be safe to just check `$ctrl.useV3` or `!$ctrl.useV3`.

```suggestion
          use-v3="$ctrl.useV3 === 'true'"
```

src/common/components/Recaptcha/RecaptchaWrapper.tsx
```diff
@@ -36,14 +37,21 @@ export const RecaptchaWrapper = ({
   buttonLabel,
   envService,
   $translate,
-  $log
+  $log,
+  $rootScope
 }: RecaptchaWrapperProps): JSX.Element => {
   const recaptchaKey = envService.read('recaptchaKey')
 
+  // Because the onSuccess callback is called by a React component, AngularJS doesn't know that an event happened and doesn't know it needs to rerender. We have to use $apply to ensure that AngularJS rerenders after the event handlers return.
```
canac, line 45: Can we break this comment into multiple lines?

```suggestion
  // Because the onSuccess callback is called by a React component, AngularJS doesn't know that an
  // event happened and doesn't know it needs to rerender. We have to use $apply to ensure that
  // AngularJS rerenders after the event handlers return.
```

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,137 @@ hr.horizontal-divider {
...
+  input[type="radio"]:hover + label {
+    .amount-description {
+      background-color: $colorGrayscale-pebble;
+    }
+  }
+
+  input[type="radio"]:focus + label {
+    .amount-box {
+      min-width: 55px;
+      background-color:$colorGrayscale-charcoal;
```
canac, line 582:
```suggestion
      background-color: $colorGrayscale-charcoal;
```

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -1,4 +1,11 @@
 <form novalidate name="$ctrl.itemConfigForm" ng-submit="$ctrl.saveGiftToCart()">
+  <div class="row">
+    <span class="u-floatRight" style="margin-top: -16px">
+      Secure
```
canac, line 4: Can we add this label to `app.config.js` and translate it like our other labels?

src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
```diff
@@ -371,8 +371,14 @@ describe('product config form component', function () {
     it('orders frequency by name', () => {
       expect($ctrl.frequencyOrder({ name: 'NA' })).toEqual(0)
       expect($ctrl.frequencyOrder({ name: 'MON' })).toEqual(1)
-      expect($ctrl.frequencyOrder({ name: 'QUARTERLY' })).toEqual(2)
-      expect($ctrl.frequencyOrder({ name: 'ANNUAL' })).toEqual(3)
+      if (!$ctrl.hideQuarterly) {
```
canac, line 374: We shouldn't have conditionals in the test. We should have additional test cases that cover each of those possible combinations of `hideAnnual` and `hideQuarterly`.

src/app/checkout/cart-summary/cart-summary.spec.js
```diff
@@ -31,9 +31,9 @@ describe('checkout', function () {
 
     describe('onSubmit', () => {
       it('should emit an event', () => {
-        jest.spyOn(componentInstance.$rootScope, '$emit').mockImplementation(() => {})
-        self.controller.onSubmit(componentInstance)
-        expect(componentInstance.$rootScope.$emit).toHaveBeenCalledWith(submitOrderEvent)
+        jest.spyOn(self.controller.$rootScope, '$emit').mockImplementation(() => {})
```
canac, line 34: The more correct way to do this is to set up a `$rootScope.$on` listener for the event and make sure that the listener is called. That better captures the higher-level intent of the test case.

canac (commented):
This looks fairly solid! It's a large PR, so I'll probably test it out and go through the code again. I could also look more carefully at the tests that changed. But here's a few comments to work through for now.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -118,11 +120,23 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
+              </button>
+            </div>
+          </div>
+        </div>
       </div>
 
-      <div class="row" ng-if="$ctrl.donorDetails['donor-type'] === 'Household'">
-        <div class="col-sm-4">
+      <div class="row" ng-if="($ctrl.donorDetails['donor-type'] === 'Household' && !$ctrl.useV3) || ($ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3 && $ctrl.showSpouseDetails)" >
+        <div ng-class="$ctrl.useV3 ? 'col-sm-6' : 'col-sm-4'">
```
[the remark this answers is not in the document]
canac: Definitely! That will simplify the logic a lot.

src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
```diff
@@ -371,14 +371,22 @@ describe('product config form component', function () {
...
-        expect($ctrl.frequencyOrder({ name: 'ANNUAL' })).toEqual(2)
-      }
+    })
+
+    it('changes frequency order when quarterly is shown', () => {
+      $ctrl.hideQuarterly = false
+      expect($ctrl.frequencyOrder({ name: 'QUARTERLY' })).toEqual(2) 
+    })
+
+    it('does not have frequency order when quarterly is hidden', () => {
```
canac, line 381: Did you mean this?

```suggestion
    it('changes frequency order when quarterly is hidden', () => {
```

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -118,11 +119,23 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
+        <div ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3">
+          <div class="col-sm-12 col-md-2">
+            <div class="form-group">
+              <label class="mb0 hidden-xs hidden-sm">
+                <span class="invisible mb0"></span>
+              </label>
+              <button type="button" class="btn btn-secondary btn-block" ng-click="$ctrl.toggleSpouseDetails()" uib-tooltip="{{'SPOUSE_DETAILS_TOOLTIP' | translate}}">
+                  <span ng-if="!$ctrl.showSpouseDetails">Add Spouse</span>
+                  <span ng-if="$ctrl.showSpouseDetails">Remove</span>
+              </button>
```
canac, line 131: Should the remove button go next to the spouse's name instead of the primary user?

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -118,11 +119,23 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
           </div>
         </div>
-
+        <div ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3">
+          <div class="col-sm-12 col-md-2">
+            <div class="form-group">
+              <label class="mb0 hidden-xs hidden-sm">
+                <span class="invisible mb0"></span>
+              </label>
+              <button type="button" class="btn btn-secondary btn-block" ng-click="$ctrl.toggleSpouseDetails()" uib-tooltip="{{'SPOUSE_DETAILS_TOOLTIP' | translate}}">
```
canac, line 128: If you add a spouse, type their name, and then remove it, the form still submits the spouse's name. I think we should fix that.

README.md
```diff
@@ -58,7 +58,10 @@ Add the following code to your page where appropriate. See the [Branded checkout
     show-cover-fees="true"
     on-order-completed="$event.$window.onOrderCompleted($event.purchase)"
     on-order-failed="$event.$window.onOrderFailed($event.donorDetails)"
-    radio-station-api-url="https://api.domain.com/getStations">
+    radio-station-api-url="https://api.domain.com/getStations"
+    hide-annual="true"
+    hide-quarterly="true">
```
canac, line 63:
```suggestion
    hide-quarterly="true"
```

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -118,11 +120,23 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
...
-
+        <div ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3 === 'true'">
+          <div class="col-sm-12 col-md-2">
+            <div class="form-group">
+              <label class="mb0 hidden-xs hidden-sm">
+                <span class="invisible mb0"></span>
+              </label>
+              <button type="button" class="btn btn-secondary btn-block" ng-click="$ctrl.toggleSpouseDetails()" uib-tooltip="{{'By adding your spouse\'s name, we can ensure that both of your contributions are linked together, providing a clearer and more accurate record of your household\'s giving.' | translate}}">
+                  <span ng-if="!$ctrl.showSpouseDetails">Add Spouse</span>
+                  <span ng-if="$ctrl.showSpouseDetails">Remove</span>
```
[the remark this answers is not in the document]
canac: I don't think this was done

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -1,4 +1,11 @@
 <form novalidate name="$ctrl.itemConfigForm" ng-submit="$ctrl.saveGiftToCart()">
+  <div class="row">
+    <span class="u-floatRight" style="margin-top: -16px">
+      {{'SECURE'}}
```
canac, line 4: I think you need the `translate` filter.

```suggestion
      {{'SECURE' | translate}}
```

canac (commented):
In basic testing, this works pretty well!

Have stakeholders hammered on this as well during UAT?

src/common/components/addressForm/addressForm.tpl.html
```diff
@@ -134,10 +135,51 @@
           <div role="alert" ng-messages="$ctrl.parentForm.addressRegion.$error" ng-if="($ctrl.parentForm.addressRegion | showErrors)">
             <div class="help-block" ng-message="required" translate>{{'SELECT_STATE_ERROR'}}</div>
           </div>
+        </div>
+      </div>
+      <div class="col-sm-6 col-md-3">
```
canac, line 140: It would be really nice to implement this without duplicating the state and ZIP code fields. Maybe I'm missing something, but it seems that just changing the `col-xx-n` class would be enough without the `ng-if`s.

src/common/components/contactInfo/contactInfo.tpl.html
rguinee: I wondered if it would do that as I was working on this. I guess the remove spouse functionality should clear those fields. Would that be handled in the controller?

rguinee:
> In basic testing, this works pretty well!
> 
> Have stakeholders hammered on this as well during UAT?

We haven't had any thorough UAT done with Mike's team yet.

src/common/components/contactInfo/contactInfo.tpl.html
canac: Yes, we'll need to clear those fields in the controller.

src/common/components/contactInfo/contactInfo.tpl.html
canac: Semi-related, could we add an aria-label of "Remove Spouse" to this button so that it's clearer to users who use accessibility software what exactly is being removed? We should make sure that that label is translated as well.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,187 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                      <div class="help-block" ng-message="required" translate>{{'AMOUNT_EMPTY_ERROR'}}</div>
+                      <div class="help-block" ng-message="minimum" translate translate-values="{currencyLimit: (1 | currency)}">{{'AMOUNT_MIN_ERROR'}}</div>
+                      <div class="help-block" ng-message="maximum" translate translate-values="{currencyLimit: (10000000 | currency)}">{{'AMOUNT_MAX_ERROR'}}</div>
+                  </div>
+                </div>
+              </fieldset>
             </div>
-            <div class="radio radio-custom-amount form-inline" ng-class="{'has-error': ($ctrl.itemConfigForm.amount | showErrors)}">
-              <div class="form-group">
+            <div ng-if="$ctrl.useV3">
```
canac, line 134: This has the same condition as the block above, so they are both showing.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,187 @@ <h4 class="panel-title border-bottom-small" translate>
...
-                 ng-repeat="suggested in $ctrl.suggestedAmounts | orderBy:'order'">
-              <label>
-                <input name="suggestedAmount"
-                       type="radio"
-                       ng-click="$ctrl.changeAmount(suggested.amount)"
-                       ng-checked="!$ctrl.customInputActive && $ctrl.itemConfig.AMOUNT === suggested.amount" />
-                {{$ctrl.suggestedAmount(suggested.amount)}} {{suggested.label}}
-              </label>
+            <div ng-if="$ctrl.useV3">
+              <fieldset aria-label="Suggested Gift Amounts. Tab for Custom Amount">
```
canac, line 83: @rguinee Thanks for the improvements here! I think I understand what you were trying to do here, but it isn't quite working for me. Maybe I'm doing something wrong. For one thing, I haven't found out how to get the label for this `fieldset` announced. For another, since the custom amount input has a tabindex of -1, I can't tab to it like this label instructs me to.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,187 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                    <input class="form-control form-control-subtle"
+                          name="amount"
+                          type="text"
+                          step="1"
+                          tabindex="-1"
+                          ng-model="$ctrl.customAmount"
+                          ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
+                          ng-focus="$ctrl.customInputActive = true;"
+                          ng-required="$ctrl.customInputActive"
+                          placeholder="{{'OTHER_PLACEHOLDER' | translate}}" />
```
canac, line 160: Can we put the `ng-if` and the `class="button-group"` directly on this fieldset and remove those two `div`s to reduce nesting?

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,187 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                  </label>
+                  <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error"
+                      ng-if="($ctrl.itemConfigForm.amount | showErrors)">
```
canac, line 163:
```suggestion
                      ng-focus="$ctrl.customInputActive = true"
```

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,137 @@ hr.horizontal-divider {
...
+    .amount-description {
+      background-color: $colorGrayscale-pebble;
+    }
+  }
+
+  input[type="radio"]:focus + label {
+    .amount-box {
+      min-width: 55px;
+      background-color: $colorGrayscale-charcoal;
+      color: $colorGrayscale-rock;
```
canac, line 583: @rguinee Would it make sense to make the color white to improve contrast and match other buttons, like the frequency buttons?

canac (commented):
I tested this some more and had some challenges with the new gift amount selectors.

canac:
@wjames111 Do you mind waiting to rebase and force-push until after we're done with code review so that it's easier to keep track of what changed? Feel free to make and push `fixup!` commits, and then you can do an interactive rebase to clean them all up right before merging. Let me know if you need any help with making fixup commits in git.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
rguinee: Good catch, how do you want to approach these issues together?

src/assets/scss/_give.scss
rguinee: Yes

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
canac: Not sure how this happened, but the suggestion I meant to make was removing the semicolon in `ng-focus`: `ng-focus="$ctrl.customInputActive = true;"` --> `ng-focus="$ctrl.customInputActive = true"`. I think GitHub got confused about which line I was commenting on. Sorry about that! You can put this back to how it was.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
canac: Here's what I'm thinking we'll need to do:

I'm not sure how to best handle the `fieldset` label. We may want to research whether putting `aria-label` on `fieldset`s is ideal. Or this might be fine as-is. I'm not familiar enough to know, I just noticed that I wasn't able to get that label to announce.

It would be great to have some kind of focus ring for the `:focus` state so that the user knows where their focus is as they tab through the form. It through me off to not have one when I was going through it. That would be a SCSS change.

I think we just need to remove `tabindex="-1"` from the custom amount input so that it's tabbable.

@rguinee Do you agree with those changes or have any other thoughts?

@wjames111 Do you feel equipped to make those changes? Or do you need any more clarity?

rguinee: That sounds good @canac! I think the `<fieldset>` issue could be that there's no `aria-labelledby` and the Gift Amount header that would have the id for `aria-labelledby` isn't part of the group. Not sure if we want to bring that `<h4>` inside the `<fieldset>` or not.

Agreed regarding `:focus`.

Agreed regarding removing `tabindex="-1"`. It was there previously and I didn't remove it.

canac: I think what I thought was a label issue is actually that I'm having trouble tabbing from the frequency dropdowns to the amount radio buttons. The label looks correct though.

Also, I think that the custom amount radio button should also have an `aria-label`.

@wjames111

canac: Ideally the old version of the amount selector would be accessible and tabbable, but I don't want to introduce scope creep. If the new v3 amount selector is accessible and tabbable, I'll be happy!

I only see two versions though: `ng-if="$ctrl.useSuggestedAmounts"` and `ng-if="!$ctrl.useSuggestedAmounts"`. Am I overlooking one?

src/common/app.config.js
```diff
@@ -356,7 +356,13 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
...
     RADIO_STATION_LIST_ERROR: 'There was an error loading radio stations in your area.',
     RADIO_STATION_SELECT_ERROR: 'There was an error selecting a radio station.',
-    PREFERRED_RADIO_STATION: 'Preferred Radio Station: '
+    PREFERRED_RADIO_STATION: 'Preferred Radio Station: ',
+    SPOUSE_DETAILS_TOOLTIP: 'By adding your spouse\'s name, we can ensure that both of your contributions are linked together, providing a clearer and more accurate record of your household\'s giving.',
+    SECURE: 'Secure',
+    ADD_SPOUSE: 'Add Spouse',
+    REMOVE_SPOUSE: 'Remove',
+    SUGGESTED_AMOUNT_HELP: 'Suggested Gift Amounts. Tab for Custom Amount',
+    SUGGESTED_AMOUNT: 'Suggested Amount'
```
canac, line 365: Did you mean this to match the label in the template?

```suggestion
    CUSTOM_AMOUNT: 'Custom Amount'
```

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
canac: This is still showing at the same time as the block above.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,183 @@ <h4 class="panel-title border-bottom-small" translate>
...
+                          step="1"
+                          aria-label="{{'CUSTOM_AMOUNT' | translate}}"
+                          ng-model="$ctrl.customAmount"
+                          ng-change="$ctrl.changeCustomAmount($ctrl.customAmount)"
+                          ng-focus="$ctrl.customInputActive = true;"
+                          ng-required="$ctrl.customInputActive"
+                          placeholder="{{'OTHER_PLACEHOLDER' | translate}}" />
+                  </label>
+                  <div role="alert" ng-messages="$ctrl.itemConfigForm.amount.$error"
+                      ng-if="($ctrl.itemConfigForm.amount | showErrors)"
```
canac, line 159:
```suggestion
                      ng-if="($ctrl.itemConfigForm.amount | showErrors)>"
```

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,137 @@ hr.horizontal-divider {
...
+  input[type="radio"]:checked + label {
+    .amount-box {
+      min-width: 55px;
+      background-color: $colorGrayscale-charcoal;
+      color: $colorCru-white;
+    }
+  }
+}
+
+.custom-amount-button {
```
canac, line 596: Did we lose the styles that made the custom amount radio button's selected state look the same as the selected state of the other radio buttons?

Screenshot 2025-05-07 at 2 37 19 PM

vs

Screenshot 2025-05-07 at 2 38 28 PM

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,137 @@ hr.horizontal-divider {
...
+      position: absolute;
+      opacity: 0;
+    }
+
+    input[type="text"] {
+      width: 100%;
+      height: 100%;
+      padding: 0 7px;
+      border: none;
+      border-radius: 2px;
```
canac, line 627: I think this would look better without a border radius because the gray rectangle the input is inside doesn't have rounded borders.

```suggestion
```

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,137 @@ hr.horizontal-divider {
...
+    }
+  }
+
+  input[type="radio"]:hover + label {
+    .amount-description {
+      background-color: $colorGrayscale-pebble;
+    }
+  }
+
+  input[type="radio"]:focus + label {
```
canac, line 579: We had also talked about adding some kind of focus ring to the active radio button. Maybe @rguinee has thoughts on the look and feel, but I'm thinking something like this (just with rounded borders on the left side).

```scss
input[type="radio"]:focus.active + label {
  border: 1px solid #3eb1c8;
}
```

canac (commented):
I think we just have a few more tweaks left!

src/common/app.config.js
wrandall22: Note, suggested amounts and custom amounts are not the same.

canac: I was thinking this would be for a custom suggested amount. Is that correct?

dr-bizz:
Can you merge master into this branch so the CI tests can run?

src/common/app.config.js
```diff
@@ -362,7 +362,7 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     ADD_SPOUSE: 'Add Spouse',
     REMOVE_SPOUSE: 'Remove',
     SUGGESTED_AMOUNT_HELP: 'Suggested Gift Amounts. Tab for Custom Amount',
-    SUGGESTED_AMOUNT: 'Suggested Amount'
+    CUSTOM_AMOUNT: 'Custom Amount'
```
canac, line 367: Make sure to do this for the Spanish version of this label too

canac (changes requested):
Great job with the discipline of creating `fixup!` commits so that it's easier to interactively rebase later!

Thanks for adding the focus styles! That's a major improvement. Can you make them rounded on just the left side to match the rounded amount?

Screenshot 2025-05-19 at 9 38 39 AM

Also, I'm still seeing two sets of radio buttons still.

Screenshot 2025-05-19 at 9 40 29 AM

canac:
@wjames111 I added this to get the suggested amounts to show up. Is there a different way to show the suggested amounts?

```patch
diff --git a/src/app/productConfig/productConfigForm/productConfigForm.component.js b/src/app/productConfig/productConfigForm/productConfigForm.component.js
index 83267cc97..9a385d2f1 100644
--- a/src/app/productConfig/productConfigForm/productConfigForm.component.js
+++ b/src/app/productConfig/productConfigForm/productConfigForm.component.js
@@ -144,7 +144,7 @@ class ProductConfigFormController {
 
     const suggestedAmountsObservable = this.designationsService.suggestedAmounts(this.code, this.itemConfig)
       .do(suggestedAmounts => {
-        this.suggestedAmounts = suggestedAmounts.filter((amount) => amount?.amount)
+        this.suggestedAmounts = [{amount: 10, order: 1} , {amount: 50, order: 2}, {amount: 100, order: 3}]
         this.useSuggestedAmounts = !isEmpty(this.suggestedAmounts)
       })
```

src/assets/scss/_give.scss
```diff
@@ -523,6 +523,142 @@ hr.horizontal-divider {
...
+      }
+    }
+  }
+}
+
+/* Check for custom amount input */
+.custom-amount-button:has(input:checked), .custom-amount-button:has(input:focus) {
+  border: 1px solid $colorGrayscale-charcoal;
+  color: $colorGrayscale-charcoal;
+  font-weight: 700;
```
canac, line 651: What do you think about changing this `font-weight` most of the other inputs?

```suggestion
  font-weight: 500;
```

canac (approved):
This looks great! Thanks all for the hard work!

dr-bizz:
Let's clean up these commits by rebasing before merging.

Also, it looks like you need to merge `master` into this branch.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -47,132 +54,155 @@ <h4 class="panel-title border-bottom-small" translate>
...
                   <div class="help-block" ng-message="required" translate>{{'AMOUNT_EMPTY_ERROR'}}</div>
                   <div class="help-block" ng-message="minimum" translate translate-values="{currencyLimit: (1 | currency)}">{{'AMOUNT_MIN_ERROR'}}</div>
-                  <div class="help-block" ng-message="maximum" translate translate-values="{currencyLimit: (10000000 | currency)}">
-                    {{'AMOUNT_MAX_ERROR'}}
-                  </div>
-                </div>
+                  <div class="help-block" ng-message="maximum" translate translate-values="{currencyLimit: (10000000 | currency)}">{{'AMOUNT_MAX_ERROR'}}</div>
+              </div>
+            </fieldset>
+            <div ng-if="$ctrl.useV3">
```
canac, line 130:
```suggestion
            <div ng-if="!$ctrl.useV3">
```

rguinee (changes requested):

dr-bizz:
You are good. Awesome work gettibng the commit down

canac:
I'm good with you merging

src/common/services/api/order.service.js
```diff
@@ -437,6 +444,54 @@ class Order {
...
+    }
+    controller.submittingOrder = true
+    controller.onSubmittingOrder({ value: true })
+
+    let submitRequest
+    if (controller.bankAccountPaymentDetails) {
+      submitRequest = this.submit()
+    } else if (controller.creditCardPaymentDetails) {
+      const cvv = this.retrieveCardSecurityCode()
+      submitRequest = this.submit(cvv)
```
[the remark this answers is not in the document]
wrandall22: Card bin is needed for the recaptcha work. It probably got lost on one of the rebases or merges.

PR #1118, opened 2024-11-20
EP-2519 - Add styles for reordering gift selection on useV3.

wjames111:
- move the frequency choice above the amount choice so that the user will pick it first. This should be conditional based on a new configuration option which defaults to the current version.
- Jira [ticket 2517](https://jira.cru.org/browse/EP-2519)

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
+++ b/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
@@ -69,7 +69,6 @@
           <h4 class="panel-title border-bottom-small" translate>
             {{'GIFT_AMOUNT'}}
           </h4>
-          <div class="panel-body pt0">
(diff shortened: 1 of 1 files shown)


dr-bizz (commented):
The code looks great!

As a personal preference, I think we should remove the padding around the amount if it is no longer the first option. - I also don't know why the padding is there in the first place.

Can we also push this to staging so we can test it there?

Screenshot 2024-11-26 at 9 40 57 AM

dr-bizz:
Ah, since this is going into the `EP-2517-branded-checkout-improvements`, you don't need to add it to staging.

wrandall22:
We won't be able to test in stage until the new config is available in WP or AEM.

dr-bizz (approved):
Looks great!

PR #1119, opened 2024-11-20
EP-2518 - Remove ConfirmationStep

caleballdrin:
### Description
There is a desire to remove step 2 of the branded checkout to reduce abandoned carts. This will mean the continue button on the first page should submit the gift and take the user to the thank you page.
https://jira.cru.org/browse/EP-2518

### Changes I Made:
- Move the submit logic to `order.service` and use it both in `branded-checkout-step-1` and `step-3`
- Conditionally show "Continue" or "Submit Your Gift" based on the `useV3` attribute
- Change `next()` page change based on the `useV3` attribute
- Add fullscreen fixed loading message
- Move checkout-error-messages to a new component that is used both in `branded-checkout-step-1` and `step-3`

### DIFF
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -7,19 +7,25 @@ import checkoutStep2 from 'app/checkout/step-2/step-2.component'
 
 import cartService from 'common/services/api/cart.service'
 import orderService from 'common/services/api/order.service'
+import analyticsFactory from '../../analytics/analytics.factory'
 import brandedAnalyticsFactory from '../../branded/analytics/branded-analytics.factory'
-
 import { FEE_DERIVATIVE } from 'common/components/paymentMethods/coverFees/coverFees.component'
 
 import template from './branded-checkout-step-1.tpl.html'
+import 'rxjs/add/operator/catch'
+import 'rxjs/add/operator/do'
+import 'rxjs/add/operator/finally'
 
 const componentName = 'brandedCheckoutStep1'
 
 class BrandedCheckoutStep1Controller {
   /* @ngInject */
-  constructor ($log, $filter, brandedAnalyticsFactory, cartService, orderService) {
+  constructor ($scope, $log, $filter, $window, analyticsFactory, brandedAnalyticsFactory, cartService, orderService) {
+    this.$scope = $scope
     this.$log = $log
     this.$filter = $filter
+    this.$window = $window
+    this.analyticsFactory = analyticsFactory
     this.brandedAnalyticsFactory = brandedAnalyticsFactory
     this.cartService = cartService
     this.orderService = orderService
@@ -154,11 +160,111 @@ class BrandedCheckoutStep1Controller {
   checkSuccessfulSubmission () {
     if (every(this.submission, 'completed')) {
       if (every(this.submission, { error: false })) {
-        this.next()
+        if (this.useV3 === 'true') {
+          this.submitOrderInternal()
+        } else {
+          this.next()
+        }
       } else {
         this.submitted = false
       }
     }
+  }
+
+  loadCart () {
+    this.errorLoadingCart = false
+
+    const cart = this.cartService.get()
+    cart.subscribe(
+      data => {
+        // Setting cart data and analytics
+        this.cartData = data
+        this.brandedAnalyticsFactory.saveCoverFees(this.orderService.retrieveCoverFeeDecision())
+        if (this.cartData && this.cartData.items) {
+          this.brandedAnalyticsFactory.saveItem(this.cartData.items[0])
+        }
+        this.brandedAnalyticsFactory.addPaymentInfo()
+      },
+      error => {
+        // Handle errors by setting flag and logging the error
+        this.errorLoadingCart = true
+        this.$log.error('Error loading cart data for branded checkout (single step)', error)
+      }
+    )
+    return cart
+  }
+
+  loadCurrentPayment () {
+    this.loadingCurrentPayment = true
+
+    const getCurrentPayment = this.orderService.getCurrentPayment()
+    getCurrentPayment.finally(() => {
+      this.loadingCurrentPayment = false
+    }).subscribe(
+      data => {
+        if (!data) {
+          this.$log.error('Error loading current payment info: current payment doesn\'t seem to exist')
+        } else if (data['account-type']) {
+          this.bankAccountPaymentDetails = data
+        } else if (data['card-type']) {
+          this.creditCardPaymentDetails = data
+        } else {
+          this.$log.error('Error loading current payment info: current payment type is unknown')
+        }
+      },
+      error => {
+        this.$log.error('Error loading current payment info', error)
+      }
+    )
+    return getCurrentPayment
+  }
+
+  checkErrors () {
+    // Then check for errors on the API
+    return this.orderService.checkErrors().do(
+      (data) => {
+        this.needinfoErrors = data
+      })
+      .catch(error => {
+        this.$log.error('Error loading checkErrors', error)
+      })
+  }
+
+  submitOrderInternal () {
+    this.loadingAndSubmitting = true
+    this.loadCart()
+      .mergeMap(() => {
+        return this.loadCurrentPayment()
+      })
+      .mergeMap(() => {
+        return this.checkErrors()
+      })
+      .mergeMap(() => {
+        return this.orderService.submitOrder(this)
+      })
+      .finally(() => {
+        this.loadingAndSubmitting = false
+      })
+      .subscribe(() => {
+        this.next()
+      })
+  }
+
+  handleRecaptchaFailure () {
+    this.analyticsFactory.checkoutFieldError('submitOrder', 'failed')
+    this.submittingOrder = false
+    this.loadingAndSubmitting = false
+    this.onSubmittingOrder({ value: false })
+
+    this.loadCart()
+
+    this.onSubmitted()
+    this.submissionError = 'generic error'
+    this.$window.scrollTo(0, 0)
+  }
+
+  canSubmitOrder () {
+    return !this.submittingOrder
   }
 }
 
@@ -169,6 +275,7 @@ export default angular
     checkoutStep2.name,
     cartService.name,
     orderService.name,
+    analyticsFactory.name,
     brandedAnalyticsFactory.name
   ])
   .component(componentName, {
@@ -189,6 +296,9 @@ export default angular
       onPaymentFailed: '&',
       radioStationApiUrl: '<',
       radioStationRadius: '<',
+      onSubmittingOrder: '&',
+      onSubmitted: '&',
       useV3: '<',
+      loadingAndSubmitting: '<'
     }
   })
--- a/src/app/branded/step-1/branded-checkout-step-1.component.js
+++ b/src/app/branded/step-1/branded-checkout-step-1.component.js
@@ -4,6 +4,7 @@ import every from 'lodash/every'
 import productConfigForm from 'app/productConfig/productConfigForm/productConfigForm.component'
 import contactInfo from 'common/components/contactInfo/contactInfo.component'
 import checkoutStep2 from 'app/checkout/step-2/step-2.component'
+import checkoutErrorMessages from 'app/checkout/checkout-error-messages/checkout-error-messages.component'
 
 import cartService from 'common/services/api/cart.service'
 import orderService from 'common/services/api/order.service'
@@ -273,6 +274,7 @@ export default angular
     productConfigForm.name,
     contactInfo.name,
     checkoutStep2.name,
+    checkoutErrorMessages.name,
     cartService.name,
     orderService.name,
     analyticsFactory.name,
--- a/src/app/branded/step-1/branded-checkout-step-1.tpl.html
+++ b/src/app/branded/step-1/branded-checkout-step-1.tpl.html
@@ -46,10 +46,26 @@
   </div>
   <div class="panel">
     <div class="panel-body text-right">
-      <button class="btn btn-primary"
+      <div class="checkout-cta pull-right" ng-if="$ctrl.useV3 === 'true'">
+       <recaptcha-wrapper
+              action="'branded_submit'"
+              on-success="$ctrl.submit"
+              on-failure="$ctrl.handleRecaptchaFailure"
+              component-instance="$ctrl"
+              button-id="'submitOrderButton'"
+              button-type="'submit'"
+              button-classes="'btn btn-primary btn-lg btn-block'"
+              button-disabled="$ctrl.errorLoadingProductConfig || !$ctrl.canSubmitOrder()"
+              button-label="'SUBMIT_GIFT'"></recaptcha-wrapper>
+        </div>
+
+      <button ng-if="$ctrl.useV3 !== 'true'" class="btn btn-primary"
               ng-click="$ctrl.submit()"
               ng-disabled="$ctrl.errorLoadingProductConfig"
               translate>{{'CONTINUE'}}</button>
-    </div>
+  </div>
   </div>
 </div>
+<loading type="fixed" ng-if="$ctrl.loadingAndSubmitting">
+  <translate>{{'SUBMITTING_GIFT'}}</translate>
+</loading>
--- a/src/app/branded/step-1/branded-checkout-step-1.tpl.html
+++ b/src/app/branded/step-1/branded-checkout-step-1.tpl.html
@@ -1,3 +1,8 @@
+<checkout-error-messages
+  needinfo-errors="$ctrl.needinfoErrors"
+  submission-error="$ctrl.submissionError"
+  submission-error-status="$ctrl.submissionErrorStatus">
+</checkout-error-messages>
 <div class="panel">
   <div class="panel-body loading-overlay-parent">
     <loading ng-if="$ctrl.loadingProductConfig">
--- a/src/app/checkout/step-3/step-3.component.spec.js
+++ b/src/app/checkout/step-3/step-3.component.spec.js
@@ -4,7 +4,6 @@ import { Observable } from 'rxjs/Observable'
 import 'rxjs/add/observable/of'
 import 'rxjs/add/observable/throw'
 
-import { cartUpdatedEvent } from 'common/components/nav/navCart/navCart.component'
 import { SignInEvent } from 'common/services/session/session.service'
 
 import module from './step-3.component'
@@ -37,6 +36,7 @@ describe('checkout', () => {
           getCurrentPayment: () => Observable.of(self.loadedPayment),
           checkErrors: () => Observable.of(['email-info']),
           submit: () => Observable.of('called submit'),
+          submitOrder: () => Observable.of('called submitOrder'),
           retrieveCardSecurityCode: () => self.storedCvv,
           retrieveLastPurchaseLink: () => Observable.of('purchaseLink'),
           retrieveCoverFeeDecision: () => self.coverFeeDecision,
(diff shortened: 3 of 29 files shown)


src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +161,176 @@ class BrandedCheckoutStep1Controller {
...
+      catchError(error => {
+        this.$log.error('Error loading checkErrors', error)
+        return Observable.throw(error)
+      }))
+  }
+
+  submitOrderInternal () {
+    this.loadingAndSubmitting = true
+    // Start by loading the cart
+    this.loadCart().pipe(
```
wrandall22, line 238: We should not be dropping the radio station load here. Currently, it is loading in parallel with the cart in `branded-checkout-step-2.component`.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +161,176 @@ class BrandedCheckoutStep1Controller {
...
+        this.needinfoErrors = data
+        console.log('checking errors')
+      }),
+      catchError(error => {
+        this.$log.error('Error loading checkErrors', error)
+        return Observable.throw(error)
+      }))
+  }
+
+  submitOrderInternal () {
```
wrandall22, line 235: I'm going to dig deeper, but I'm hoping I can find a way to not have to recreate this logic here by including step-3 on this step-1 page.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +161,176 @@ class BrandedCheckoutStep1Controller {
...
+        } else {
+          return Observable.throw({ data: 'Current payment type is unknown' })
+        }
+        console.log('submitRequest', submitRequest)
+        // Return the observable from the submit request to continue processing
+        return submitRequest
+      })).subscribe({
+      next: () => {
+        // If order submission was successful, process analytics and proceed to the next steps
+        this.brandedAnalyticsFactory.purchase(this.donorDetails, this.cartData, this.orderService.retrieveCoverFeeDecision())
```
wrandall22, line 275: This is already being called when loading the thank you summary

wrandall22 (changes requested):
I still have a lot of digging to do, but figured I'd give an initial pass. Good job so far!

src/common/components/Recaptcha/Recaptcha.tsx
```diff
@@ -89,29 +89,29 @@ export const Recaptcha = ({
         if (data?.success === true && isValidAction(data?.action)) {
           if (data.score < 0.5) {
             $log.warn(`Captcha score was below the threshold: ${data.score}`)
-            onFailure(componentInstance)
+            onFailure.call(componentInstance,componentInstance)
```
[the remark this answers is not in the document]
canac: The goal was so that controller methods like `submitOrderInternal` can use `this` like normal controller methods. It's a hidden footgun if `this` is bound to `window`. But it will be great if that won't be needed in the future.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -89,6 +93,9 @@ class BrandedCheckoutStep1Controller {
   submit () {
     this.resetSubmission()
     this.submitted = true
+    if (this.useV3 === 'true') {
+      this.$scope.$digest()
+    }
```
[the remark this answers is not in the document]
canac: When you bind events with `ng-*`, Angular knows about them and runs its dirty-checking logic after the event listener completes. Because this method was called in response to a React event that Angular doesn't know about, the change to `submitted` wasn't triggering the `$onChange` handlers. That's why we're manually forcing a recheck. I believe we could also do `$rootScope.$apply(() => { ... })` in `Recaptcha.tsx`, which is probably a better solution.

src/app/branded/step-1/branded-checkout-step-1.component.js
canac, line 238: Instead of using `.pipe()`, does this work if you do `this.loadCart().mergeMap(...).mergeMap(...).mergeMap(...)`?

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +161,176 @@ class BrandedCheckoutStep1Controller {
...
+        this.loadingCurrentPayment = false
+        this.$log.error('Error loading current payment info', error)
+        return Observable.throw(error) // Propagate error
+      })
+    )
+  }
+
+  checkErrors () {
+    // Then check for errors on the API
+    return this.orderService.checkErrors().pipe(
```
canac, line 224: Instead of using `.pipe()` can you do `this.orderService.checkErrors().do(...).catch(...)`? I see those operators being used in other places in the codebase.

src/app/branded/step-1/branded-checkout-step-1.tpl.html
```diff
@@ -46,10 +90,26 @@ <h3 class="panel-name" translate>{{'PAYMENT'}}</h3>
...
   <div class="panel">
     <div class="panel-body text-right">
-      <button class="btn btn-primary"
+      <div class="checkout-cta pull-right" ng-if="$ctrl.useV3 === 'true'">
+       <recaptcha-wrapper
+       
+              action="'branded_submit'"
+              on-success="$ctrl.submit"
+              on-failure="$ctrl.handleRecaptchaFailure"
+              component-instance="$ctrl.selfReference"
```
canac, line 99: You should be able to do this and not need to set `this.selfReference` in the controller.

```suggestion
              component-instance="$ctrl"
```

src/common/components/loading/loading.scss
```diff
@@ -66,6 +66,13 @@ loading{
     z-index: 2;
   }
 
+  .loading-fixed:first-child {
```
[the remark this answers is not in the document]
caleballdrin: The existing options have the loading icon and text centered in the div. Since our form is very long, the user can't see the loading text unless they scroll up. This new style fixes the loading text to the center of the viewport.

src/app/branded/step-1/branded-checkout-step-1.component.js
canac: I'm sure it's possible, but there will be more branching in `step-3`. The tricky part will be changing `step-3` to accept a new prop that lets us save the 3 forms and wait for them to finish saving before doing the submission.

src/app/branded/step-1/branded-checkout-step-1.component.js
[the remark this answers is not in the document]
wrandall22: Yes, using apply is a better way to do this, typically.

src/common/components/Recaptcha/Recaptcha.tsx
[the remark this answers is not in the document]
wrandall22: Yeah, if we're going to make this `.call` we should refactor the instances so we're only passing in `componentInstance` once.

src/app/branded/step-1/branded-checkout-step-1.component.js
wrandall22: Or we could possibly do the handling of the 3 forms in an intermediary, invisible-to-the-user step?

wrandall22: I guess, step-3 itself has review data in the template, so we probably don't want to use it directly; however, it might make sense to refactor the common logic out to a service and use it in both places (step-3 and branded-step-1)

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +160,177 @@ class BrandedCheckoutStep1Controller {
...
+      })
+      .mergeMap(() => {
+      // Clear previous submission errors
+        delete this.submissionError
+        delete this.submissionErrorStatus
+
+        // Prevent multiple submissions
+        if (this.submittingOrder) { return Observable.of(null) }
+
+        this.radioStationName = this.orderService.retrieveRadioStationName()
```
wrandall22, line 256: I think I was wrong about needing this back. As I dig deeper, it looks like this was only used by the review page to display it back to the user, and we have gotten rid of the review step.

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,160 @@ describe('order service', () => {
...
+        mockController.bankAccountPaymentDetails = {}
+        self.orderService.submitOrder(mockController)
+        .subscribe(() => {
+          done()
+        })
+
+        expect(self.orderService.submit).toHaveBeenCalled()
+        expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+
+        expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
```
wrandall22, line 1267: Usually these expectations should be within the subscribe

src/app/checkout/step-3/step-3.component.spec.js
```diff
@@ -323,130 +324,7 @@ describe('checkout', () => {
...
-          expect(self.controller.onSubmittingOrder).toHaveBeenCalledWith({ value: false })
-          expect(self.controller.onSubmitted).toHaveBeenCalled()
-        })
-
-        it('should submit the order normally if paying with a bank account', () => {
-          self.controller.bankAccountPaymentDetails = {}
-          self.controller.submitOrder()
-
-          expect(self.controller.orderService.submit).toHaveBeenCalled()
-          expect(self.controller.analyticsFactory.purchase).toHaveBeenCalledWith(self.controller.donorDetails, self.controller.cartData, self.coverFeeDecision)
```
wrandall22, line 359: I see we're losing the tests for the analytics events.

src/app/checkout/step-3/step-3.component.spec.js
```diff
@@ -323,130 +324,7 @@ describe('checkout', () => {
...
-          expect(self.controller.orderService.clearCardSecurityCodes).toHaveBeenCalled()
-          expect(self.controller.changeStep).toHaveBeenCalledWith({ newStep: 'thankYou' })
```
wrandall22, line 361: We might still want this here

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,160 @@ describe('order service', () => {
...
+      })
+
+      it('should handle an error submitting an order with a bank account', (done) => {
+        self.orderService.submit.mockImplementation(() => Observable.throw({ data: 'error saving bank account' }))
+
+        mockController.bankAccountPaymentDetails = {}
+        self.orderService.submitOrder(mockController).subscribe(()=>{},
+        () => {
```
wrandall22, line 1275:
```suggestion
        self.orderService.submitOrder(mockController).subscribe(
          () => {
            fail()
          },
          () => {
```

wrandall22: This would also mean indenting the function body one more as well.

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,160 @@ describe('order service', () => {
...
+          done();
+        })
+      })
+
+      it('should submit the order with a CVV if paying with a credit card', (done) => {
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue('1234');
+        mockController.creditCardPaymentDetails = {}
+        mockController.coverFeeDecision = true
+        self.orderService.submitOrder(mockController)
+        .subscribe(() => {
```
wrandall22, line 1293:
```suggestion
        self.orderService.submitOrder(mockController).subscribe(() => {
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,160 @@ describe('order service', () => {
...
+          done()
+        })
+
+        expect(self.orderService.submit).toHaveBeenCalledWith('1234')
+        expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+        expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
```
wrandall22, line 1299: Expectations inside of subscribe

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,160 @@ describe('order service', () => {
...
+        mockController.creditCardPaymentDetails = {}
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue(undefined);
+        mockController.coverFeeDecision = true
+        self.orderService.submitOrder(mockController).subscribe(() => {
+          done()
+        })
+
+        expect(self.orderService.submit).toHaveBeenCalledWith(undefined)
+        expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+        expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
```
wrandall22, line 1312: Expectations inside of subscribe

wrandall22 (changes requested):
I got a start on the review, a few minor things to change so far. Will try to get more done tomorrow.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,116 @@ class BrandedCheckoutStep1Controller {
...
+    // Return the observable instead of subscribing here
+    return this.cartService.get().pipe(
+      tap(data => {
+        // Setting cart data and analytics
+        this.cartData = data
+        this.brandedAnalyticsFactory.saveCoverFees(this.orderService.retrieveCoverFeeDecision())
+        this.brandedAnalyticsFactory.saveItem(this.cartData.items[0])
+        this.brandedAnalyticsFactory.addPaymentInfo()
+        this.brandedAnalyticsFactory.reviewOrder()
+        console.log('done loading cart')
```
canac, line 185: The four logs in this file and one in `order.service.js` should be removed before merging.

src/common/services/api/order.service.spec.js
canac: @wrandall22 After the code-freeze ends we should disable the setting that is ignore test files in standardjs so that the computers will do the formatting for us

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,116 @@ class BrandedCheckoutStep1Controller {
...
         this.submitted = false
       }
     }
   }
+
+  loadCart () {
+    this.errorLoadingCart = false
+
+    // Return the observable instead of subscribing here
+    return this.cartService.get().pipe(
```
canac, line 177: I think it would be simpler to do:

```ts
const cart = this.cartService.get();
cart.subscribe(...);
return cart;
```

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,116 @@ class BrandedCheckoutStep1Controller {
...
       } else {
         this.submitted = false
       }
     }
   }
+
+  loadCart () {
+    this.errorLoadingCart = false
+
+    // Return the observable instead of subscribing here
```
canac, line 176: This comment can be removed

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,116 @@ class BrandedCheckoutStep1Controller {
...
+        this.$log.error('Error loading cart data for branded checkout step 2', error)
+        return Observable.throw(error) // Rethrow the error so the observable chain can handle it
+      })
+    )
+  }
+
+  loadCurrentPayment () {
+    this.loadingCurrentPayment = true
+
+    return this.orderService.getCurrentPayment().pipe(
```
canac, line 199: This can use the same "save observable to a variable, subscribe to it, and return it" pattern.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,116 @@ class BrandedCheckoutStep1Controller {
...
+  checkErrors () {
+    // Then check for errors on the API
+    return this.orderService.checkErrors().do(
+      (data) => {
+        this.needinfoErrors = data
+        console.log('checking errors')
+      })
+      .catch(error => {
+        this.$log.error('Error loading checkErrors', error)
+        return Observable.throw(error)
```
canac, line 229: In my tests, not returning anything from `catch` doesn't change the observable, so I think you can remove this line.

src/app/branded/step-1/branded-checkout-step-1.tpl.html
```diff
@@ -1,3 +1,47 @@
+<div class="alert alert-danger" role="alert" ng-if="$ctrl.needinfoErrors || $ctrl.submissionError">
```
canac, line 1: I think we were going to put these error messages in a separate component.

wrandall22 (changes requested):

wrandall22 (changes requested):

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,111 @@ class BrandedCheckoutStep1Controller {
...
+
+    this.loadCart()
+
+    this.onSubmitted()
+    this.submissionError = 'generic error'
+    this.$window.scrollTo(0, 0)
+  }
+
+  canSubmitOrder () {
+    return !this.submittingOrder
```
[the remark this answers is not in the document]
caleballdrin: This logic in step 2 doesn't work for step 1 because the cartData, donorDetails, and other data are only saved once the user hits submit. So we need the submit button to be enabled for them to do that.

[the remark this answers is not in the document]
wrandall22: Ah, I see. So we have `this.itemConfig`, which has data about the cart item and `this.donorDetails`, which has the contact information. We could use these to determine if they have filled out some of the items (e.g. email). Unfortunately, since the payment information is in an isolate scope, we don't have access to it here without using events. Since 2/3 of the `needInfo` errors are payment method data, maybe it is not worth the effort to check for email.

src/common/services/api/order.service.spec.js
```diff
@@ -2,7 +2,7 @@ import angular from 'angular'
 import 'angular-mocks'
 import omit from 'lodash/omit'
 import cloneDeep from 'lodash/cloneDeep'
-import { Observable } from 'rxjs/Observable'
+import { Observable } from 'rxjs'
```
wrandall22, line 5: Let's put this back the way it was.

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,161 @@ describe('order service', () => {
...
+              done()
+            })
+      })
+
+      it('should handle an error submitting an order with a bank account', (done) => {
+        self.orderService.submit.mockImplementation(() => Observable.throw({ data: 'error saving bank account' }))
+
+        mockController.bankAccountPaymentDetails = {}
+        self.orderService.submitOrder(mockController).subscribe(
+          ()=>{
```
wrandall22, line 1274: Whitespace

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1207,161 @@ describe('order service', () => {
...
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue(undefined);
+        mockController.coverFeeDecision = true
+        self.orderService.submitOrder(mockController).subscribe(() => {
+          expect(self.orderService.submit).toHaveBeenCalledWith(undefined)
+          expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+          expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
+          done()
+        })
+
+
```
wrandall22, line 1313: We don't need these empty lines

wrandall22:
I think this branch is good to resolve conflicts and do some squashing. Also, please put together a more robust PR description.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -7,22 +7,27 @@ import checkoutStep2 from 'app/checkout/step-2/step-2.component'
 
 import cartService from 'common/services/api/cart.service'
 import orderService from 'common/services/api/order.service'
+import analyticsFactory from '../../analytics/analytics.factory'
 import brandedAnalyticsFactory from '../../branded/analytics/branded-analytics.factory'
-
 import { FEE_DERIVATIVE } from 'common/components/paymentMethods/coverFees/coverFees.component'
 
 import template from './branded-checkout-step-1.tpl.html'
+import { Observable } from 'rxjs/Observable'
```
canac, line 15: We also need to add imports for each of the observable functions that we use in this component:

```js
import 'rxjs/add/observable/throw'
import 'rxjs/add/operator/catch'
import 'rxjs/add/operator/do'
```

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -7,22 +7,27 @@ import checkoutStep2 from 'app/checkout/step-2/step-2.component'
...
-  constructor ($log, $filter, brandedAnalyticsFactory, cartService, orderService) {
+  constructor ($scope, $log, $filter, $window, analyticsFactory, brandedAnalyticsFactory, cartService, orderService) {
+    this.$scope = $scope
     this.$log = $log
     this.$filter = $filter
+    this.analyticsFactory = analyticsFactory
     this.brandedAnalyticsFactory = brandedAnalyticsFactory
     this.cartService = cartService
     this.orderService = orderService
+    this.$window = $window
```
canac, line 30: Can you move this line up with all the other $ variables so that the order of assignments matches the order of the arguments?

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,113 @@ class BrandedCheckoutStep1Controller {
...
+        this.cartData = data
+        this.brandedAnalyticsFactory.saveCoverFees(this.orderService.retrieveCoverFeeDecision())
+        this.brandedAnalyticsFactory.saveItem(this.cartData.items[0])
+        this.brandedAnalyticsFactory.addPaymentInfo()
+      },
+      error => {
+        // Handle errors by setting flag and logging the error
+        this.errorLoadingCart = true
+        this.$log.error('Error loading cart data for branded checkout (single step)', error)
+        return Observable.throw(error) // Rethrow the error so the observable chain can handle it
```
canac, line 189: In my testing, you don't need to rethrow the observable (unlike promise `.catch` handlers or JS `catch` blocks). Simply don't return anything from the error handler.

```suggestion
```

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,113 @@ class BrandedCheckoutStep1Controller {
...
+          this.creditCardPaymentDetails = data
+        } else {
+          this.$log.error('Error loading current payment info: current payment type is unknown')
+        }
+        this.loadingCurrentPayment = false
+      },
+      error => {
+        this.loadingCurrentPayment = false
+        this.$log.error('Error loading current payment info', error)
+        return Observable.throw(error) // Propagate error
```
canac, line 215: You can remove this line as well.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,113 @@ class BrandedCheckoutStep1Controller {
...
+      }
+    )
+    return cart
+  }
+
+  loadCurrentPayment () {
+    this.loadingCurrentPayment = true
+
+    const getCurrentPayment = this.orderService.getCurrentPayment()
+    getCurrentPayment.subscribe(
```
canac, line 199: You could use `.finally` here:

```js
getCurrentPayment.finally(() => {
  this.loadingCurrentPayment = false
}).subscribe(...)
```

If you do this make sure to add `import rxjs/Observable/operator/finally`.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,113 @@ class BrandedCheckoutStep1Controller {
...
+        return this.loadCurrentPayment()
+      })
+      .mergeMap(() => {
+        return this.checkErrors()
+      })
+      .mergeMap(() => {
+        return this.orderService.submitOrder(this)
+      })
+      .subscribe(() => {
+        this.next()
```
canac, line 245: This could also use `.finally` since we need to do the same cleanup in the success and error handlers.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -7,22 +7,27 @@ import checkoutStep2 from 'app/checkout/step-2/step-2.component'
...
 import cartService from 'common/services/api/cart.service'
 import orderService from 'common/services/api/order.service'
+import analyticsFactory from '../../analytics/analytics.factory'
 import brandedAnalyticsFactory from '../../branded/analytics/branded-analytics.factory'
-
 import { FEE_DERIVATIVE } from 'common/components/paymentMethods/coverFees/coverFees.component'
 
 import template from './branded-checkout-step-1.tpl.html'
+import { Observable } from 'rxjs/Observable'
+import checkoutErrorMessages from 'app/checkout/checkout-error-messages/checkout-error-messages.component'
```
canac, line 16: Can you move this import up with the other component imports?

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -169,7 +275,9 @@ export default angular
     checkoutStep2.name,
     cartService.name,
     orderService.name,
-    brandedAnalyticsFactory.name
+    analyticsFactory.name,
+    brandedAnalyticsFactory.name,
+    checkoutErrorMessages.name
```
canac, line 280: Can you move this line up with the other components?

src/app/checkout/step-3/step-3.component.spec.js
```diff
@@ -474,7 +349,34 @@ describe('checkout', () => {
...
+    })
+
+    describe('submitOrder', () => {
+      beforeEach(() => {
+        jest.spyOn(self.controller.analyticsFactory, 'purchase')
+        self.controller.retrieveCoverFeeDecision = () => true;
+      })
+
+      it('should call analyticsFactory when it is not branded checkout', () => {
+        jest.spyOn(self.controller.orderService, 'retrieveCoverFeeDecision')
```
canac, line 363: You can remove this line and the same line in the next test, and add `jest.spyOn(self.controller.orderService, 'retrieveCoverFeeDecision').mockReturnValue(true)` to the `beforeEach` block.

src/common/components/Recaptcha/RecaptchaWrapper.tsx
```diff
@@ -50,11 +52,23 @@ export const RecaptchaWrapper = ({
...
+    $rootScope.$apply(()=> {
+      onSuccess.call(componentInstance)
+    })
+  })
+
+  const onFailureWrapped = (() => {
+    $rootScope.$apply(()=> {
+      onFailure.call(componentInstance)
+    })
+  })
```
canac, line 66: Looks like `standardjs` doesn't run on TSX files.

```suggestion
  const onSuccessWrapped = (() => {
    $rootScope.$apply(() => {
      onSuccess.call(componentInstance)
    })
  })

  const onFailureWrapped = (() => {
    $rootScope.$apply(() => {
      onFailure.call(componentInstance)
    })
  })
```

src/common/services/api/order.service.js
```diff
@@ -399,6 +402,55 @@ class Order {
...
+    if (controller.bankAccountPaymentDetails) {
+      submitRequest = this.submit()
+    } else if (controller.creditCardPaymentDetails) {
+      const cvv = this.retrieveCardSecurityCode()
+      submitRequest = this.submit(cvv)
+    } else {
+      submitRequest = Observable.throw({ data: 'Current payment type is unknown' })
+    }
+    return submitRequest
+      .do(() => {
```
canac, line 426: You should add imports for the `do` and `catch` operators as well.

src/common/services/api/order.service.js
```diff
@@ -399,6 +402,55 @@ class Order {
...
+        if (error.config && error.config.data && error.config.data['security-code']) {
+          error.config.data['security-code'] = error.config.data['security-code'].replace(/./g, 'X') // Mask security-code
+        }
+        this.$log.error('Error submitting purchase:', error)
+        controller.onSubmitted()
+        controller.submissionErrorStatus = error.status
+        controller.submissionError = isString(error && error.data) ? (error && error.data).replace(/[:].*$/, '') : 'generic error' // Keep prefix before first colon for easier ng-switch matching
+        this.$window.scrollTo(0, 0)
+
+        return Observable.throw(error) // Return the error as an observable
```
canac, line 451: Re-throwing isn't necessary

src/common/services/api/order.service.js
```diff
@@ -399,6 +402,55 @@ class Order {
...
+    } else if (controller.creditCardPaymentDetails) {
+      const cvv = this.retrieveCardSecurityCode()
+      submitRequest = this.submit(cvv)
+    } else {
+      submitRequest = Observable.throw({ data: 'Current payment type is unknown' })
+    }
+    return submitRequest
+      .do(() => {
+        controller.submittingOrder = false
+        controller.onSubmittingOrder({ value: false })
```
canac, line 428: I see these two lines and `controller.onSubmitted` in `.catch` as well. Looks like another good candidate for using `.finally`.

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+          mockController.bankAccountPaymentDetails = {}
+          self.orderService.submitOrder(mockController).subscribe(
+            () => {
+              expect(self.orderService.submit).toHaveBeenCalled()
+              expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+      
+              expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
+              done()
+            })
+      })
```
canac, line 1268: While working on another PR, I discovered that it's best to add `done` as the error handler for `subscribe`. If the observable unexpectedly throws, the test will abort immediately instead of timing out.

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+
+      it('should handle an error submitting an order with a bank account', (done) => {
+        self.orderService.submit.mockImplementation(() => Observable.throw({ data: 'error saving bank account' }))
+
+        mockController.bankAccountPaymentDetails = {}
+        self.orderService.submitOrder(mockController).subscribe(
+          () => {
+            fail()
+          },
```
canac, line 1277: Some existing tests incorrectly use `fail`, but that is a mocha function, not a jest function. You can pass an error message to `done()` to get the test to fail and print the message.

```suggestion
          () => done("Observable unexpectedly succeeded"),
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+          done()
+        })
+      })
+
+      it('should handle an error submitting an order with a credit card', (done) => {
+        self.orderService.submit.mockImplementation(() => Observable.throw({ data: 'CardErrorException: Invalid Card Number: some details' }))
+        mockController.creditCardPaymentDetails = {}
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue('1234');
+        self.orderService.submitOrder(mockController).subscribe(
+          () => {},
```
canac, line 1320:
```suggestion
          () => done("Observable unexpectedly succeeded"),
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+      it('should submit the order without a CVV if paying with an existing credit card or the cvv in session storage is missing', (done) => {
+        mockController.creditCardPaymentDetails = {}
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue(undefined);
+        mockController.coverFeeDecision = true
+        self.orderService.submitOrder(mockController).subscribe(() => {
+          expect(self.orderService.submit).toHaveBeenCalledWith(undefined)
+          expect(self.orderService.clearCardSecurityCodes).toHaveBeenCalled()
+          expect(mockController.$scope.$emit).toHaveBeenCalledWith(cartUpdatedEvent)
+          done()
+        })
```
canac, line 1312:
```suggestion
        }, done)
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+            done();
+          })
+      })
+
+      it('should mask the security code on a credit card error', (done) => {
+        self.orderService.submit.mockReturnValue(Observable.throw({ data: 'some error', config: { data: { 'security-code': '1234' } } }))
+        self.orderService.retrieveCardSecurityCode = jest.fn().mockReturnValue('1234');
+        mockController.creditCardPaymentDetails = {}
+        self.orderService.submitOrder(mockController).subscribe(
+          () => {},
```
canac, line 1336:
```suggestion
          () => done("Observable unexpectedly succeeded"),
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+          () => { // error handler
+            expect(self.orderService.submit).toHaveBeenCalledWith('1234')
+            expect(self.$log.error.logs[0]).toEqual(['Error submitting purchase:', { data: 'some error', config: { data: { 'security-code': 'XXXX' } } }])
+            done();
+          })
+      })
+
+      it('should throw an error if neither bank account or credit card details are loaded', (done) => {
+        self.orderService.submitOrder(mockController).subscribe(
+          () => {},
```
canac, line 1346:
```suggestion
          () => done("Observable unexpectedly succeeded"),
```

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,159 @@ describe('order service', () => {
...
+            expect(self.$window.scrollTo).toHaveBeenCalledWith(0, 0);
+            done();
+          })
+      })
+
+      it('should clear out cover fee data', (done) => {
+        mockController.creditCardPaymentDetails = {}
+        self.orderService.submitOrder(mockController).subscribe(() => {
+          done()
+        })
```
canac, line 1361:
```suggestion
        }, done)
```

src/common/services/api/order.service.js
canac: Actually, `do` accepts a second argument, which is the error handler. I think using it instead of `.catch` would be better since you can add a success and error handler at the same time.

canac (commented):
This is looking really good! I mostly just have style feedback at this point.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,12 +159,113 @@ class BrandedCheckoutStep1Controller {
...
+  loadCart () {
+    this.errorLoadingCart = false
+
+    const cart = this.cartService.get()
+    cart.subscribe(
+      data => {
+        // Setting cart data and analytics
+        this.cartData = data
+        this.brandedAnalyticsFactory.saveCoverFees(this.orderService.retrieveCoverFeeDecision())
+        this.brandedAnalyticsFactory.saveItem(this.cartData.items[0])
```
canac, line 182: This fails when `loadCart` is called from `handleRecaptchaFailure` because `data.items` is empty. It would probably be best to check whether `items` has at least one item before doing this step.

caleballdrin: @wrandall22 is it worth fixing this or are we going to remove `handleRecaptchaFailure()`?

wrandall22: It is probably worth it for the sake of covering an edge case.

src/app/branded/step-1/branded-checkout-step-1.component.js
```diff
@@ -154,21 +161,121 @@ class BrandedCheckoutStep1Controller {
...
+      }
+    )
+    return cart
+  }
+
+  loadCurrentPayment () {
+    this.loadingCurrentPayment = true
+
+    const getCurrentPayment = this.orderService.getCurrentPayment()
+    getCurrentPayment.finally(() => {
```
wrandall22, line 200: `finally` should come after `subscribe`

src/common/services/api/order.service.spec.js
```diff
@@ -1206,4 +1208,157 @@ describe('order service', () => {
...
+      it('should throw an error if neither bank account or credit card details are loaded', (done) => {
+        self.orderService.submitOrder(mockController).subscribe(
+          () => done("Observable unexpectedly succeeded"),
+          () => { // error handler
+            expect(self.orderService.submit).not.toHaveBeenCalled()
+            expect(self.orderService.clearCardSecurityCodes).not.toHaveBeenCalled()
+            expect(self.$log.error.logs[0]).toEqual(['Error submitting purchase:', { data: 'Current payment type is unknown' }])
+            expect(mockController.submissionError).toEqual('Current payment type is unknown')
+            expect(self.$window.scrollTo).toHaveBeenCalledWith(0, 0);
+            done();
```
wrandall22, line 1351: Remove semicolons

src/app/branded/step-1/branded-checkout-step-1.component.js
canac: Actually, `subscribe` doesn't return an observable, it returns a subscription object, so `subscribe` should be the last call in the chain.

wrandall22: It just feels so wrong....oh well

caleballdrin:
@wrandall22 can I merge this into [EP-2517-branded-checkout-improvements](https://github.com/CruGlobal/give-web/pull/1117) ?

wrandall22 (approved):
I typically prefer not to have merge commits like 346ee7b0ab86dd8137c9abf4bac5e3bff4a2f6c6, but the rest of the commits were well structured, good job!

PR #1120, opened 2024-11-26
EP-2525 - Hide unnecessary fields

wjames111:
There is a desire to reduce friction by removing fields that are unnecessary for checkout to work properly. The new config option should hide/show all of these fields. Currently proposed fields:
- Middle Initial
- ‘Optional’ section - ie notes
-Additional your information header
- Suffix
- Phone Number
- Spouse information

Jira Ticket [EP-2525](https://jira.cru.org/browse/EP-2525)

### DIFF
--- a/src/common/components/contactInfo/contactInfo.tpl.html
+++ b/src/common/components/contactInfo/contactInfo.tpl.html
@@ -32,7 +32,7 @@
     <p><button id="retryButton1" type="button" class="btn btn-primary" ng-click="$ctrl.loadDonorDetails()" translate>{{'RETRY'}}</button></p>
   </div>
 
-  <div class="mb_x">
+  <div class="mb_x" ng-if="!$ctrl.nameFieldsDisabled">
     <div class="row">
       <div class="col-md-6">
         <a id="householdTypeLink" href="" class="tab-btn btn btn-default btn-block" ng-click="$ctrl.donorDetails['donor-type'] = 'Household'" ng-class="{'on': $ctrl.donorDetails['donor-type'] === 'Household'}" translate>{{'GIVE_AS_INDIVIDUAL'}}</a>


src/common/components/addressForm/addressForm.component.js
```diff
@@ -9,12 +9,14 @@ const componentName = 'addressForm'
 
 class AddressFormController {
   /* @ngInject */
-  constructor ($log, geographiesService) {
+  constructor ($scope, $log, geographiesService) {
     this.$log = $log
     this.geographiesService = geographiesService
+    console.log($scope)
```
dr-bizz, line 15: Remove log

src/common/components/addressForm/addressForm.component.js
```diff
@@ -9,12 +9,14 @@ const componentName = 'addressForm'
...
   }
 
   $onInit () {
+
```
dr-bizz, line 19: I don't think you need this line? Unless you think it makes it easier to read

src/app/productConfig/productConfigModal/productConfig.modal.component.js
```diff
@@ -136,6 +136,6 @@ export default angular
     bindings: {
       resolve: '<',
       close: '&',
-      dismiss: '&'
+      dismiss: '&',
```
dr-bizz, line 139: I think normally we don't add a comma on the last line.  but i might be wrong

dr-bizz (commented):
These are my intially thoughts. Nothing crazy, but code styling. I will test the PR on my local.

src/app/designationEditor/designationEditor.tpl.html
```diff
@@ -45,8 +45,11 @@ <h2 class="mb0">
             Page Options
           </button>
 
-          <product-config product-code="{{$ctrl.designationNumber}}"
-                          button-label="{{'Give a Gift' | translate}}"></product-config>
+          <product-config 
+            product-code="{{$ctrl.designationNumber}}"
+            button-label="{{'Give a Gift' | translate}}"
+            >
+          </product-config>
```
wrandall22, line 52: No need for this change.

src/app/branded/step-1/branded-checkout-step-1.tpl.html
```diff
@@ -30,7 +29,10 @@ <h3 class="panel-name" translate>{{'YOUR_INFORMATION'}}</h3>
         on-submit="$ctrl.onContactInfoSubmit(success)"
         donor-details="$ctrl.donorDetails"
         radio-station-api-url="$ctrl.radioStationApiUrl"
-        radio-station-radius="$ctrl.radioStationRadius">
+        radio-station-radius="$ctrl.radioStationRadius"
+        use-v3="$ctrl.useV3"
+        >
```
wrandall22, line 34: Let's put the `>` back up with the last property

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -214,7 +214,7 @@ <h4 class="panel-title border-bottom-small mt0 pt-" translate>
         </div><!-- // panel -->
       </div>
     </div>
-    <div class="panel panel-default give-modal-panel">
+    <div ng-if="$ctrl.useV3 !== 'true'" class="panel panel-default give-modal-panel">
```
wrandall22, line 217: One concern I realize I have with this is our datadog tests. These rely on the special instructions in order to work. We'll need to make sure that our tests are pointing at a v2 branded checkout.

src/app/productConfig/productConfigModal/productConfig.modal.tpl.html
```diff
@@ -14,7 +14,8 @@
       default-frequency="$ctrl.defaultFrequency"
       update-query-param="$ctrl.updateQueryParam(key, value)"
       submitted="$ctrl.submitted"
-      on-state-change="$ctrl.onStateChange(state)">
+      on-state-change="$ctrl.onStateChange(state)"
+      >
```
wrandall22, line 18: This change should be reverted.

src/common/components/addressForm/addressForm.tpl.html
```diff
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-sm-12">
-    <div class="form-group is-required" ng-class="{'has-error': ($ctrl.parentForm.addressCountry | showErrors) || $ctrl.loadingCountriesError}">
+    <div ng-if="$ctrl.useV3 !== 'true'" class="form-group is-required" ng-class="{'has-error': ($ctrl.parentForm.addressCountry | showErrors) || $ctrl.loadingCountriesError}">
```
wrandall22, line 3: We need to make sure this defaults properly to US without the field being here. I'm still not sold on this change from the stakeholders.

src/common/components/contactInfo/contactInfo.component.js
```diff
@@ -191,6 +191,7 @@ export default angular
       donorDetails: '=?',
       onSubmit: '&',
       radioStationApiUrl: '<',
-      radioStationRadius: '<'
+      radioStationRadius: '<',
+      useV3: '<?'
```
wrandall22, line 195: Is there a reason you didn't use `<`

src/common/components/addressForm/addressForm.component.js
```diff
@@ -69,6 +69,7 @@ export default angular
       address: '=',
       parentForm: '<',
       onAddressChanged: '&',
-      addressDisabled: '<'
+      addressDisabled: '<',
+      useV3: '<?'
```
wrandall22, line 73: Is there a reason you didn't use `<`

wrandall22 (changes requested):

src/common/components/addressForm/addressForm.tpl.html
dr-bizz: If it's defaulting to the US, why remove it, as many users won't need to engage with it? Or is it that the stakeholders want to remove any field that could be unnecessary?

dr-bizz (changes requested):
We need to ensure the country is present on this form.

When I enter my British address, I'm unable to change the country, which means the input fields doesn't change to the International input fields. This means it requires me to select a US State, which is wrong.

`69 Drayton Place,` (Address)
`Irthlingborough,` (city)
`Northamptonshire,` (County)
`NN9 5TD,` (post code)
`England` (country)

dr-bizz:
This feedback needs to go back to the stakeholders, and we need to come up with that 1. removes unnecessary fields without removing one users' needs and 2. allows the user to enter the field if they desire.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -121,7 +121,7 @@ <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMAT
 
       </div>
 
-      <div class="row" ng-if="$ctrl.donorDetails['donor-type'] === 'Household'">
+      <div class="row" ng-if="$ctrl.donorDetails['donor-type'] === 'Household' && $ctrl.useV3 !== 'true'">
```
dr-bizz, line 124: What about the organization's name? What if a company donates and has its offices in an office park (In a building with lots of other offices)? They would likely need to add their organization name so they can receive post

wrandall22: I think a case can be made here that certain campaigns will only be sent to individuals and use the new version which hides the organization data. It is possible that the stakeholders will want more granularity for these configs, but that should come out in UAT.

src/common/components/addressForm/addressForm.component.js
```diff
@@ -69,6 +69,6 @@ export default angular
       address: '=',
       parentForm: '<',
       onAddressChanged: '&',
-      addressDisabled: '<'
+      addressDisabled: '<',
```
wrandall22: Let's remove this change.

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -32,7 +32,7 @@
     <p><button id="retryButton1" type="button" class="btn btn-primary" ng-click="$ctrl.loadDonorDetails()" translate>{{'RETRY'}}</button></p>
   </div>
 
-  <div class="mb_x" ng-if="!$ctrl.nameFieldsDisabled">
+  <div class="mb_x">
```
wrandall22, line 35: We need to keep this conditional.

wrandall22 (changes requested):

src/common/components/contactInfo/contactInfo.tpl.html
```diff
@@ -45,7 +45,7 @@
 
   <form novalidate ng-keyup="$event.key === 'Enter' && $ctrl.submitDetails()" name="$ctrl.detailsForm">
     <div class="mb">
-      <h4 class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMATION'}}</h4>
+      <h4 ng-if="$ctrl.useV3 !== 'true'" class="panel-title  border-bottom-small  visible" translate>{{'YOUR_INFORMATION'}}</h4>
```
wrandall22, line 48: We can't hide the first name field.

wrandall22 (changes requested):

src/common/components/contactInfo/contactInfo.tpl.html
wrandall22: I gotcha. I think you did the right thing here.

wrandall22 (approved):
It may be worth updating the PR description to note which fields this PR is actually hiding.

PR #1123, opened 2024-12-02
Reenable Recaptcha

wrandall22:
This PR re-enables Recaptcha, but with some extra checks to ensure we don't fail when Recaptcha returns us a message instead of data.

### DIFF
--- a/src/common/components/Recaptcha/Recaptcha.test.tsx
+++ b/src/common/components/Recaptcha/Recaptcha.test.tsx
@@ -45,7 +45,7 @@ describe('Recaptcha component', () => {
     expect(recaptchaEnabledButton.innerHTML).toEqual('Label')
   })
 
-  xit('should disable the button until ready', async () => {
+  it('should disable the button until ready', async () => {
     global.window.grecaptcha = undefined
 
     const { getByRole } = render(buildRecaptcha())
@@ -56,7 +56,7 @@ describe('Recaptcha component', () => {
     await waitFor(() => expect((recaptchaEnabledButton as HTMLButtonElement).disabled).toEqual(false))
   })
 
-  xit('should successfully pass the recaptcha', async () => {
+  it('should successfully pass the recaptcha', async () => {
     //@ts-ignore
     global.fetch = jest.fn(() => {
       return Promise.resolve({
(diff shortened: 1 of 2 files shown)


src/common/components/Recaptcha/Recaptcha.tsx
```diff
@@ -71,50 +71,50 @@ export const Recaptcha = ({
...
+      try {
+        const token = await grecaptcha.execute(recaptchaKey, { action: action })
+        const serverResponse = await fetch(`${apiUrl}/recaptcha/verify`, {
+          method: 'POST',
+          body: JSON.stringify({ token: token }),
+          headers: { 'Content-Type': 'application/json' }
+        })
+        const data = await serverResponse.json()
+
+        if (!data || !data.score || !data.action) {
```
dr-bizz, line 89: Are we positive that the action was returning `undefined` and not `"undefined"` (as a string)?

wrandall22:
Billing has already been enabled so we shouldn't hit this specific error again.

src/common/components/Recaptcha/Recaptcha.test.tsx
```diff
@@ -238,7 +238,51 @@ describe('Recaptcha component', () => {
...
+
+    onSuccess.mockImplementation(() => console.log('success after weird'))
+
+    const { getByRole } = render(
+      buildRecaptcha()
+    )
+
+    await userEvent.click(getByRole('button'))
+    await waitFor(() => {
+      expect(onSuccess).toHaveBeenCalledTimes(1)
```
dr-bizz, line 261: You could write, but I'm not too fussy about it as this is urgent and the other tests also have this one.
```suggestion
      expect(onSuccess).toHaveBeenCalled()
```

dr-bizz (approved):

src/common/components/Recaptcha/Recaptcha.tsx
dr-bizz: I see. `action` itself is undefined.

wrandall22: I'm about 90% sure; I think it was something like `{ message: "Over Enterprise free quota. Please ensure your project has an active billing account" }` as a response.

src/common/components/Recaptcha/Recaptcha.test.tsx
wrandall22: I like having it as 1 time to ensure the return statements are in effect and not double submitting the gift.

dr-bizz: That's fine my me

PR #1124, opened 2024-12-04
[HS-1269093] Ignore giving links without URLs

canac:
Some designation accounts have multiple empty giving links on them, which causes issues in the Give a Gift modal. As we try to figure out how they are getting added, this PR ignores them in the UI.

🔀 Manually merged into staging and tested in staging

Jeff's account doesn't have the extra giving links in staging, so I copied the production response to `https://give.cru.org/content/give/us/en/designations/0/9/5/0/5/0950547.infinity.json` and made a local override for give-stage2.cru.org to use the same response.

https://secure.helpscout.net/conversation/2782484877/1269093/


src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
```diff
@@ -195,6 +195,15 @@ describe('product config form component', function () {
       expect($ctrl.analyticsFactory.giveGiftModal($ctrl.productData))
     })
 
+    it('should filter out givingLinks with empty urls', () => {
+      const validLink = { url: 'https://give.cru.org/1234567' }
+      jest.spyOn($ctrl.designationsService, 'givingLinks').mockReturnValue(Observable.of([{}, {}, validLink]))
```
wrandall22, line 200: I think it would be nicer to have more realistic data in this test. An example empty item looks like this:

```
{
  name: undefined
  order: 0
  url: undefined
}
```

An example valid item would have all 3 properties.

canac:
@wrandall22 I realized that it makes more sense to have sanitization happen in the designation service instead of in the controller.

src/app/productConfig/productConfigForm/productConfigForm.component.spec.js
canac: I ended up removing this test, but I made sure that new test uses data that looks like Jeff's in the test case.

src/common/services/api/designations.service.js
```diff
@@ -253,6 +253,10 @@ class DesignationsService {
           // Map giving links
           if (data.data['jcr:content'].givingLinks) {
             angular.forEach(data.data['jcr:content'].givingLinks, (v, k) => {
+              if (!v || !v.name || !v.url) {
+                return
```
wrandall22, line 257: Should this instead be `continue`? How will this work if the first item is invalid, but the second is valid?

src/common/services/api/designations.service.spec.js
```diff
@@ -209,6 +209,20 @@ describe('designation service', () => {
         })
       self.$httpBackend.flush()
     })
+
+    it('should ignore givingLinks without names or urls', () => {
+      const response = angular.copy(designationResponse)
+      response['jcr:content'].givingLinks.item1 = { 'jcr:primaryType': 'nt:unstructured' }
```
wrandall22, line 215: We can test my question above by also adding an `item2` that is valid and ensuring it is in the response.

src/common/services/api/designations.service.js
canac: If it were a `for` loop, we'd want to use `continue`. But since it's a `forEach` callback, we can return early and `angular.forEach` will call our callback again with the next item.

src/common/services/api/designations.service.spec.js
```diff
@@ -209,6 +209,22 @@ describe('designation service', () => {
...
+      response['jcr:content'].givingLinks.item2 = { 'jcr:primaryType': 'nt:unstructured', url: 'https://example2.com', name: 'Name 2' }
+      self.$httpBackend.expectGET('https://give-stage2.cru.org/content/give/us/en/designations/0/1/2/3/4/0123456.infinity.json')
+        .respond(200, response)
+      self.designationsService.givingLinks('0123456')
+        .subscribe(givingLinks => {
+          expect(givingLinks).toEqual([
+            { name: 'Name', url: 'https://example.com', order: 0 },
+            { name: 'Name', url: 'https://example.com', order: 2 }
```
canac, line 223: Note that `order: 1` is skipped. `order` is used by the edit giving links modal to sort the links, so missing numbers doesn't matter there. Interestingly, the giving links aren't sorted in the product config modal.

canac:
I discovered that none of the assertions in the designations service tests were being run because the tests were completing synchronously. I updated them to use `done` to wait to finish the test until the subscribe handler is called.

wrandall22 (changes requested):

src/common/services/api/designations.service.spec.js
```diff
@@ -90,58 +92,63 @@ describe('designation service', () => {
...
     })
 
-    it('should handle an empty response', () => {
+    it('should handle an empty response', done => {
       self.$httpBackend.expectPOST('https://give-stage2.cru.org/cortex/itemselections/crugive/a5t4fmspmhbkez6cwbnd6mrkla74hdgcupbl4xjb=/options/izzgk4lvmvxgg6i=/values/jzaq=/selector?FollowLocation=true&zoom=code,offer:code,definition,definition:options:element:selector:choice,definition:options:element:selector:choice:description,definition:options:element:selector:choice:selectaction,definition:options:element:selector:chosen,definition:options:element:selector:chosen:description')
         .respond(200, '')
       self.designationsService.productLookup('/itemselections/crugive/a5t4fmspmhbkez6cwbnd6mrkla74hdgcupbl4xjb=/options/izzgk4lvmvxgg6i=/values/jzaq=/selector', true)
-        .subscribe(() => {
-          fail('success should not have been called')
+        .subscribe(done => {
```
wrandall22, line 126: This `done` param is unnecessary

src/common/services/api/designations.service.spec.js
```diff
@@ -90,58 +92,63 @@ describe('designation service', () => {
...
+    it('should get product details for a designation number', done => {
       self.$httpBackend.expectPOST('https://give-stage2.cru.org/cortex/items/crugive/lookups/form?FollowLocation=true&zoom=code,offer:code,definition,definition:options:element:selector:choice,definition:options:element:selector:choice:description,definition:options:element:selector:choice:selectaction,definition:options:element:selector:chosen,definition:options:element:selector:chosen:description',
         { code: '0354433' })
         .respond(200, lookupResponse)
       self.designationsService.productLookup('0354433')
         .subscribe((data) => {
           expect(data).toEqual(expectedResponse)
-        })
+          done()
+        }, done)
```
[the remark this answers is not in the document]
canac: I'm actually intentionally calling `done` with the error if the observable errors. That will immediately fail the test and print the error instead of the test eventually timing out without printing any error information.

That pattern is described here: https://jestjs.io/docs/asynchronous#callbacks

src/common/services/api/designations.service.spec.js
```diff
@@ -90,58 +92,63 @@ describe('designation service', () => {
...
 
-    it('should handle an empty response', () => {
+    it('should handle an empty response', done => {
       self.$httpBackend.expectPOST('https://give-stage2.cru.org/cortex/itemselections/crugive/a5t4fmspmhbkez6cwbnd6mrkla74hdgcupbl4xjb=/options/izzgk4lvmvxgg6i=/values/jzaq=/selector?FollowLocation=true&zoom=code,offer:code,definition,definition:options:element:selector:choice,definition:options:element:selector:choice:description,definition:options:element:selector:choice:selectaction,definition:options:element:selector:chosen,definition:options:element:selector:chosen:description')
         .respond(200, '')
       self.designationsService.productLookup('/itemselections/crugive/a5t4fmspmhbkez6cwbnd6mrkla74hdgcupbl4xjb=/options/izzgk4lvmvxgg6i=/values/jzaq=/selector', true)
-        .subscribe(() => {
-          fail('success should not have been called')
+        .subscribe(done => {
+          done('success should not have been called')
```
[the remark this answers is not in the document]
canac: Same thing here, calling `done` with an argument fails the test. That's what we want in this case. But I'll switch it back to `fail` since some other tests do it that way.

src/common/services/api/designations.service.spec.js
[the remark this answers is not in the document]
wrandall22: Then we should call `fail` instead of `done` because otherwise it will succeed and give us false test successes.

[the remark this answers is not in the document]
canac: Apparently, `fail` isn't defined in jest tests. It seems that that is a jasmine thing. Calling `fail()` in a test will technically fail the test, but only because it throws:

```
ReferenceError: fail is not defined
```

I probably wasn't explaining it well before but calling `done()` will complete the test successfully, but calling `done(err)` or `done("message")` will fail the test immediately. When I do `observable.subscribe(() => {}, done)`, that is equivalent to `observable.subscribe(() => {}, err => done(err))`. If the observable errors, `done` will be called with an argument, and the test will immediately fail, which is what we want.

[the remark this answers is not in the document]
wrandall22: Ah right, forgot about that `done(message)` implementation. I suppose I'm still more used to Jasmine than Jest.

wrandall22 (approved):

wrandall22:
We should probably let Mike Nelson know about these changes going out as well.

PR #1125, opened 2024-12-06
EP-2555 - Improve Error Logging to DataDog RUM

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2555)

This adds a few error logs that get sent to DD RUM in a way that shows up in [Error Tracking](https://app.datadoghq.com/error-tracking?query=%40application.id%3A3937053e-386b-4b5b-ab4a-c83217d2f953%20%22Error%20submitting%20purchase%3A%2A%22&fromUser=false&refresh_mode=sliding&source=browser&from_ts=1732889704580&to_ts=1733494504580&live=true). The console errors show up in the logs, but we can't create monitors based on logs, only based on errors. This will allow creation of a monitor like [this one](https://app.datadoghq.com/monitors/159738138) so we can watch for high numbers of checkout errors.

This is all in response to the Sev1 on Saturday (11/30/24) where there were front-end checkout errors present for 9 hours before we heard about them.

### DIFF
--- a/src/app/checkout/step-3/step-3.component.js
+++ b/src/app/checkout/step-3/step-3.component.js
@@ -176,7 +176,7 @@ class Step3Controller {
         error.config.data['security-code'] = error.config.data['security-code'].replace(/./g, 'X') // Mask security-code
       }
       componentInstance.$log.error('Error submitting purchase:', error)
-      datadogRum.addError(new Error(`Error submitting purchase: ${JSON.stringify(error)}`)) // here in order to show up in Error Tracking in DD
+      datadogRum.addError(new Error(`Error submitting purchase: ${JSON.stringify(error)}`), { context: 'Checkout Submission', errorCode: error.status }) // here in order to show up in Error Tracking in DD
       componentInstance.onSubmitted()
       componentInstance.submissionErrorStatus = error.status
       componentInstance.submissionError = isString(error && error.data) ? (error && error.data).replace(/[:].*$/, '') : 'generic error' // Keep prefix before first colon for easier ng-switch matching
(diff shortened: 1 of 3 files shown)


src/app/checkout/step-3/step-3.component.js
```diff
@@ -175,6 +176,7 @@ class Step3Controller {
         error.config.data['security-code'] = error.config.data['security-code'].replace(/./g, 'X') // Mask security-code
       }
       componentInstance.$log.error('Error submitting purchase:', error)
+      datadogRum.addError(new Error(`Error submitting purchase: ${JSON.stringify(error)}`)) // here in order to show up in Error Tracking in DD
```
wrandall22, line 179: I opted to not tweak the test that checks `$log.error` because I found it not worth the time (mocking was not playing nicely).

src/common/components/Recaptcha/Recaptcha.tsx
```diff
@@ -94,7 +95,9 @@ export const Recaptcha = ({
 
         if (data?.success === true && isValidAction(data?.action)) {
           if (data.score < 0.5) {
-            $log.warn(`Captcha score was below the threshold: ${data.score}`)
+            const errorMessage = `Captcha score was below the threshold: ${data.score}`
+            $log.warn(errorMessage)
+            datadogRum.addError(new Error(`Error submitting purchase: ${errorMessage}`))
```
dr-bizz, line 100: To help with debugging, you could add context to the error logs. But this isn't required.
```suggestion
            datadogRum.addError(new Error(`Error submitting purchase: ${errorMessage}`),
            { context: 'Recaptcha flow', errorCode: "lowScore" }
            )
```

src/app/checkout/step-3/step-3.component.js
dr-bizz: I think the main thing here is that we test to ensure we get an error when an issue arises, and the test does that.

dr-bizz (approved):
This looks good. I haven't tested the code locally, I can do if needed.

But as I'm out for the rest of the day, I will approve it to prevent you from getting blocked.

wrandall22:
The Error Tracking link above shows my tests from localhost. To test on local, one would need to set the `DATADOG_RUM_CLIENT_TOKEN` environment variable, update the `app.config.js` file to capture 100% of the sessions when env = `development`, then run tests on Chrome (I couldn't get Firefox to do it for some reason, maybe I wasn't on the https url).

PR #1126, opened 2024-12-11
EP-2558 - Update radio stations API

wrandall22:
[Jira](https://jira.cru.org/browse/EP-2558)

Part of sunsetting Studio Enterprise means reworking the radio stations API. Nobody is using it in BCO right now, but we want the code to work when they do.

Note: Currently the first call to the API takes a long time; we may want to put some sort of loading indicator in the radio station dropdown.

### DIFF
--- a/src/common/services/api/order.service.js
+++ b/src/common/services/api/order.service.js
@@ -372,8 +372,8 @@ class Order {
   }
 
   storeRadioStationData (radioStationData) {
-    this.sessionStorage.setItem('radioStationName', radioStationData.Description)
-    this.sessionStorage.setItem('radioStationCallLetters', radioStationData.MediaId)
+    this.sessionStorage.setItem('radioStationName', Object.values(radioStationData)[0])
+    this.sessionStorage.setItem('radioStationCallLetters', Object.keys(radioStationData)[0])
   }
 
   retrieveRadioStationName () {
--- a/README.md
+++ b/README.md
@@ -58,8 +58,7 @@ Add the following code to your page where appropriate. See the [Branded checkout
     show-cover-fees="true"
     on-order-completed="$event.$window.onOrderCompleted($event.purchase)"
     on-order-failed="$event.$window.onOrderFailed($event.donorDetails)"
-    radio-station-api-url="https://api.domain.com/getStations"
-    radio-station-radius="100">
+    radio-station-api-url="https://api.domain.com/getStations">
 </branded-checkout>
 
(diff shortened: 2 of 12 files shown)


wrandall22:
@clonge in case you want to review any of the changes I made here for the new radio station API response model structure.

clonge:
Looks good to me. What you have should work, but not sure if you want to accommodate FL's defaults? The current form has the following default options:
```
<option value="ZWEB">FamilyLife.com</option>
<option value="ZPOD">Podcasts</option>
<option value="ZYOU videos">YouTube videos</option>
<option value="RADIO">Radio</option>
<option value="ONE">OnePlace.com</option>
<option value="ZMFL">My FamilyLife App</option>
<option value="ZZZZ">None of the above yet!</option>
```
These are there in case the api fails for any reason. In the current form the radio station field is required, but FL wants the user to still be able to complete the donation even if the api fails. I'm not sure the best way to implement these fallback values in your code. These default options are returned by the api when it succeeds.

wrandall22:
One reality is that this is technically meant to be usable by any ministry, not just FL (I know this means they would need to match the same API schema), which makes me a bit hesitant to add defaults on the client side which are specific to FL.

wrandall22:
If it's worth the trouble, I could have them configure defaults in the form of
```
[
  { callLetter: 'name' },
  ...
]
```

src/common/services/api/order.service.js
```diff
@@ -372,8 +372,8 @@ class Order {
   }
 
   storeRadioStationData (radioStationData) {
-    this.sessionStorage.setItem('radioStationName', radioStationData.Description)
-    this.sessionStorage.setItem('radioStationCallLetters', radioStationData.MediaId)
+    this.sessionStorage.setItem('radioStationName', Object.values(radioStationData)[0])
+    this.sessionStorage.setItem('radioStationCallLetters', Object.keys(radioStationData)[0])
```
dr-bizz, line 376: I assume we know the names of these properties or they change?

clonge:
Yep, understood. I don't think it's critical. We will need to test thoroughly when/if FL switches to branded-checkout anyway. If it's a big issue they will bring it up then.

dr-bizz (commented):
I started to review this, but I need more time to do a deep dive.

src/common/services/api/order.service.js
wrandall22: The names of the properties change based on zip code/API response.

dr-bizz (approved):
The code looks good!

PR #1128, opened 2025-01-07
Update to Recaptcha Enterprise

wrandall22:
This PR switches from Recaptcha v3 to Recaptcha Enterprise. The scoring has moved from an AEM API call to the Cortex Gateway, so we will not know it in real time, but on gift submission. We also changed the actions to match up to one pre-existing Google defined action and a similar wording for branded.

### DIFF
--- a/src/common/services/checkoutHelpers/checkout.service.js
+++ b/src/common/services/checkoutHelpers/checkout.service.js
@@ -1,0 +1,8 @@
+export function initializeRecaptcha () {
+  const script = this.$window.document.createElement('script')
+  script.src = `https://www.google.com/recaptcha/enterprise.js?render=${this.envService.read('recaptchaKey')}`
+  script.id = 'give-checkout-recaptcha'
+  if (!this.$window.document.getElementById(script.id)) {
+    this.$window.document.head.appendChild(script)
+  }
+}
--- a/src/common/services/checkoutHelpers/checkout.service.js
+++ b/src/common/services/checkoutHelpers/checkout.service.js
@@ -1,8 +1,24 @@
-export function initializeRecaptcha () {
-  const script = this.$window.document.createElement('script')
-  script.src = `https://www.google.com/recaptcha/enterprise.js?render=${this.envService.read('recaptchaKey')}`
-  script.id = 'give-checkout-recaptcha'
-  if (!this.$window.document.getElementById(script.id)) {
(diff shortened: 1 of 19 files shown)


wrandall22:
Plan is to go live on Tuesday 1/14 at 2pm.

src/common/services/checkoutHelpers/checkout.service.js
```diff
@@ -0,0 +1,8 @@
+export function initializeRecaptcha () {
```
canac, line 1: Can we reduce the coupling between this method and the controllers that use it by passing in `$window` and `envService` as arguments?

Or we could make this a real Angular service and inject `$window` and `envService` as dependencies.

src/common/services/checkoutHelpers/checkout.service.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    afterEach(() => {
+      const foundScript = self.$window.document.getElementById('give-checkout-recaptcha')
+      if (foundScript) {
+        document.head.removeChild(foundScript)
+      }
+    })
+
+    it('should add a script even if one already exists', () => {
+      self.$window.document.head.appendChild(script)
+      self.checkoutService.initializeRecaptcha.call(self)
```
canac, line 42: And in the following test

```suggestion
      self.checkoutService.initializeRecaptcha()
```

src/common/services/checkoutHelpers/checkout.service.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  describe('initializeRecaptcha()', () => {
+    beforeEach(() => {
+      script.src = 'https://www.google.com/recaptcha/enterprise.js?render=123'
+      script.id = 'test-script'
+    })
+
+    afterEach(() => {
+      const foundScript = self.$window.document.getElementById('give-checkout-recaptcha')
+      if (foundScript) {
+        document.head.removeChild(foundScript)
```
canac, line 36: Is there a reason why in some places we're using `self.$window.document` and in some places were using `document` directly? Could we pick one or the other?

wrandall22: Sorry, left over from debugging.

canac (approved):
Looks great! I see the Recaptcha token and action in the request in staging for regular checkout and branded checkout.

PR #1129, opened 2025-01-07
EP-2560

wjames111:
- have all Contact information (FN, LN, *ORG Name, *Spouse name, Email, and Phone) under "Contact information". 
- update the text to read "Contact Information", instead of "Your information"  
- Jira ticket available [here](https://jira.cru.org/browse/EP-2560)


wrandall22:
Code looks fine, let's get it on staging. For future, it might be nice to remove the duplicate code for email address and put it into a component or directive.

PR #1130, opened 2025-01-08
[EP-2362] Save contact info after sign up

canac:
# Changes

* Replace the `onStateChange` binding with dedicated `onSignUp` and `onSignIn` bindings
* After sign up, use the order service to save the donor details to the user's account. If saving fails, show the contact info form with the fields from sign up pre-populated.

I made this a PR instead of committing directly so it's easier for you to see what changed and see if I'm going in the right direction with this.

### DIFF
--- a/src/common/components/contactInfoModal/contactInfoModal.tpl.html
+++ b/src/common/components/contactInfoModal/contactInfoModal.tpl.html
@@ -1,4 +1,8 @@
-<contact-info submitted="$ctrl.submitted" on-submit="$ctrl.onSubmit(success)"></contact-info>
+<contact-info
+  submitted="$ctrl.submitted"
+  on-submit="$ctrl.onSubmit(success)"
+  donor-details="$ctrl.signUpDonorDetails"
+></contact-info>
 <div class="row">
   <div class="col-sm-5 hidden-xs">
     <button id="cancelButtonTop" class="btn btn-default" ng-click="$ctrl.onCancel()" translate>Cancel</button>
--- a/src/common/components/registerAccountModal/registerAccountModal.component.js
+++ b/src/common/components/registerAccountModal/registerAccountModal.component.js
@@ -87,6 +87,10 @@ class RegisterAccountModalController {
     this.stateChanged('sign-up')
   }
 
+  onSignIn () {
+    this.stateChanged('sign-in')
+  }
+
   onIdentitySuccess () {
     this.sessionService.removeOktaRedirectIndicator()
 
--- a/src/common/components/registerAccountModal/registerAccountModal.component.js
+++ b/src/common/components/registerAccountModal/registerAccountModal.component.js
@@ -13,6 +13,14 @@ import userMatchModal from 'common/components/userMatchModal/userMatchModal.comp
 import contactInfoModal from 'common/components/contactInfoModal/contactInfoModal.component'
 import failedVerificationModal from 'common/components/failedVerificationModal/failedVerificationModal.component'
 
+import { Observable } from 'rxjs/Observable'
+import 'rxjs/add/observable/forkJoin'
+import 'rxjs/add/observable/of'
+import 'rxjs/add/operator/do'
+import 'rxjs/add/operator/map'
+import 'rxjs/add/operator/switchMap'
+import merge from 'lodash/merge'
+
 const componentName = 'registerAccountModal'
 
 class RegisterAccountModalController {
@@ -42,7 +50,7 @@ class RegisterAccountModalController {
 
   $onInit () {
     this.$rootScope.$on(LoginOktaOnlyEvent, () => {
-      this.getDonorDetails()
+      this.checkDonorDetails()
     })
 
     // Ensure loading icon isn't rendered on screen.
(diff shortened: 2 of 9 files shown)


src/common/components/contactInfoModal/contactInfoModal.tpl.html
```diff
@@ -1,4 +1,4 @@
-<contact-info submitted="$ctrl.submitted" on-submit="$ctrl.onSubmit(success)"></contact-info>
+<contact-info submitted="$ctrl.submitted" on-submit="$ctrl.onSubmit(success)" donor-details="$ctrl.signUpDonorDetails"></contact-info>
```
dr-bizz, line 1: Where is `signUpDonorDetails` defined? Were you also meant to commit changes to the contactInfoModal.component.js

src/common/components/registerAccountModal/registerAccountModal.tpl.html
```diff
@@ -31,9 +31,8 @@ <h3 translate>Create Your Account</h3>
   ></sign-in-modal>
   <sign-up-modal
     ng-switch-when="sign-up"
-    on-state-change="$ctrl.stateChanged(state)"
-    on-success="$ctrl.onIdentitySuccess()"
-    on-failure="$ctrl.onIdentityFailure()"
+    on-sign-up="$ctrl.onSignUp(donorDetails)"
+    on-sign-in="$ctrl.onSignIn()"
```
dr-bizz, line 35: You also need to update the file `src/common/services/session/sessionModal.tpl.html` as signUpModal is also used there

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -148,13 +148,36 @@ class SignUpModalController {
             }
             onSuccess(postData)
           }
+        },
+        // TODO: make sure that this callback is only called after the sign up completes successfully
+        postSubmit: (postData, onSuccess) => {
```
dr-bizz, line 153: It looks like this is fired after the Registration API is fired to Okta, but before registration is completed. I think that is okay, bt as you said, we will need to check

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -148,13 +148,36 @@ class SignUpModalController {
...
+          this.$scope.$apply(() => this.onSignUp({ donorDetails }))
         }
       }
     })
 
     this.signIn()
 
     this.oktaSignInWidget.on('afterRender', this.afterRenderFn.bind(this))
+
+    this.$scope.$apply(() => this.onSignUp())
```
dr-bizz, line 180: Could you explain why you need to run `this.onSignUp()` here?

dr-bizz (approved):
It looks great! I have a few questions, but I don't want to block you so I will approve the PR.

src/common/components/signUpModal/signUpModal.component.js
canac: Good catch! That was accidentally left in for testing

src/common/components/contactInfoModal/contactInfoModal.tpl.html
canac: It is set in `checkDonorDetails` if there was an error saving the donor details

src/common/components/registerAccountModal/registerAccountModal.tpl.html
canac: Good catch, that should be fixed now

canac:
@dr-bizz Sorry for the force-push. I rebased on top of #1132, combined several fixup commits, and added the organization fields you added. I'll go through it myself more in tomorrow morning to see if I missed anything, but I'm pushing for you to take a look at too when you get a chance.

src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -117,7 +134,29 @@ class RegisterAccountModalController {
...
+            // If there was an error, save the donor details from sign up so that they will be added
+            // to the contact info form. The error handler below will change the step to contact-info.
+            this.signUpDonorDetails = signUpDonorDetails
+          }
+        })
+      }
+
+      // Pass donorDetails to the next step
+      return Observable.of(donorDetails)
+    }).subscribe({
```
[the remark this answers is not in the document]
dr-bizz: I switched `do` to `catch` as it was easier to read, but these changes are just suggestions based on preference. 

You are correct, you don't need to write `Observable.throw(error)` if you write `do`

src/common/components/registerAccountModal/registerAccountModal.component.spec.js
```diff
@@ -169,16 +169,16 @@ describe('registerAccountModal', function () {
...
       jest.spyOn($ctrl, 'stateChanged').mockImplementation(() => {})
     })
 
-    describe('orderService.getDonorDetails success', () => {
+    describe('orderService.checkDonorDetails success', () => {
       describe('\'registration-state\' COMPLETED', () => {
         it('changes state to \'contact-info\'', () => {
           $ctrl.orderService.getDonorDetails.mockImplementation(() => Observable.of({ 'registration-state': 'COMPLETED' }))
-          $ctrl.getDonorDetails()
+          $ctrl.checkDonorDetails()
```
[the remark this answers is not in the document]
canac: Good idea. I added some tests.

PR #1133, opened 2025-01-13
EP-2560-QA-2

wjames111:
Address additional requests made in [EP-2517](https://github.com/CruGlobal/give-web/pull/1117)

- in `src/app/branded/step-1/branded-checkout-step-1.tpl.html` move `>` line up
- account for no phone number on the form when adding custom validators on v3.
- First and Last should take up half the space as other lines do when having 2 fields.
- in `src/common/components/contactInfo/emailField/emailField.tpl.html` change file's indentation to 2 spaces.


wrandall22 (changes requested):
Just one thing I saw for this.

PR #1134, opened 2025-01-14
[EP-2362] Remove session modal's signup modal

canac:
`sessionModalService.createAccount` was only being used in one place: `onSignUpWithOkta` in `signIn.component.js`. Upon careful review, that function is never callable because `sign-in-form` hides the create account link when `on-sign-in-page="true"`. I removed that prop from the component and then removed `createAccount` from the session modal because it is no longer being used.

### DIFF
--- a/src/common/services/session/sessionModal.service.js
+++ b/src/common/services/session/sessionModal.service.js
@@ -78,8 +78,7 @@ const SessionModalService = /* @ngInject */ function ($uibModal, $log, modalStat
       },
       backdrop: 'static',
       keyboard: false
-    }).result,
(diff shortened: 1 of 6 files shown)


src/common/services/session/sessionModal.service.js
```diff
@@ -79,7 +79,6 @@ const SessionModalService = /* @ngInject */ function ($uibModal, $log, modalStat
       backdrop: 'static',
       keyboard: false
     }).result,
```
dr-bizz, line 81: Lint might complain about the comma at the end of `result,`

dr-bizz (approved):
Great work!

PR #1135, opened 2025-01-15
[EP-2362] Improvements to register account modal

canac:
* Only call `checkDonorDetails` once in `$onInit`. Previously it was called twice in`$onInit`.
* Skip the contact info modal when `registration-state` is `MATCHED`.
* Show "Checking your donor account" in the modal title while `checkDonorDetails` is running and rename state from `loading` to `loading-donor` for clarity.

### DIFF
--- a/src/common/components/registerAccountModal/registerAccountModal.component.js
+++ b/src/common/components/registerAccountModal/registerAccountModal.component.js
@@ -155,6 +155,8 @@ class RegisterAccountModalController {
         // Workflow Complete if 'registration-state' is COMPLETED
         if (donorDetails['registration-state'] === 'COMPLETED') {
           this.onSuccess()
+        } else if (donorDetails['registration-state'] === 'MATCHED') {
+          this.onContactInfoSuccess()
         } else if (donorDetails['registration-state'] === 'FAILED') {
           this.stateChanged('failed-verification')
         } else {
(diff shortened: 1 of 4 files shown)


src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -62,20 +62,15 @@ class RegisterAccountModalController {
...
-      // Proceed to Step 2
-      this.checkDonorDetails()
-    } else {
-      // Proceed to Step 1.
-      this.stateChanged('sign-in')
-    }
-
-    // If there is a session change, update the state if needed.
+    // Set the state initially and update the state if the session changes
     this.subscription = this.sessionService.sessionSubject.subscribe(() => {
```
dr-bizz, line 66: Ah, this makes more sense, as Subjects are hot observables.

src/common/components/registerAccountModal/registerAccountModal.tpl.html
```diff
@@ -36,7 +37,7 @@ <h3 translate>Create Your Account</h3>
     on-sign-in="$ctrl.onSignIn()"
     is-inside-another-modal="true"
   ></sign-up-modal>
-  <div ng-switch-when="loading">
+  <div ng-switch-when="loading-donor">
```
dr-bizz, line 40: You decided to keep the loading in the registerAccountModal?

dr-bizz (approved):
Great work ln this

src/common/components/registerAccountModal/registerAccountModal.tpl.html
canac: I wanted the "Checking your donor account" message to show up in the title. So I kept the register account modal loading state so that it can change the modal title.

A bunch of the modals inside `registerAccountModal` are setting `modalTitle`. But that variable doesn't seem to be used anywhere in the template, so it could use some refactoring.

PR #1136, opened 2025-01-17
[EP-2362] Remove sessionHandleOktaRedirect

canac:
Remove the `sessionHandleOktaRedirect` service and the code in all components that was using it.


dr-bizz (approved):
Thank you for doing this! This is great!

I have left 1 comment on the PR, as I changed the `designationEditor.component.js` page to use `$rootScope` instead of `$scope` which I think we could change back.

PR #1138, opened 2025-01-24
Adjusts the padding of radio button labels to avoid overlap.

rguinee:
Moves the radio buttons that were overlapping to the right (inside the panel). This change affects radio buttons on the productConfigForm as well (moves them to the right by 20px). 
<br>
image

### DIFF
--- a/src/assets/scss/_give.scss
+++ b/src/assets/scss/_give.scss
@@ -735,3 +735,11 @@ img.full-width {
     text-align: center !important;
   }
 }
+
+.saved-payment-methods {
+  .row {
+      label {
+        padding-left: 15px;
+      }
+  }
+}


src/assets/scss/_gift-config.scss
```diff
@@ -31,7 +31,7 @@ label.custom-amount {
 .checkbox {
   label {
     min-height: 29px;
-    padding-left: 0px;
+    padding-left: 20px;
```
wrandall22, line 34: What is the reason for this change on top of the other? I think the scope of the issue is only in saved payment methods.

rguinee:
The thought was to avoid future instances where a radio input shifts
outside a panel as it did in the saved payment method selection.

wrandall22:
One concern would be that it makes the product config modal no longer line up on custom amounts to the rest of the form, which is left aligned. It makes the eye go back and forth.

rguinee:
Yeah it certainly does that. I can drop the change from that file and make
the padding-left in saved-payment-methods 45px.

src/assets/scss/_give.scss
```diff
@@ -735,3 +735,11 @@ img.full-width {
...
   }
 }
+
+.saved-payment-methods {
+  .row {
+      label {
+        padding-left: 15px;
+      }
+  }
+}
```
wrandall22, line 745: Let's make sure to add a newline to the end of this file (you should set up your IDE to do this automatically).

rguinee: Ah, yes. I had that unchecked. Adjusted that in VS Code and I'll add that to the commit.

src/assets/scss/_give.scss
```diff
@@ -735,3 +735,11 @@ img.full-width {
     text-align: center !important;
   }
 }
+
+.saved-payment-methods {
+  .row {
+      label {
+        padding-left: 15px;
```
wrandall22, line 742: Was this meant to be `45px`?

rguinee: What I realized in testing is that this should account for the -15px margin in the row class. Initially, I assumed it would need to be 45px, but there's discrepancies between testing this in BCO-improvements and in staging. This change simply resolves the 15px margin that comes from the row class. And since this is based from master I didn't want to make adjustments based on changes in another branch. Am I thinking about that the right way?

wrandall22: Yes, for sure. We can add more commits to the other PRs as needed.

wrandall22 (approved):

PR #1141, opened 2025-01-30
[EP-2362] Refactor signOut observable logic

canac:
Until I realized that the sign out bug is in `giveaem` (https://github.com/CruGlobal/giveaem/pull/565), I was working on refactoring the `signOut` observable logic to use chained Observable operators instead of the `new Observable` syntax. Although it doesn't fix anything, I think my changes are still an improvement.

I also fixed many of the `signOut` tests to use `done`. All tests with asynchronous logic need to use `done` to keep the `expect` calls from silently not running.


dr-bizz (approved):
Thank you for improving the signout

PR #1144, opened 2025-02-03
[no-Jira] Only collect test coverage in CI

canac:
Currently, give-web always runs tests with test coverage enabled. This prints the code coverage results, which means I always have to scroll up many lines to see which tests succeeded or failed. It also makes the tests take a little longer to run.

This change makes tests default to not having test coverage enabled, while still enabling test coverage in CI. If developers want to see code coverage info locally, they can always run `yarn test --coverage=true`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
         run: yarn install --immutable --immutable-cache
       
       - name: Run Tests
-        run: yarn test
+        run: yarn test --coverage=true
       
       - name: Upload Codecov Reports
         uses: codecov/codecov-action@v2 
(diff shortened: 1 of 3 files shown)


wrandall22:
Could we add a simple script for running with coverage in package.json? I prefer to run with coverage most times, and would prefer not to have to type out the whole command.

canac:
> I prefer to run with coverage most times, and would prefer not to have to type out the whole command.

If devs frequently run jest with test coverage enabled, we don't have to make this change. I'm just frequently running tests for a specific file, and in that case the coverage info is basically useless because most files in the project have 0% test coverage. It's a lot of useless noise in my workflow, but I don't want to disrupt others' workflows.

wrandall22 (approved):

dr-bizz (approved):

PR #1145, opened 2025-02-04
EP-2562 - Filter invalid CVV

wrandall22:
In order to make the [Datadog Monitor](https://app.datadoghq.com/monitors/159980320) more effective in bringing up issues during checkout on the client side, we need to filter out user errors (such as entering the wrong cvv). If we want to filter out other user errors later, we can refactor the filtering logic.

### DIFF
--- a/src/app/checkout/step-3/step-3.component.js
+++ b/src/app/checkout/step-3/step-3.component.js
@@ -41,6 +41,7 @@ class Step3Controller {
     this.sessionStorage = $window.sessionStorage
     this.selfReference = this
     this.isBranded = envService.read('isBrandedCheckout')
+    this.datadogRum = datadogRum
 
     this.$scope.$on(SignInEvent, () => {
       this.$onInit()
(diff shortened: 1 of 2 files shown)


src/app/checkout/step-3/step-3.component.js
```diff
@@ -173,13 +174,24 @@ class Step3Controller {
...
+  // Log error to Datadog in order to show up in Error Tracking (RUM)
+  logToDatadogRum (error) {
+    if (error?.data) {
+      if (error.data.includes('InvalidCVV2Exception')) {
+        this.datadogRum.addError(new Error('Invalid CVV'), { context: 'Checkout Submission', errorCode: error.status })
+        return
+      }
+    }
+    this.datadogRum.addError(new Error(`Error submitting purchase: ${JSON.stringify(error)}`), { context: 'Checkout Submission', errorCode: error.status })
+  }
```
dr-bizz, line 194: Could you simplify it to the below?
```suggestion
    logToDatadogRum (error) {
    let errorMessage = `Error submitting purchase: ${JSON.stringify(error)}`;
    if (error?.data) {
      if (error.data.includes('InvalidCVV2Exception')) {
        errorMessage = 'Invalid CVV';
      }
    }
    this.datadogRum.addError(new Error(errorMessage), { context: 'Checkout Submission', errorCode: error.status })
  }
```

src/app/checkout/step-3/step-3.component.js
```diff
@@ -41,6 +41,7 @@ class Step3Controller {
     this.sessionStorage = $window.sessionStorage
     this.selfReference = this
     this.isBranded = envService.read('isBrandedCheckout')
+    this.datadogRum = datadogRum
```
dr-bizz, line 44: Why did you create a local version of `datadogRum`?

dr-bizz (commented):
The code looks great. I think you could simplify the code a little, but that is it.

src/app/checkout/step-3/step-3.component.js
wrandall22: It makes it easier to test

dr-bizz (approved):
It looks great! Thank you for simplifying the `logToDatadogRum` function

wrandall22:
This was in staging at one point, I can double check it.

wrandall22:
This is still in staging, yes.

dr-bizz:
Interesting, I was having crazy merge conflicts. It removed the datadog import on the step3 checkout file. And staging no longer uses `componentInstance`

dr-bizz:
It seems it was in staging, but `componentInstance` was never defined. which is why I didn't accept the changes when merging. I'll see if I can add it back into staging

wrandall22:
In staging, this logic is all in order service, due to Caleb A's changes to branded checkout.

dr-bizz:
Okay, so maybe I'll just leave it then

PR #1149, opened 2025-02-10
[EP-2362] Retry signIn once to mitigate ephemeral fetch errors

canac:
## Description

Sometimes `this.oktaSignInWidget.showSignInAndRedirect` throws because of a `fetch` error. We are current unsure of why the request sometimes fails. In the meantime, to mitigate this, we are automatically re-rendering the widget and retrying the sign in once. We only retry once to avoid getting stuck in an infinite loop.

## Testing

* Open the sign in modal
* Set network throttling to Offline in DevTools
* Click the create account button
* Verify that the requests to set up the sign up form happen exactly twice


dr-bizz (approved):
This looks great. It's hard to test, but I believe this will work. I'll have to test it on my local when the error occurs

PR #1151, opened 2025-02-10
Design the sign up form

dr-bizz:
## Description
In this PR, I designed the signup form as the design. - https://www.figma.com/design/PA7M9Fpg5x0SEYWAXH1W8Q/Okta?node-id=42-300&t=dSfewiNOIhn1Of3Z-0

## Major changes

- **Added JS to control label:** Since the label and inputs aren't siblings, I've had to add some JS that notifies the label when the user focuses on the input or the input has a value. - We aren't using placeholders because the field name disappears when you focus on an input. Not UX-friendly.

## Minor changes

- **Sign-up steps:** I've added the three steps to the signup process, giving the user an idea of how far they are in the process.
- **CSS:** Added lots of styles to the Okta form.


canac (commented):
This is amazing! The Material UI styles are pretty cool. This is also a great use of `AbortController`.

I noticed a few UI things:

1. The Back button doesn't have a focus ring when you tab to it.

2. The error message for a field covers up the focus ring.

Focused without error:
Screenshot 2025-02-10 at 3 53 22 PM

Focused with error:
Screenshot 2025-02-10 at 3 53 16 PM

3. The country dropdown (and possibly other dropdowns) has slightly rounded borders when focused (you may have to zoom in to see):

Screenshot 2025-02-10 at 3 54 24 PM

4. Some address error messages don't line up and you can see a faint outline of the select inside its box:

Screenshot 2025-02-10 at 4 02 19 PM

Screenshot 2025-02-10 at 4 04 30 PM




5. Somehow I was able to get the back button to show up on step 1. 

6. Is the Next button text supposed to turn from black to white on hover?

Incredible work on this, it looks great!

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -15,6 +15,52 @@ <h3 translate>Create Your Account</h3>
...
+      <div class="step">
+        <img
+          alt="Step 3 - Security"
+          ng-src="{{$ctrl.imgDomain}}/assets/img/okta-sign-up-steps/three.svg"
+          ng-class="{'semi-transparent': $ctrl.currentStep < 3}"
+        />
+      </div>
+    </div>
+    
     <div id="osw-container"></div>
```
canac, line 64: What do you think about putting a loading indicator in here so that the user sees something while the form is loading?

dr-bizz:
> 6. Is the Next button text supposed to turn from black to white on hover?

I'm unsure; I see you haven't made it go white on our buttons. I'll let @rguinee speak into how to style the hovered button. - I'm good with changing mine, I just want to ensure we do it right.

src/assets/img/okta-sign-up-steps/dots.svg
```diff
@@ -0,0 +1,3 @@
...
+</svg>
```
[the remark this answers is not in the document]
dr-bizz: The dots don't look as good using your suggestion. As they are so small, the circles don't look right.
I got the dots from the Figma files.

OG Version:
Screenshot 2025-02-11 at 1 56 27 PM

New Version:
Screenshot 2025-02-11 at 1 56 41 PM

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -15,6 +15,52 @@ <h3 translate>Create Your Account</h3>
...
+          ng-if="$ctrl.currentStep === 1"
+        />
+        <img
+          alt="Step 1 - Identity"
+          ng-src="{{$ctrl.imgDomain}}/assets/img/okta-sign-up-steps/one-complete.svg"
+          ng-if="$ctrl.currentStep > 1"        />
+      </div>
+      <span class="between-steps">
+        <img
+          alt="dots between steps"
```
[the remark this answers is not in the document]
dr-bizz: I think you are correct. I'll remove them.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -525,6 +527,39 @@ class SignUpModalController {
...
+  initializeFloatingLabels () {
+    // As the Label and Input fields are not directly related in the DOM, we need to manually
+    // add the active class to the label when the input is focused or has a value.
+
+    // Remove any existing listeners before adding new ones
+    this.floatingLabelAbortControllers.forEach(controller => controller.abort())
+    this.floatingLabelAbortControllers = []
+
+    document.querySelectorAll('.o-form-content input[type="text"], .o-form-content input[type="password"]').forEach((input) => {
+      const label = document.querySelector(`.o-form-content label[for="${input.id}"]`)?.parentNode;
```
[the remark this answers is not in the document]
dr-bizz: This is black magic! I love it!

src/assets/scss/components/_signInForm.scss
```diff
@@ -16,57 +16,65 @@ sign-in-form {
 
   .details-wrapper {
     display: flex;
-    flex-wrap: wrap; // Keep the summaries on the first line and wrap the description to the next line
+    flex-wrap: wrap;
```
canac, line 19: Looks like a lot of the comments in this file got moved away from the styles that they're referring to.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -428,6 +434,10 @@ class SignUpModalController {
     }
   }
 
+  ready (context) {
+    this.$scope.$apply(() => this.isLoading = false)
```
canac, line 438:
```suggestion
    this.$scope.$apply(() => {
      this.isLoading = false
    })
```

src/assets/scss/components/_signUpModal.scss
```diff
@@ -55,7 +82,8 @@ sign-up-modal {
   section[data-se="password-authenticator--rules"] {
     display: none;
   }
-  div[data-se="o-form-fieldset-credentials.passcode"] ~ section[data-se="password-authenticator--rules"] {
+
+  div[data-se="o-form-fieldset-credentials.passcode"]~section[data-se="password-authenticator--rules"] {
```
canac, line 86:
```suggestion
  div[data-se="o-form-fieldset-credentials.passcode"] ~ section[data-se="password-authenticator--rules"] {
```

canac (commented):
Great work with everything from the last review! I had a few more questions and thoughts:

1. In the design, the Next button is called "Continue". Should we rename the button?
2. Thoughts on changing the "Create an Account" button text to "Create Account"?
3. Step 3 of the design has a place to input a 6-digit code sent to the user's email. Is that accurate in the design because I see in this modal there is a password form. I'm assuming that the design isn't correct in this case.
4. When focused, some of the inputs still have rounded corners. It makes them look funny because they have rectangular borders when they aren't focused.

Screenshot 2025-02-11 at 4 17 39 PM
Screenshot 2025-02-11 at 4 17 21 PM
Screenshot 2025-02-11 at 4 17 10 PM
Screenshot 2025-02-11 at 4 30 08 PM


5. It appears that you sorted a lot of the CSS rules alphabetically. Personally, I disagree with this change because there is often a more logical order to CSS rules than an alphabetic sort. i.e. putting `width` and `height` next to each other, putting `margin` and `padding` next to each other, putting flex-related styles next to `display: flex`, etc. If we want to apply a consistent sort order to our CSS rules, I think we should look into something like [Stylelint](https://stylelint.io/) with a config like [`stylelint-config-idiomatic-order`](https://www.npmjs.com/package/stylelint-config-idiomatic-order). I think you should revert the alphabetical sort changes, but what are your thoughts?

dr-bizz:
> When focused, some of the inputs still have rounded corners. It makes them look funny because they have rectangular borders when they aren't focused.

I see what is happening. I think it only happens on the password input as it's the first and last input, which I added a border radius to, so it fits with the form border.

canac (commented):
Thanks for those tweaks! The phone number and account type fields still have focus rings that don't quite match the borders. Besides that, this looks great!

Screenshot 2025-02-12 at 9 28 02 AM
Screenshot 2025-02-12 at 9 27 45 AM

canac:
@rguinee "The top of the modal does not have a radius set" and probably "Modal Body Spacing" should be done by me in the PR that creates the panel/header wrapper for these modals.

dr-bizz:
> Is it possible to just have "Zip Code" or "Postal Code" rather than "Zip / Postal Code"

I copied the fields and their names from the contact info form we use on the checkout and branded checkout. I highly doubt the stakeholders will approve that since we have a "Zip / Postal Code" everywhere else.
Also, I can ask, but I assume we use "Zip / Postal Code" for translations since most languages don't use the phrase "zip code".

rguinee:
I assumed that, too. Given that we might as well drop "Zip." I can float
the idea to Mike while we're reviewing other label terminology like
"Security Code vs CVV"

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -419,6 +416,18 @@ class SignUpModalController {
...
+    // Handle inactivity error
+    // The Okta widget has an issue where if the page is idle for a period of time,
+    // the Okta interaction session will expire, causing the widget to show an error "You have been logged out due to inactivity..."
+    // The user then has to click "back to sign in," to rerender the Okta widget.
+    // To avoid the error showing, we check if the context formName is 'terminal'. If it is, we know there was an error,
+    // and we re-render the widget to refresh the widget.
+    // We also set a flag to prevent the widget from refreshing multiple times.
+    if (context.formName === 'terminal' && !this.refreshedDueToInactivity) {
+      this.refreshedDueToInactivity = true
+      this.reRenderWidget()
```
canac, line 431: Do we need to set `this.refreshedDueToInactivity` back to `false` at some point?

canac (commented):
Thanks for fixing the border radiuses! They all look great now!

Sorry, I noticed one more thing. When you type a letter in the password field, its height changes by what looks like 1 pixel. I don't see this on any other fields.

https://github.com/user-attachments/assets/ca6abaca-593a-4a16-8a98-1d5bdfa4fd5d

src/assets/scss/components/_oktaSigninWidget.scss
rguinee, line 1: @canac's comment about the password input height changing. I thought it would be because of differences in line-height between Verdana and Cru's font. It seems that the o-form input's have a height set at 100%. Removing that attribute resolves the pixel displacement issue when a user keys in a character and when the user swaps input type between password and text.

rguinee (commented):
Looking great, thanks for taking a pass on input field clean up!

dr-bizz:
@canac @rguinee What do I need to do to get this merged into 2362-okta? I'm wanting to merge it in today. 

I'm currently working on the `refreshedDueToInactivity` functionality, but I should be done soon.

rguinee (approved):
Great work!

canac (approved):

PR #1153, opened 2025-02-13
[EP-2362] Okta sign up modal cleanup

canac:
## Description

Clean up some of the sign up modal styles now that the register account modal rework has merged.

* Cleaned up some duplicated styles.
* Remove left and right padding from Okta container.
* Added line under Create Account header.
* Made modal a little narrower.
* Updated the Okta gray to be lighter, like the design and like some other places in this modal.

### DIFF
--- a/src/assets/scss/components/_signUpModal.scss
+++ b/src/assets/scss/components/_signUpModal.scss
@@ -1,20 +1,19 @@
 sign-up-modal {
   position: relative;
 
-  .sign-up-header {
+  .signup-header {
     text-align: center;
+    padding-bottom: 1rem;
   }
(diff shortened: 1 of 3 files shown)


src/assets/scss/components/_signUpModal.scss
```diff
@@ -1,20 +1,19 @@
 sign-up-modal {
   position: relative;
 
-  .sign-up-header {
+  .signup-header {
```
dr-bizz, line 4: I see you didn't edit thr HTML, was this edit on purpose?

dr-bizz (approved):

PR #1154, opened 2025-02-15
Adding styleLint to Give

dr-bizz:
## Descriotion
This pull request includes several updates to the project's SCSS files, as well as changes to the project's configuration files. The primary focus is on improving the styling and ensuring consistency across the codebase. The most important changes include updating the SCSS files to fix formatting issues, adding new stylelint configurations, and updating dependencies in the `package.json` file.

I didn't update all the SCSS files/folders, as it would be tii much for one PR, and I want to get this right first.

### SCSS Formatting and Improvements:
* Fixed various formatting issues in multiple SCSS files, such as `_account_management.scss`, `_carousel.scss`, `_cart.scss`, `_checkout.scss`, `_form.scss`, and `_gift-config.scss`.

### Stylelint Configuration:
* Added new stylelint configurations by extending `stylelint-config-standard`, `stylelint-config-standard-scss`, and `stylelint-config-idiomatic-order` in the `.stylelintrc` file.
* Updated `package.json` to include new stylelint dependencies and added a new script for linting styles. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L59-R75) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L87-R93)

### Dependency Updates:
* Updated several dependencies in the `package.json` file, including `css-loader`, `mini-css-extract-plugin`, and added new stylelint-related packages.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,3 @@
+{ 
+  "extends": ["stylelint-config-standard", "stylelint-config-standard-scss", "stylelint-config-idiomatic-order"]
+}
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "jest": "^29.7.0",
     "jest-date-mock": "^1.0.7",
     "jest-environment-jsdom": "^29.4.2",
-    "mini-css-extract-plugin": "^0.8.0",
+    "mini-css-extract-plugin": "1.6.2",
     "moment-locales-webpack-plugin": "^1.1.0",
     "ngtemplate-loader": "^2.0.1",
     "sass": "^1.49.0",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "babel-loader": "^8.0.6",
     "babel-plugin-angularjs-annotate": "^0.10.0",
     "copy-webpack-plugin": "^5.0.4",
-    "css-loader": "^3.2.0",
+    "css-loader": "5.2.7",
     "html-loader": "^0.5.5",
     "jest": "^29.7.0",
     "jest-date-mock": "^1.0.7",
@@ -68,6 +68,11 @@
     "sass-loader": "^10",
     "standard": "^16.0.4",
     "style-loader": "^1.0.0",
+    "stylelint": "^16.14.1",
+    "stylelint-config-idiomatic-order": "^10.0.0",
+    "stylelint-config-standard": "^37.0.0",
+    "stylelint-config-standard-scss": "^14.0.0",
+    "stylelint-webpack-plugin": "^5.0.1",
     "typescript-eslint": "^0.0.1-alpha.0",
     "webpack": "^4.47.0",
     "webpack-bundle-analyzer": "^3.4.1",
@@ -84,7 +89,8 @@
     "test:log": "jest --silent=false",
     "lint": "standard",
     "lint:write": "standard --fix",
-    "lint:ts": "tsc"
+    "lint:ts": "tsc",
+    "lint:styles": "stylelint \"src/assets/scss/*.{css,scss}\""
   },
   "standard": {
     "env": {
--- a/src/assets/scss/_form.scss
+++ b/src/assets/scss/_form.scss
@@ -1,36 +1,40 @@
 
 .form-group.has-error {
   .form-control {
-    box-shadow: inset 0px 0px 0px 1px $state-danger-text;
+    box-shadow: inset 0 0 0 1px $state-danger-text;
   }
+
   label > span, label > translate {
     color: $state-danger-text;
     font-weight: 600;
-    &:before {
-      content: "\f071";
+
+    &::before {
       display: inline-block;
-      font-family: "fontawesome", "Font Awesome 5 Pro";
       margin-right: 6px;
+      content: "\f071";
+      font-family: fontawesome, "Font Awesome 5 Pro";
     }
   }
 }
 
 .form-group.has-success {
   .form-control {
-    box-shadow: inset 0px 0px 0px 1px $colorHelper-success;
+    box-shadow: inset 0 0 0 1px $colorHelper-success;
   }
+
   label > span, label > translate {
     color: $colorHelper-success;
-    &:before {
-      content: "\f00c";
+
+    &::before {
       display: inline-block;
-      font-family: "fontawesome", "Font Awesome 5 Pro";
       margin-right: 6px;
+      content: "\f00c";
+      font-family: fontawesome, "Font Awesome 5 Pro";
     }
   }
 }
 
-label:not(:has(input[type=checkbox])) {
+label:not(:has(input[type="checkbox"])) {
   & > span, & > translate {
     display: block;
     margin-bottom: 4px;
@@ -38,21 +42,22 @@ label:not(:has(input[type=checkbox])) {
 }
 
 .is-required {
-  label > span:after, label > translate:after {
-    content: "*";
+  label > span::after, label > translate::after {
     display: inline-block;
+    margin-left: 6px;
+    content: "*";
+    font-family: fontawesome, "Font Awesome 5 Pro";
     font-size: 115%;
     line-height: 0;
-    font-family: "fontawesome", "Font Awesome 5 Pro";
-    margin-left: 6px;
   }
-  input[type=text],
-  input[type=number],
-  input[type=password],
-  input[type=email],
-  input[type=search],
-  input[type=url],
-  input[type=tel],
+
+  input[type="text"],
+  input[type="number"],
+  input[type="password"],
+  input[type="email"],
+  input[type="search"],
+  input[type="url"],
+  input[type="tel"],
   select {
   }
 }
@@ -72,23 +77,26 @@ label:not(:has(input[type=checkbox])) {
 
 .phone-row {
   position: relative;
+
   select.form-control-subtle {
     display: inline-block !important;
     width: 23%;
     margin-right: -4px;
   }
-  input[type=tel] {
+
+  input[type="tel"] {
     display: inline-block !important;
     width: 50% !important;
     margin: 0 1% 0 2% !important;
     vertical-align: bottom;
   }
+
   a {
     position: absolute;
     top: 50%;
     right: -20px;
-    transform: translateY(-50%);
     color: $colorHelper-wanger;
+    transform: translateY(-50%);
     .fa {font-size: 18px;}
   }
 }
@@ -103,45 +111,55 @@ label:not(:has(input[type=checkbox])) {
   select, input {margin-left: 3px;}
   select, input, .fa {float: left;}
   .fa {line-height: 39px;}
+
   select.form-control-subtle {
     display: inline-block !important;
     width: 12%;
-    padding-left: 0;
     padding-right: 0;
+    padding-left: 0;
   }
-  input[type=name] {
+
+  input[type="name"] {
     display: inline-block !important;
     width: 50% !important;
     margin: 0 1% 0 2% !important;
     vertical-align: bottom;
   }
+
   select.prefix {
     @include respond-to(extrasmall) {
-      padding-left: 0;
       padding-right: 0;
+      padding-left: 0;
       margin-bottom: 5px;
     }
   }
+
   input.first {
     width: 30%;
+
     @include respond-to(extrasmall) {
       width: 73%;
       margin-bottom: 5px;
     }
   }
+
   input.last {
     width: 30%;
+
     @include respond-to(extrasmall) {
       width: 60%;
     }
   }
+
   input.middle {
     width: 8%;
+
     @include respond-to(extrasmall) {
       width: 12%;
       margin-left: 14px;
     }
   }
+
   input.first,
   input.last,
   input.middle {
@@ -156,12 +174,13 @@ label:not(:has(input[type=checkbox])) {
   line-height: 1.25rem;
 }
 
-.btn:focus, .btn:active:focus { outline: none;  -webkit-appearance: none; }
+.btn:focus, .btn:active:focus {  appearance: none; outline: none; }
 
 form.new-gift {
   .input-group-addon {padding: 0;}
+
   .btn-link {
-    border-width: 0;
     padding: 8px 22px;
+    border-width: 0;
   }
 }
--- a/src/assets/scss/_gift-config.scss
+++ b/src/assets/scss/_gift-config.scss
@@ -1,18 +1,16 @@
 .btn-radio, .custom-amount{
-  padding-top: 1%;
-  padding-bottom: 1%;
-  padding-left: 15px;
-  padding-right: 15px;
   min-width: 85px;
+  padding: 1% 15px 1% 15px;
   border: 1px solid $colorGrayscale-stone;
+
   &.btn-wide {
     min-width: 150px;
   }
 
-  input[type=radio]:checked + label{
+  input[type="radio"]:checked + label{
+    width: 100%;
     background: #67747c;
     color: #fff;
-    width: 100%;
   }
 
   input[type="radio"]{
@@ -31,10 +29,10 @@ label.custom-amount {
 .checkbox {
   label {
     min-height: 29px;
-    padding-left: 0px;
+    padding-left: 0;
     margin-bottom: 0;
-    font-weight: normal;
     cursor: pointer;
+    font-weight: normal;
   }
 }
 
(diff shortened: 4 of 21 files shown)


dr-bizz:
@wrandall22 I've cleaned up the styles and fixed the webpack issue I created.

wrandall22 (commented):
A few comments. I want to get John and Ryan's opinion about the rules themselves. I'm a bit curious about the loss of browser compatibility with removing all those webkit-type rules (maybe this is handled for us or no longer necessary, not sure).

src/assets/scss/_form.scss
```diff
@@ -156,12 +159,18 @@ label:not(:has(input[type=checkbox])) {
   line-height: 1.25rem;
 }
 
-.btn:focus, .btn:active:focus { outline: none;  -webkit-appearance: none; }
+.btn:focus, .btn:active:focus {
+  appearance: none;
```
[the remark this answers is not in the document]
dr-bizz: Chrome has supported the use of `appearance ` without the prefix `webkit` since July 2020, and similar support has been available for other browsers.

The global percentage of users' browsers that don't support it is 1.4%. 

Below are the stats broken down by browser:

| Browser | Global percentage that doesn't support it |
| --------| ----------------- |
| Chrome | `0.34%`|
| Edge | `0.12%`|
| Safari | `0.15%`|
| FireFox | `0.09%`|
| Opera | `0.02%`|
| Chrome (Android) | `0%`|
| Safari (IOS) | `0.68%`|

src/assets/scss/_gift-config.scss
```diff
@@ -31,54 +29,63 @@ label.custom-amount {
...
 label.btn-radio {
   .btn span {
     font-size: 18px ;
   }
+
   input[type="radio"] ~ i.fa.fa-circle-o {
-    font-size: 1em;  color: #f0f0f0;    display: inline;
-    -webkit-transition-property: all; /* Safari */
-    -webkit-transition-duration: 200ms; /* Safari */
-    -webkit-transition-delay: 200ms; /* Safari */
```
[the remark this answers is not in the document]
dr-bizz: We are not losing any functionality. I've removed the `-webkit-` prefix, as this is now standard. The properties are still being applied, but without the prefix.

The global percentage of Safari user's that don't support `transition-delay` without the prefix is `0.02%`

```
transition-delay: 200ms;
transition-duration: 200ms;
transition-property: all;
```

src/assets/scss/_form.scss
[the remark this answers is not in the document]
j2trumpet: I always check caniuse.com for this kind of stuff. https://caniuse.com/?search=appearance

src/assets/scss/_tabs.scss
```diff
@@ -1,291 +1,330 @@
...
         transition: all 0.3s ease !important;
       }
+
       &.slide-left.sliding, &.slide-right.sliding {
         opacity: 0 !important;
       }
+
       &.slide-left.active, &.slide-right.active {
-        -webkit-transform: translate3d(0, 0, 0);
-        transform: translate3d(0, 0, 0);
```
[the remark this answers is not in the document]
dr-bizz: So the third digit in `translate3d` is the `z` axis. since it is `0`, we can just use the `translate` as it's the same.

src/assets/scss/_search.scss
```diff
@@ -1,69 +1,78 @@
...
-      border-radius: $border-radius;
       height: 50px;
-      line-height: 48px;
       padding: 0 50px;
-      font-size: 20px;
+      border-radius: $border-radius;
       margin-top: 0;
+
+      /* Discover more button BG: */
+      background: #F9B625;
```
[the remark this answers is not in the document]
dr-bizz: I'll look into this.

src/assets/scss/_homepage.scss
```diff
@@ -140,53 +142,59 @@ $border-radius: 4px;
...
       .hero-lead-second, 
       .hero-lead-third {
         width: 100%;
         border-bottom: 1px solid #333;
+
         .image-container {
           padding-top: 9px;
           padding-bottom: 9px;
-          background-color: rgba(0,0,0,.8);
+          background-color: rgba(0,0,0,80%);
```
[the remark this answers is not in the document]
dr-bizz: I'll take a look

src/assets/scss/_give.scss
```diff
@@ -157,92 +158,91 @@ hr.horizontal-divider {
...
 
   @include respond-to(supertiny) {
-    &.vdivide [class*='col-']:not(:last-child):after {
+    &.vdivide [class*='col-']:not(:last-child)::after {
       top: 30px;
     }
   }
 
   &.row-eq-height {
-    display: -webkit-box;
```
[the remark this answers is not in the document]
dr-bizz: The global percentage of users' major browsers that don't support it is 0.03%. 

Below are the stats broken down by major browser:

| Browser | Global percentage that doesn't support it |
| --------| ----------------- |
| Chrome | `0%`|
| Edge | `0%`|
| Safari | `0%`|
| FireFox | `0.02%`|
| Opera | `0%`|
| Chrome (Android) | `0%`|
| Safari (IOS) | `0.01%`|

src/assets/scss/_gift-config.scss
```diff
@@ -87,22 +94,14 @@ label.btn-radio {
...
+
 div[data-toggle="buttons"] label {
   margin-bottom: 10px;
   font-weight: 600;
-
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  -o-user-select: none;
-  user-select: none;
```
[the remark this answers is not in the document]
dr-bizz: It's good to be curious as it ensures we are only deploying our best work.

After investigating it, I found that Safari only supports it with the `-webkit-` prefix. So, I'll look into this Lint rule to see if we can make an exception.

src/assets/scss/_form.scss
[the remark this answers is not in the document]
dr-bizz: > I always check caniuse.com for this kind of stuff. https://caniuse.com/?search=appearance

Yes, so do I!

src/assets/scss/_search.scss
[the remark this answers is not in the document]
j2trumpet: That would be nice!

src/assets/scss/_gift-config.scss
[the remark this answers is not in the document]
dr-bizz: I found a list of properties that still use the prefixes, so I'll be adding this rule.

```
 "property-no-vendor-prefix": {
          "ignoreProperties": ["backdrop-filter", "user-select", "initial-letter", "text-decoration", "text-stroke", "text-fill-color", "line-clamp", "box-decoration-break", "stretch", "available"]
        }
```

src/assets/scss/_give.scss
```diff
@@ -537,45 +561,51 @@ hr.horizontal-divider {
...
+			margin-left: $gutter * .25;
 		}
 	}
 }
 
 @media screen and (min-width: 550px ) and (max-width: 991px ) {
   .split-row {
     width: 100%;
-    margin-top: $gutter*0.5;
+    margin-top: $gutter * 0.5;
```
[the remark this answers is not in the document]
dr-bizz: It looks like version 14 had this rule, it was called `number-leading-zero` but they removed it on version 15.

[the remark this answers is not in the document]
dr-bizz: Is this something you think we need to create a custom rule for?

src/assets/scss/_homepage.scss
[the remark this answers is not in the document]
dr-bizz: we can use the `alpha-value-notation` rule

.stylelintrc
```diff
@@ -0,0 +1,51 @@
...
+        "selector-class-pattern": null,
+        "scss/dollar-variable-pattern": null,
+        "selector-id-pattern": null,
+        "scss/no-global-function-names": null,
+        "selector-type-no-unknown": null,
+        "property-no-vendor-prefix": [true,
+          {
+            "ignoreProperties": ["backdrop-filter", "user-select", "initial-letter", "text-decoration", "text-stroke", "text-fill-color", "line-clamp", "box-decoration-break", "stretch"]
+          }
+        ],
```
j2trumpet, line 46: Is this config removing all vendor prefixes except for those on the properties listed on line 44?

dr-bizz: Yes it is

dr-bizz: These properties still need prefixes to be supported on all major browsers.

src/assets/scss/_give.scss
```diff
@@ -537,45 +561,51 @@ hr.horizontal-divider {
...
-  margin-bottom: $gutter*.25;
   display: block;
+  margin-bottom: $gutter * .25;
+  color: lighten($colorGrayscale-text, 40%);
   line-height: 1.15em;
 }
 
 .repeating-row {
 	clear: both;
 	line-height: 1em;
```
j2trumpet, line 572: Is there a configuration we could add to make indentation uniform? This jumps from two spaces to eight.

src/assets/scss/_tabs.scss
```diff
@@ -298,181 +337,161 @@
...
-        border: 1px solid #10cfbd;
         content: "";
-        height: calc(100% + 1px);
-        left: 0;
-        position: absolute;
-        top: 0;
-        -webkit-transform: translate3d(0, 100%, 0px);
-        transform: translate3d(0, 100%, 0px);
-        -webkit-transition: -webkit-transform 0.3s ease 0s;
+        transform: translate3d(0, 100%, 0);
```
j2trumpet, line 362: Looks like another instance where `translate3d` could be swapped for just `translate`.

src/assets/scss/_tabs.scss
```diff
@@ -298,181 +337,161 @@
...
-      transition: transform 0.5s ease 0s;
-      display: block;
-      font-weight: 700;
-      line-height: 1.5;
+      span {
+        display: block;
+        backface-visibility: hidden;
+        font-weight: 700;
+        line-height: 1.5;
+        transform: translate3d(0, 5px, 0);
```
j2trumpet, line 371: Looks like another instance where `translate3d` could be swapped for just `translate`.

src/assets/scss/_give.scss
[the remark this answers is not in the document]
dr-bizz: I found a way to do it.

src/assets/scss/_search.scss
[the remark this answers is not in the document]
dr-bizz: done.

src/assets/scss/_gift-config.scss
```diff
@@ -65,11 +65,13 @@ label.btn-radio {
   }
 
   &:hover input[type="radio"] ~ i.fa {
+    /* stylelint-disable-next-line */
```
wrandall22, line 68: Are these here because there is no variable?

src/assets/scss/_tabs.scss
```diff
@@ -146,36 +150,42 @@
...
     > li::after {
+      /* stylelint-disable-next-line */
       background-color: #f55753;
     }
   }
 
   &.nav-tabs-warning {
-    > li::after {
+    > li::after 
+    /* stylelint-disable-next-line */{
```
j2trumpet, line 181: Is this behaving properly? Comparing this line to #173/#174, the curly bracket's in a different place.

src/assets/scss/_give.scss
dr-bizz: Done

src/assets/scss/_gift-config.scss
dr-bizz: Yes. The rule sees if there are any HEX values and creates a warning. So if there isn't a variable, we have to use this ignore to prevent the warning.

I would prefer the rule to check the variables and then the styles, so we so we don't have to add this ignore. But it is still good as it pushes devs to use variables as much as they can.

src/assets/scss/_tabs.scss
dr-bizz: That's my bad.

src/assets/scss/_gift-config.scss
j2trumpet: I concur. Writing a new HEX should be the exception, not the rule, so there shouldn't be many of these ignores in a good, color-consistent design.

.stylelintrc
```diff
@@ -0,0 +1,70 @@
...
+            "place-content",
+            "place-items",
+            "place-self",
+            "scroll-margin",
+            "border-block",
+            "border-block-end",
+            "border-block-start",
+            "border-inline",
+            "border-inline-end",
+            "border-inline-start"
```
canac, line 36: None of these are necessary

.stylelintrc
```diff
@@ -0,0 +1,70 @@
...
+        "no-invalid-position-at-import-rule": null,
+        "no-descending-specificity": null,
+        "selector-class-pattern": null,
+        "scss/dollar-variable-pattern": null,
+        "selector-id-pattern": null,
+        "scss/no-global-function-names": null,
+        "selector-type-no-unknown": null,
+        "property-no-vendor-prefix": [true,
+          {
+            "ignoreProperties": ["backdrop-filter", "user-select", "initial-letter", "text-decoration", "text-stroke", "text-fill-color", "line-clamp", "box-decoration-break", "stretch"]
```
canac, line 48: Only `user-select` is necessary now

```suggestion
            "ignoreProperties": ["user-select"]
```

.stylelintrc
```diff
@@ -0,0 +1,70 @@
...
+          }
+        ],
+        "alpha-value-notation": "number",
+        "scale-unlimited/declaration-strict-value": [
```
canac, line 52: Does this rule add much value? It seems like it overlaps with the one that disallows raw hex colors and makes you use a color value instead. Are there cases that this rule catches that the raw hex rule doesn't?

canac (commented):
I few more thoughts on how to trim down our overrides.

.stylelintrc
dr-bizz: I would like to keep them in. Even if they are not necessary now, I want to set up the style lint so we don't have to remember to add these back if we use one of these properties.

.stylelintrc
dr-bizz: They aren't necessary now, but I would like to keep them in. I want to set up the style lint so we don't have to remember to add these back if we use one of these properties.

.stylelintrc
```diff
@@ -0,0 +1,70 @@
+{
```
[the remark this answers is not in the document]
dr-bizz: I'm good with it either way. Happy to rename it

.stylelintrc
dr-bizz: The `color-no-hex` just disallows HEX values. So, a value like `red` will pass even though there could be a red variable.
`scale-unlimited/declaration-strict-value` can be configured to catch RGB colours as well.

canac: Ah, yes. Good point.

.stylelintrc
canac: Gotcha. Were did this list come from?

.stylelintrc
canac: Gotcha. Were did this list come from?

canac:
@dr-bizz I'm sure you were going to do this, but can we add the stylelint check to CI so it runs automatically in our PRs?

dr-bizz:
Yes, I had planned on adding it as a CI test

.stylelintrc
dr-bizz: I've got the list from Fully Stacked at https://fullystacked.net/prefix/

.stylelintrc
dr-bizz: I've got the list from Fully Stacked at https://fullystacked.net/prefix/

dr-bizz: I got this list by going through the list of properties that this rule will fix ([which can be seen here](https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties/#ignoreshorthands-regex-regex-string)), and if the property isn't supported by all the major browsers, I prevent it from using it.

wrandall22:
It looks like the merge commit has made the git history a bit messy. We should get this rebased from `master` and also squash things like the webpack lint commit.

dr-bizz:
I can rebase the git commits. I have been holding off as I don't want to remove data on the files and lines you reviewed.

If everyone is happy with me rebasing, I can do it. If not, I'll wait to do it until I'm ready to deploy.

.stylelintrc
canac: That's a great reference, thanks! I know I suggested making this a JSON file, but it could be even better to make it `stylelint.config.js` so that we can add comments and mention where this list came from. Or at least note that those are the vendor prefixes that are still needed as of November 2024 when that article was last updated. I think that could be helpful as this list will presumably shrink over time.

.stylelintrc
canac: Got it. I assume all the block and inline related styles are because we haven't yet been given permission to drop support for IE11? This is another place that could warrant a code comment if you switch the config file to a JS file.

dr-bizz: No, it was mostly due to Chrome just starting to support it in 2020.

I just did the calculation, and it is only 1.57% of global users use browsers that don't support these properties. so we could allow some of these stylles.

canac:
> If everyone is happy with me rebasing, I can do it.

I have no objections to a rebase. It feels like we're pretty close to consensus on the rules chosen.

package.json
```diff
@@ -84,7 +91,8 @@
     "test:log": "jest --silent=false",
     "lint": "standard",
     "lint:write": "standard --fix",
-    "lint:ts": "tsc"
+    "lint:ts": "tsc",
+    "lint:styles": "stylelint \"src/{assets/scss,app}/**/*.{css,scss}\" --quiet-deprecation-warnings"
```
canac, line 95: Is there a reason that we are ignoring `src/assets/crubrand/`, `src/assets/cru-scss/`, and `src/common/components/loading/loading.scss`? If that's intentional, can you run the lint on everything and put the paths to ignore in a `.stylelintignore`? That way, devs can install the Stylelint VS Code extension and have it auto-fix problems on save, while ignoring any SCSS files that we don't want to apply our style rules to.

```suggestion
    "lint:styles": "stylelint 'src/**/*.scss' --quiet-deprecation-warnings"
```

canac (approved):
I have a suggestion to improve the dev experience when editing in VS Code, but besides that this looks great! Thank you for the detailed comments documenting each of the rules and why we are customizing them!

package.json
dr-bizz: It was intentional to leave out `src/assets/crubrand/`, `src/assets/cru-scss/` as it took a lot to do `src/assets/scss`.
When running it on `'src/**/*.scss'`, it updates 161 files and still returns over 3000 errors that I would have to manually edit.

I think that is best for another PR.

canac: That's totally fine. I'm just asking that we put ignored paths in a `.stylelintignore` file so that the Stylelint VS Code extension doesn't try to auto-fix them on save.

dr-bizz: I've added an ignore property to `stylelint.config.mjs` so I can run styleLint on `src/**/*.{css,scss}` and capture all SCSS/CSS files which includes `src/common/components/loading/loading.scss`.

canac: Is there a reason you don't want to add a `.stylelintignore` file and put the ignore paths there instead? Try editing the `crubrand` SCSS file with the Stylelint VS Code plugin installed and the configuration from #1162 and you can see that it will auto-format everything, which we don't want. Maybe I'm wrong, but it seems like devs would want to use that extension to auto-format the SCSS files that we have formatting turned on for.

dr-bizz: So when I use the `.vscode/settings`, it ignores the folders I've put in the ignore property in `stylelint.config.mjs`.

dr-bizz: Other files do get auto-formated, but not `src/assets/crubrand/` and `src/assets/cru-scss/`

canac: You're right. I apologize, I must have missed something during testing or neglected to restart Stylelint after changing the config file.

PR #1155, opened 2025-02-17
EP-2566 Add Card Bin

caleballdrin:
Include the first 6 digits of the credit card in the Google Recaptcha assessment. This is PCI-compliant and provides Google with more data to improve risk assessments.
https://jira.cru.org/browse/EP-2566

- When saving a Credit Card, pass the first 6 digits to the `addCreditCardPayment` method of `order.service`
- Add functions to store, retrieve and clear the card bin
- Include the card bin when submitting the gift

### DIFF
--- a/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
+++ b/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
@@ -158,6 +158,7 @@ class CreditCardController {
                 'expiry-month': this.creditCardPayment.expiryMonth,
                 'expiry-year': this.creditCardPayment.expiryYear,
                 'last-four-digits': tokenObj.maskedCardNumber,
+                'card-bin': this.creditCardPayment.cardNumber?.replace(/\s+/g, '').substring(0, 6) || null,
                 transactionId: tokenObj.transactionID,
                 cvv: this.creditCardPayment.securityCode
               }
--- a/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.spec.js
+++ b/src/common/components/paymentMethods/creditCardForm/creditCardForm.component.spec.js
@@ -154,6 +154,7 @@ describe('credit card form', () => {
           'expiry-month': 12,
           'expiry-year': 2019,
           'last-four-digits': '1111',
+          'card-bin': '411111',
           transactionId: '<transaction id>',
           cvv: '123'
         }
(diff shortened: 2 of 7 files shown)


caleballdrin:
@wrandall22 I left some console logs in this PR to make it easier for you to test.

wrandall22 (changes requested):
Please fix the tests and lint. Also, I'm not 100% sure if the bin will be captured by the card creation process or just at gift submission; we won't know that until the back-end team does their work.

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.spec.js
```diff
@@ -129,7 +129,7 @@ describe('credit card form', () => {
         cardholderName: 'Person Name',
         expiryMonth: 12,
         expiryYear: 2019,
-        securityCode: '123'
+        securityCode: '123',
```
wrandall22, line 132: Remove the trailing comma.

src/common/components/paymentMethods/creditCardForm/creditCardForm.component.js
```diff
@@ -158,6 +158,7 @@ class CreditCardController {
                 'expiry-month': this.creditCardPayment.expiryMonth,
                 'expiry-year': this.creditCardPayment.expiryYear,
                 'last-four-digits': tokenObj.maskedCardNumber,
+                'card-bin': this.creditCardPayment.cardNumber?.replace(/\s+/g, '').substring(0, 6) || null,
```
wrandall22, line 161: Could you write a test for this logic?

wrandall22 (changes requested):
I'll try to do some QA on this tomorrow, for now a couple minor things.

wrandall22:
I've done QA, you can remove the console logs now.

wrandall22:
Could you squash the last 3 commits into the others? That should be the last thing.

wrandall22:
The 2nd commit no longer does what it says it is doing, but it fits under the 3rd commit's message logically. You can combine the two commits and keep the 3rd commit's wording.

caleballdrin:
@wrandall22 I'm pretty sure the 2nd commit, "Remove the current cardBin when selecting a different exisiting payment" has the same changes it had before. It is setting the `cardBin` to `null` in the `selectPayment` function.

wrandall22:
You are right. I was thrown off by the use of `storeCardBin` instead of `clearCardBins`

wrandall22 (approved):

PR #1156, opened 2025-02-19
Bump the amount of sessions we are capturing in RUM

wrandall22:
We want to do this in preparation for testing the new v3 of branded checkout.

### DIFF
--- a/src/common/datadog.config.js
+++ b/src/common/datadog.config.js
@@ -12,7 +12,7 @@ const dataDogConfig = /* @ngInject */ function (envServiceProvider) {
       env: envServiceProvider.get(),
       allowedTracingUrls: [envServiceProvider.read('apiUrl') + '/cortex'],
       version: process.env.GITHUB_SHA,
-      sessionSampleRate: envServiceProvider.is('staging') ? 100 : 10,
+      sessionSampleRate: envServiceProvider.is('staging') ? 100 : 50,
       sessionReplaySampleRate: envServiceProvider.is('staging') ? 100 : 1,
       trackUserInteractions: true,
       trackResources: true,


src/common/datadog.config.js
```diff
@@ -12,7 +12,7 @@ const dataDogConfig = /* @ngInject */ function (envServiceProvider) {
       env: envServiceProvider.get(),
       allowedTracingUrls: [envServiceProvider.read('apiUrl') + '/cortex'],
       version: process.env.GITHUB_SHA,
-      sessionSampleRate: envServiceProvider.is('staging') ? 100 : 10,
+      sessionSampleRate: envServiceProvider.is('staging') ? 100 : 50,
       sessionReplaySampleRate: envServiceProvider.is('staging') ? 100 : 1,
       trackUserInteractions: true,
```
rguinee, line 17: Do we need to use the frustrations property `trackFrustrations: true,` ? Wasn't sure which version of DataDog we're on.

rguinee (commented):
Do we have JavaScript error tracking turned off for a reason? Isn't that how validation errors are handled?

src/common/datadog.config.js
wrandall22: We're on version [5.16.0](https://github.com/CruGlobal/give-web/blob/master/package.json#L8), so we [don't need](https://docs.datadoghq.com/real_user_monitoring/guide/browser-sdk-upgrade/#frustration-signals-are-collected-automatically) this property

wrandall22:
> Do we have JavaScript error tracking turned off for a reason? Isn't that how validation errors are handled?

I don't believe we have it turned off. Though, we haven't uploaded the source maps.

rguinee (approved):
Forgot to come back to this to submit the review.

PR #1158, opened 2025-02-20
[no-Jira] Stop recommending npm commands

canac:
Developers definitely should not be installing or adding dependencies using `npm` because the yarn lockfile will not be used. They will end up with slightly different versions of packages than expected. While `npm run test` etc. technically work, there is no reason to recommend `npm` commands in a `yarn`-based project.

I also replaced the longer `yarn run test` with the shorter, equivalent `yarn test` and fixed the `--dev` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,8 +171,8 @@ Use yarn for faster installs and to update the yarn lock file: https://yarnpkg.c
 
 ### Install & Run
 
-1. `yarn` or `npm install`
-2. `yarn start` or `npm start`
+1. `yarn`
+2. `yarn start`
 3. Browse to [`http://localhost:9000`](http://localhost:9000)
 Note: For session cookies to work correctly, add the below entries to your hosts file, and use [`https://localhost.cru.org:9000`](https://localhost.cru.org:9000) for development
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -182,15 +182,15 @@ Note: For session cookies to work correctly, add the below entries to your hosts
 
 ### Development Tasks
 
-- `yarn run test` or `npm run test` to run karma tests
-- `yarn run lint` or `npm run lint` to run eslint
-- `yarn run build` or `npm run build` to generate minified output files. These files are output to `/dist`. `common.js` must be included before any of the other JS files.
-- `yarn run build:analyze` or `npm run build:analyze` to open a visualization of bundle sizes after building
+- `yarn test` to run karma tests
```
wrandall22, line 185: Are these still Karma, or are they Jest?

wrandall22 (approved):

README.md
canac: Good point, it's Jest now. ESLint is still kind of configured, but `yarn lint` runs standard, so I should update that too.

PR #1159, opened 2025-02-21
[EP-2362] Improve styles and back button

canac:
## Description

* The first commit changes the selector to look for fields by their name and only applies the border radius changes to the account type field
* The second commit fixes the height of the zip code focus ring
* The third commit fixes the state dropdown having a bottom border radius instead of being a rectangle
* The fourth commit makes the enter advance to the next step instead of the previous step
  * `<button>` elements are unintuitively `type="submit"` by default, so the back button needed `type="button"`

### DIFF
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -176,7 +176,7 @@ sign-up-modal {
     }
 
     // Allow the account type dropdown to have rounded corners if the organisation field is hidden
-    &:has(+ .o-form-fieldset[style="display: none;"]) .chzn-container a {
+    &:has([data-se="o-form-input-accountType"]):has(+ .o-form-fieldset[style="display: none;"] [data-se="o-form-input-organizationName"]) .chzn-container a {
       border-bottom-left-radius: 6px;
       border-bottom-right-radius: 6px;
     }
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -125,7 +125,7 @@ sign-up-modal {
     }
 
     .chzn-container-single .chzn-drop {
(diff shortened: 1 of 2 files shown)


src/assets/scss/components/_oktaSigninWidget.scss
```diff
@@ -175,8 +175,8 @@ sign-up-modal {
       }
     }
 
-    // Allow the account type dropdown to have rounded corners if the organisation field is hidden
-    &:has(+ .o-form-fieldset[style="display: none;"]) .chzn-container a {
+    // Allow the account type dropdown to have rounded corners if the organization field is hidden
```
dr-bizz, line 178: 😂 `organisation` is spelt correctly.

src/assets/scss/components/_oktaSigninWidget.scss
```diff
@@ -175,10 +175,12 @@ sign-up-modal {
       }
     }
 
-    // Allow the account type dropdown to have rounded corners if the organisation field is hidden
-    &:has(+ .o-form-fieldset[style="display: none;"]) .chzn-container a {
-      border-bottom-left-radius: 6px;
-      border-bottom-right-radius: 6px;
+    // Allow the account type dropdown to have rounded corners if the organization field is hidden
```
dr-bizz, line 178:
```suggestion
    // Allow the account type dropdown to have rounded corners if the organisation field is hidden
```

dr-bizz (changes requested):

dr-bizz:
Great work! Thank you for making these changes!

src/assets/scss/components/_oktaSigninWidget.scss
canac: I was just switching to the American spelling for consistency in the rest of the repo and to match the input name `[data-se="o-form-input-organizationName"]`, but I can revert it.

dr-bizz (approved):

PR #1161, opened 2025-03-04
Fixing the sign up flow

dr-bizz:
## Description 
In this PR, I update the Sign up form to allow the user to fully sign up to Okta, get redirected to Okta for sign in and update the user's details on the Give DB.

I've made some pretty big changes in this PR, but there is still quite a bit of work to be done around cleaning up the contact info logic and ensuring it shows when expected.
I haven't done it in this PR as I need to focus on updating the API next so we can create an endpoint that connects the Okta account to the Give account.

## Changes
**I first removed where we switched to the contact info modal.**
This should only take the user to the contact info modal if there is an error saving the user's data on the Give DB.
It must happen once the user returns from Okta after a successful sign-in. _I will need to configure this in a later PR._

**Enter the verification email code without an extra click**
When you get to the verification step, it says to click on the email, which if you do, it shows a code. 
I've added some code which clicks the button "Enter verify code instead", so it renders the verify code input field by default to avoid an unnecessary click from the user.

**Sign user into Okta by redirect**
The Okta sign-in widget allows users to sign in directly without redirecting them to the Okta website. While this enhances the user experience, we need to ensure users have the opportunity to save their credentials within the Okta sign-in form. This will allow their browser to remember their email and password for future visits.

Since we use the sign-in widget, the browser doesn't have a mechanism for saving users' credentials directly within the Okta context.

### DIFF
--- a/src/common/components/signUpModal/signUpModal.component.js
+++ b/src/common/components/signUpModal/signUpModal.component.js
@@ -391,24 +391,6 @@ class SignUpModalController {
   }
 
   postSubmit (response, onSuccess) {
-    const donorDetails = {
-      name: {
-        'given-name': this.$scope.firstName,
-        'family-name': this.$scope.lastName
-      },
-      'donor-type': this.$scope.accountType,
-      'organization-name': this.$scope.organizationName,
-      email: this.$scope.email,
-      phone: this.$scope.primaryPhone,
-      mailingAddress: {
-        streetAddress: this.$scope.streetAddress,
-        locality: this.$scope.city,
-        region: this.$scope.state,
-        postalCode: this.$scope.zipCode,
-        country: this.$scope.countryCode
-      }
-    }
-    this.$scope.$apply(() => this.onSignUp({ donorDetails }))
     onSuccess(response)
   }
 
--- a/src/common/components/signUpModal/signUpModal.component.js
+++ b/src/common/components/signUpModal/signUpModal.component.js
@@ -5,6 +5,10 @@ import pick from 'lodash/pick'
 import 'rxjs/add/operator/finally'
 import { map, catchError } from 'rxjs/operators'
 import { Observable } from 'rxjs/Observable'
+import merge from 'lodash/merge'
+import 'rxjs/add/observable/forkJoin'
+import 'rxjs/add/operator/map'
+import 'rxjs/add/operator/switchMap'
 import OktaSignIn from '@okta/okta-signin-widget'
 import sessionService, { Roles } from 'common/services/session/session.service'
 import orderService from 'common/services/api/order.service'
@@ -37,6 +41,7 @@ class SignUpModalController {
     this.geographiesService = geographiesService
     this.imgDomain = envService.read('imgDomain')
     this.publicCru = envService.read('publicCru')
+    this.$injector = angular.injector()
   }
 
   $onInit () {
@@ -52,6 +57,9 @@ class SignUpModalController {
 
   $onDestroy () {
     this.oktaSignInWidget?.remove()
+    if (angular.isDefined(this.getDonorDetailsSubscription)) {
+      this.getDonorDetailsSubscription.unsubscribe()
+    }
   }
 
   initializeVariables () {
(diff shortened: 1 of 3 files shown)


src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -451,6 +447,15 @@ class SignUpModalController {
     })
   }
 
+  showVerificationCodeField () {
+    // The verification code field is only shown when the button link "Enter a verification code instead" is clicked.
+    // This makes the process of creating an account more streamlined as we remove that click.
+    const verificationCodeButtonLink = document.querySelector('.button-link.enter-auth-code-instead-link')
+    if (verificationCodeButtonLink) {
+      verificationCodeButtonLink.click();
+    }
```
canac, line 456: I don't think an `if` is bad, but you could also write that like this:

```suggestion
    verificationCodeButtonLink?.click();
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -636,6 +649,52 @@ class SignUpModalController {
...
+  saveDonorDetails () {
+    const signUpDonorDetails = {
+      name: {
+        'given-name': this.$scope.firstName,
+        'family-name': this.$scope.lastName
+      },
+      'donor-type': this.$scope.accountType,
+      'organization-name': this.$scope.organizationName,
+      email: this.$scope.email,
+      phone: this.$scope.primaryPhone,
```
canac, line 662: See #1160

```suggestion
      'phone-number': this.$scope.primaryPhone,
```

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -636,6 +649,52 @@ class SignUpModalController {
...
+        region: this.$scope.state,
+        postalCode: this.$scope.zipCode,
+        country: this.$scope.countryCode
+      }
+    }
+
+    if (angular.isDefined(this.getDonorDetailsSubscription)) {
+      this.getDonorDetailsSubscription.unsubscribe()
+    }
+    this.getDonorDetailsSubscription = this.orderService.getDonorDetails().switchMap((donorDetails) => {
```
canac, line 678: Did you want to remove this logic from `src/common/components/registerAccountModal/registerAccountModal.component.js` or are you doing that later?

canac (changes requested):
Thanks for putting the verify email address input to step 3! It looks like the CSS rule that adds 15px of `bottom-margin` needs to be overridden.

Screenshot 2025-03-04 at 3 29 02 PM

When that margin is removed, some work will also need to be done to give the error message rounded borders.

Screenshot 2025-03-04 at 3 29 26 PM

The Next button after putting in the verification code doesn't match our brand:

Screenshot 2025-03-04 at 3 33 14 PM

The form to submit the TOTP for Google Authenticator needs a rounded focus border, and the label doesn't float up when focused like the other Material Design labels do. The whitespace around the "Enter code displayed from application" label also looks a little funny.

Screenshot 2025-03-04 at 3 35 41 PM

The continue button after setting up Google Authenticator also needs to match our branding:

Screenshot 2025-03-04 at 3 38 08 PM

After clicking Continue to create my account, I saw this error in the DevTools console multiple times: `Logger.js:19 [okta-signin-widget] "afterRender" event handler error: TypeError: Cannot read properties of undefined (reading 'forEach')`

When I login with the account, it saved my name and email address, but not my phone number or address. A comment a left should fix the phone number, but I'm not sure what's wrong with the address.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I will clean this up in a later PR

dr-bizz:
I'm hoping Jason can remove the optional MFA step, as it's really not needed. and it would remove the need to style and write JS for all of these MFA steps

canac (approved):
I'm going to call this good for now! Like you said hopefully we can turn off the MFA step so that we don't have to do the styling work for it. I'm curious if removing those steps will also fix the `Logger.js:19 [okta-signin-widget] "afterRender" event handler error: TypeError: Cannot read properties of undefined (reading 'forEach')` error I was seeing. I can retest after Jason makes his changes.

Can you run `yarn lint` and fix the failing test before merging?

PR #1162, opened 2025-03-05
[no-Jira] Set up Stylelint extension to format on save

canac:
Add VS Code configuration to make the Stylelint extension format SCSS files on save.


dr-bizz (approved):
Awesome!

PR #1163, opened 2025-03-06
[EP-2362] Improve Okta signup errors

canac:
## Changes

* I noticed that we are using `innerHTML` in one place with a string that comes back from the Okta API. This creates the possibility of a script-injection attack, so I refactored it. I also documented the other places that use `innerHTML` to note that they use hard-coded strings to draw attention to the fact that `innerHTML` must not be set to untrusted strings.
* `errorSummary` can be an array, so I added support for that.
* The password error message had rounded borders when it should not.
* I updated the "We found some errors" error box to list the fields that had errors. Since the email and password fields are the only ones that can have errors from the backend that aren't caught on the frontend, I am looking at the machine-readable error properties to convert them into a human-readable field name.
* I changed the injected error messages to clear the previous messages before creating new ones. This prevents stale error messages from sticking around.
* The "Create Account" button was being reverted back to its default text after a failed submission.
* The "Continue" button's focus ring was being clipped on the right side.
  * Before Screenshot 2025-03-06 at 2 03 59 PM
  * After Screenshot 2025-03-06 at 2 03 43 PM
* Improvements to the email verification step
  * Update UI to show that the user is on step 3
  * Remove the back button because Okta will just take them back to the verification step anyway
  * Update the next button text to be "Verify"
* The resend email icon was too short
  * Before Before
  * After After
  * I also submitted a PR with Okta's sign-in widget to fix that bug there https://github.com/okta/okta-signin-widget/pull/3801

@dr-bizz I need to write tests for this PR, but I'd love your feedback first to make sure I'm going in the right direction with these changes.

EP-2362

### DIFF
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -169,7 +169,7 @@ sign-up-modal {
 
       // We don't include password as it's the only field on it's step and has the password requirements below it.
       input[type=text]:not(.chzn-search input),
-      .o-form-input-error,
+      &:not([data-se="o-form-fieldset-credentials.passcode"]) .o-form-input-error,
       .chzn-container a {
         border-bottom-left-radius: 6px;
         border-bottom-right-radius: 6px;
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -27,6 +27,11 @@ sign-up-modal {
 #okta-sign-in.auth-container {
   $okta-grey: #f5f5f5;
   width: 100%;
+  overflow: visible;
+
+  .beacon-container {
+    display: none;
+  }
 
   // ============== Container ============== //
 
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -29,6 +29,14 @@ sign-up-modal {
   width: 100%;
   overflow: visible;
 
+  .clearfix {
+    &:after, &:before {
+      // Fix warning icon height
+      // This may be removed if https://github.com/okta/okta-signin-widget/pull/3801 is merged upstream
+      height: calc(100% + 2px);
+    }
+  }
+
   .beacon-container {
     display: none;
   }
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -171,13 +171,28 @@ sign-up-modal {
     }
 
     &:last-of-type {
-
-      // We don't include password as it's the only field on it's step and has the password requirements below it.
-      input[type=text]:not(.chzn-search input),
-      &:not([data-se="o-form-fieldset-credentials.passcode"]) .o-form-input-error,
-      .chzn-container a {
+      // Usually the last fieldset has rounded bottom borders
+      // ".chzn-container a" is for select boxes
+      input[type=text]:not(.chzn-search input), .o-form-input-error, .chzn-container a {
         border-bottom-left-radius: 6px;
         border-bottom-right-radius: 6px;
+      }
+
+      // However, if it is a password field with a password requirements sibling, the password
+      // requirements box gets the rounded border instead of the input or error message
+      &[data-se="o-form-fieldset-credentials.passcode"]:has(+ [data-se="password-authenticator--rules"]) {
+        input[type=text]:not(.chzn-search input), .o-form-input-error {
+          border-bottom-left-radius: unset;
+          border-bottom-right-radius: unset;
+        }
+      }
+
+      // Also, if the field has errors, the error box gets the rounded border instead of the input
+      &:has(.okta-form-input-error) {
+        input[type=text]:not(.chzn-search input) {
+          border-bottom-left-radius: unset;
+          border-bottom-right-radius: unset;
+        }
       }
     }
 
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -170,8 +170,7 @@ sign-up-modal {
 
     &:first-of-type {
 
-      input[type=text]:not(.chzn-search input),
-      input[type=password],
+      input,
       .chzn-container a {
         border-top-left-radius: 6px;
         border-top-right-radius: 6px;
@@ -181,7 +180,9 @@ sign-up-modal {
     &:last-of-type {
       // Usually the last fieldset has rounded bottom borders
       // ".chzn-container a" is for select boxes
-      input[type=text]:not(.chzn-search input), .o-form-input-error, .chzn-container a {
+      input,
+      .o-form-input-error,
+      .chzn-container a {
         border-bottom-left-radius: 6px;
         border-bottom-right-radius: 6px;
       }
@@ -189,18 +190,17 @@ sign-up-modal {
       // However, if it is a password field with a password requirements sibling, the password
       // requirements box gets the rounded border instead of the input or error message
       &[data-se="o-form-fieldset-credentials.passcode"]:has(+ [data-se="password-authenticator--rules"]) {
-        input[type=text]:not(.chzn-search input), .o-form-input-error {
+        input,
+        .o-form-input-error {
           border-bottom-left-radius: unset;
           border-bottom-right-radius: unset;
         }
       }
 
       // Also, if the field has errors, the error box gets the rounded border instead of the input
-      &:has(.okta-form-input-error) {
-        input[type=text]:not(.chzn-search input) {
-          border-bottom-left-radius: unset;
-          border-bottom-right-radius: unset;
-        }
+      &:has(.okta-form-input-error) input {
+        border-bottom-left-radius: unset;
+        border-bottom-right-radius: unset;
       }
     }
 
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -155,6 +155,12 @@ sign-up-modal {
     border-top: 1px solid $okta-grey;
     width: 100%;
 
+    input:-webkit-autofill {
+      // Override Okta's webkit autofill styles
+      border-radius: unset;
+      -webkit-box-shadow: unset;
+    }
+
     &[data-se="o-form-fieldset-userProfile.state"],
     &[data-se="o-form-fieldset-userProfile.zipCode"] {
       width: 50%;
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -389,7 +389,10 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     SOMETHING_WENT_WRONG: 'Something went wrong. Please try again',
     OKTA_EMAIL_ALREADY_EXISTS: 'The email address you used belongs to an existing Okta user.',
     OKTA_ERROR_WHILE_SAVING_EMAIL: 'There was an error saving your email address. Make sure it was entered correctly.',
-    OKTA_ERROR_WHILE_SAVING_DATA: 'There was an error saving your contact info. Please try again or contact eGift@cru.org for assistance.'
+    OKTA_ERROR_WHILE_SAVING_DATA: 'There was an error saving your contact info. Please try again or contact eGift@cru.org for assistance.',
+    OKTA_SIGNUP_FIELDS_ERROR: 'Please review the following fields: {{fields}}.',
+    OKTA_EMAIL_FIELD: 'email',
+    OKTA_PASSWORD_FIELD: 'password'
   })
 
   $translateProvider.translations('es', {
@@ -603,7 +606,14 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     RADIO_STATION: 'Radio Station',
     RADIO_STATION_LIST_ERROR: 'There was an error loading radio stations in your area.',
     RADIO_STATION_SELECT_ERROR: 'There was an error selecting a radio station.',
-    PREFERRED_RADIO_STATION: 'Preferred Radio Station: '
+    PREFERRED_RADIO_STATION: 'Preferred Radio Station: ',
+    SOMETHING_WENT_WRONG: 'Something went wrong. Please try again',
+    OKTA_EMAIL_ALREADY_EXISTS: 'The email address you used belongs to an existing Okta user.',
+    OKTA_ERROR_WHILE_SAVING_EMAIL: 'There was an error saving your email address. Make sure it was entered correctly.',
+    OKTA_ERROR_WHILE_SAVING_DATA: 'There was an error saving your contact info. Please try again or contact eGift@cru.org for assistance.',
+    OKTA_SIGNUP_FIELDS_ERROR: 'Please review the following fields: {{fields}}.',
+    OKTA_EMAIL_FIELD: 'email',
+    OKTA_PASSWORD_FIELD: 'password'
   })
   $translateProvider.preferredLanguage('en')
 }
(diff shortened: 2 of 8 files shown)


src/common/app.config.js
```diff
@@ -387,9 +387,9 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     RADIO_STATION_SELECT_ERROR: 'There was an error selecting a radio station.',
     PREFERRED_RADIO_STATION: 'Preferred Radio Station: ',
     SOMETHING_WENT_WRONG: 'Something went wrong. Please try again',
-    OKTA_EMAIL_ALREADY_EXISTS: 'The email address you used belongs to an existing Okta user.',
-    OKTA_ERROR_WHILE_SAVING_EMAIL: 'There was an error saving your email address. Make sure it was entered correctly.',
-    OKTA_ERROR_WHILE_SAVING_DATA: 'There was an error saving your contact info. Please try again or contact eGift@cru.org for assistance.'
```
dr-bizz, line 392: We want to keep this. I know we're not currently using it, but if there was an error we need to show this text. I have created a task to create these pages.

dr-bizz: It prob needs to be added to the ES translation too.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -406,6 +422,12 @@ class SignUpModalController {
     // Save errors to local variable to inject into the form
     // Since errors are cleared on each step change
     this.signUpErrors = error.xhr.responseJSON.errorCauses
+    // Wait for the Okta widget to create the error message box and set the sign up button text
+    // before augmenting the error message box and resetting the sign up button text
+    this.$timeout(() => {
+      this.updateSignUpButtonText()
```
dr-bizz, line 428: Why have you moved the `updateSignUpButtonText` function into the `afterError` function?

This needs to go in the `afterRender` function as the buttons on the steps is now "Sign up".

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -420,12 +442,21 @@ class SignUpModalController {
...
+      // Verification step page
+      this.$scope.$apply(() => {
+        this.currentStep = 4
+      })
+      this.showVerificationCodeField()
+    }
+
     // All steps
-    this.updateSignUpButtonText()
+    this.injectBackButton()
```
dr-bizz, line 454: You've added 2 `injectBackButton` functions

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -435,11 +466,6 @@ class SignUpModalController {
     if (this.loadingRegionsError && this.currentStep === 2) {
       this.injectRegionLoadError()
     }
-    // Step 3: Security
-    this.showVerificationCodeField()
-
-    // This needs to be last to ensure even the verification code field is styled correctly
```
dr-bizz, line 441: I think we should keep this comment in

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -420,12 +442,21 @@ class SignUpModalController {
       return
     }
 
+    if (context.formName === 'enroll-authenticator') {
```
dr-bizz, line 445:
```suggestion
    // Step 4: Email Verification
    if (context.formName === 'enroll-authenticator') {
```

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: You should leave the `updateSignUpButtonText` in the `afterRender` function and either call it again or create a new function that fixes the submit button, but only when it's on step 3.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -420,12 +442,21 @@ class SignUpModalController {
...
+      // Verification step page
+      this.$scope.$apply(() => {
+        this.currentStep = 4
+      })
```
dr-bizz, line 449:
```
this.currentStep = 4
```
This should be added to the `postSubmit` function as that is the function that runs when the data is successfully saved

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -476,28 +505,71 @@ class SignUpModalController {
   }
 
   injectErrorMessages (errors = this.signUpErrors) {
```
dr-bizz, line 507: I'm unsure what is broken, but the errors are no longer showing next to the fields and I'm unable to see the errors on step 3

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -476,28 +505,71 @@ class SignUpModalController {
...
+    this.clearInjectedErrorMessages()
+
+    if (!errors) {
```
dr-bizz, line 510:
```suggestion
    if (!errors.length) {
```
We should check if errors length is falsey.

When testing `!['a', 'b']` it returns `false`.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -476,28 +505,71 @@ class SignUpModalController {
...
-        const errorText = `${errorIconHtml} ${error.errorSummary}`
-        if (!existingErrorParentElement || existingErrorParentElement.innerHTML !== errorText) {
-          const errorElement = document.createElement('div')
-          errorElement.classList.add('okta-form-input-error', 'o-form-input-error', 'o-form-explain')
-          field.parentNode.classList.add('o-form-has-errors')
-          errorElement.setAttribute('role', 'alert')
-          errorElement.innerHTML = errorText
-          field.parentNode.appendChild(errorElement)
-        }
+        const errorElement = document.createElement('div')
```
dr-bizz, line 518: Why did you remove the code that ensures we only add the error to the field once?

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I've made sure I downloaded your latest version.

src/assets/scss/components/_oktaSigninWidget.scss
```diff
@@ -166,14 +179,29 @@ sign-up-modal {
...
 
     &:last-of-type {
-
-      // We don't include password as it's the only field on it's step and has the password requirements below it.
-      input[type=text]:not(.chzn-search input),
-      .o-form-input-error,
-      .chzn-container a {
+      // Usually the last fieldset has rounded bottom borders
+      // ".chzn-container a" is for select boxes
+      input[type=text], .o-form-input-error, .chzn-container a {
```
dr-bizz, line 184: `input[type=text]:not(.chzn-search input)` is needed otherwise the search box inside country field has rounded borders too, which we don't want.

dr-bizz (changes requested):
Good start! There are a few things which need adjusting. I believe I had the latest copy of your code.

src/common/components/signUpModal/signUpModal.component.js
canac: * Go through the sign up flow to create an account, but don't enter the verification code.
* Close the sign up modal.
* Open the sign up modal again.

If you follow those steps, Okta will immediately take you to the verification step. I was thinking that `afterRender` would be the best place to detect that we're on the `enroll-authenticator` form. It can't just go in `postSubmit` because `postSubmit` isn't called when the user opens the sign up modal the second time. But please let me know if there is a better place to put this.

src/common/components/signUpModal/signUpModal.component.js
canac: The goal here was to handle the case where errors was `null` or `undefined`. It was an alternative to `errors?.forEach`. But since I don't see a way that `errors` could not be an array, I'll just remove this check.

src/common/components/signUpModal/signUpModal.component.js
canac: I added `this.clearInjectedErrorMessages()` to the top of the function to delete all error messages.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: This is what I'm seeing on your branch when running it locally on an incognito window. The button isn't changed.

Screenshot 2025-03-07 at 3 39 46 PM

dr-bizz: Sorry, I don't fully understand your suggestion.

Due to the button not being changed, we need to ensure that`updateSignUpButtonText` runs on each page.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: The errors are showing up now. Unsure why they weren't before.
Screenshot 2025-03-07 at 3 50 00 PM
Screenshot 2025-03-07 at 3 50 08 PM

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: I like the check, as it means we don't run the entire function on each step if there are no errors.

src/common/components/signUpModal/signUpModal.component.js
dr-bizz: Awesome!

src/assets/scss/components/_oktaSigninWidget.scss
```diff
@@ -157,23 +170,36 @@ sign-up-modal {
...
+      // ".chzn-container a" is for select boxes
+      input:not(.chzn-search input), .o-form-input-error, .chzn-container a {
         border-bottom-left-radius: 6px;
         border-bottom-right-radius: 6px;
       }
+
+      // However, if it is a password field with a password requirements sibling, the password
+      // requirements box gets the rounded border instead of the input or error message
+      &[data-se="o-form-fieldset-credentials.passcode"]:has(+ [data-se="password-authenticator--rules"]) {
+        input:not(.chzn-search input), .o-form-input-error {
```
dr-bizz, line 190: I don't think you need this here as it won't have a dropdown.
```suggestion
        input, .o-form-input-error {
```

src/assets/scss/components/_oktaSigninWidget.scss
```diff
@@ -157,23 +170,36 @@ sign-up-modal {
...
+          border-bottom-left-radius: unset;
+          border-bottom-right-radius: unset;
+        }
+      }
+
+      // Also, if the field has errors, the error box gets the rounded border instead of the input
+      &:has(.okta-form-input-error) {
+        input:not(.chzn-search input) {
```
dr-bizz, line 198: again here
```suggestion
        input
```

src/common/components/signUpModal/signUpModal.component.js
canac: I think what happened is I accidentally changed the `updateSignUpButtonText` call to `injectBackButton` and them removed it when you [pointed out](https://github.com/CruGlobal/give-web/pull/1163#discussion_r1985126680) I had two `injectBackButton` calls. I added back the `updateSignUpButtonText` call to `afterRender` since I didn't mean to remove it.

src/assets/scss/components/_oktaSigninWidget.scss
canac: Speaking of the `:not(.chzn-search input)`, I don't think that it is needed anywhere. I made a screen recording with audio so I could explain what I was seeing. But I'm thinking we can drop the `:not(.chzn-search input)` on all of these selectors because `.chzn-search input` never selects anything visible to the user. But maybe I'm missing something.

https://github.com/user-attachments/assets/d698b4dd-687d-4ae1-88ba-1ee3d05fc088

dr-bizz: You're correct; we can remove it. Apart from line 199, we need to have the selector there
```
input[type=text]:not(.chzn-search input
```

canac: Why is the one on line 199 needed? Everything is still looking good without it.

dr-bizz: When testing it locally, the account type select input was styled incorrectly.

canac: I'm still not seeing any issues with account type selector, but do you mind sharing a screenshot, and I'll add the `:not` part of the selector back?

dr-bizz: I think we're on about different lines, as it's still included in your PR. It was on line `199` at one point, but now it's changed to `227`.

https://github.com/CruGlobal/give-web/blob/bc9398ecb8683b0d0cce9bfdf87d5455b6cf1b9f/src/assets/scss/components/_oktaSigninWidget.scss#L227

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -596,16 +601,12 @@ class SignUpModalController {
...
-    errorElement.innerHTML = errorMessage
+    errorElement.innerHTML = this.$sanitize(errorMessage)
     errorElement.prepend(createErrorIcon())
 
     const retryButtonElement = document.createElement('a')
     retryButtonElement.classList.add('cru-retry-button')
-    // SECURITY: this.translations contains only trusted first-party strings
-    // innerHTML must not be set to an untrusted string
-    retryButtonElement.innerHTML = this.translations.retry
+    retryButtonElement.innerHTML = this.$sanitize(this.translations.retry)
```
dr-bizz, line 609: Good call using sanitize

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -1066,6 +1137,35 @@ describe('signUpForm', function () {
...
+            property: 'accountType',
+            errorSummary
+          }
+        ]
+        $ctrl.injectErrorMessages()
+
+        const errorElement = document.querySelector(inputErrorFieldSelector);
+        expect(errorElement.innerHTML).not.toContain(errorSummary);
+        expect(errorElement.innerHTML).toContain('&lt;script&gt;');
+      });
```
dr-bizz, line 1167: Thank you again for adding this

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -1110,6 +1210,44 @@ describe('signUpForm', function () {
...
+
+        afterEach(() => {
+          // Ensure that the text provided by Okta is not removed
+          expect(errorBox.textContent).toContain('There was an error')
+        })
+
+        it('notifies the user which fields had errors in the error box', () => {
+          $ctrl.injectErrorMessages()
+
+          expect(errorBox.textContent).toContain('Please review the following fields: email, password.')
```
dr-bizz, line 1234: Does name not show up?

dr-bizz (approved):
Thank you for your work on this! It's really great.  I'm going to go ahead and approve this as I think it's good to go, but I've also added some slight improvements or comments.

Something else I've spotted while testing is when Google enters the data, the input borders get rounded. You should be able to fix it with this style.
```
#okta-sign-in .o-form .input-fix input:-webkit-autofill, #okta-sign-in .o-form .textarea-fix input:-webkit-autofill {
    border-radius: 0;
}
```

src/common/components/signUpModal/signUpModal.component.spec.js
dr-bizz: I haven't tested that, just a thought

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -263,6 +265,9 @@ class SignUpModalController {
     this.loadingCountriesError = false
     return this.geographiesService.getCountries()
       .map((data) => {
+        // Order countries in alphabetical order
+        data.sort((a, b) => a['display-name'].localeCompare(b['display-name']))
+
```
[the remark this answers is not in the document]
canac: > Can you also make 'USA' at the top of the list so people don't have to scroll as much.

This field defaults to USA, so US donors won't have to scroll through the list at all. That makes me lean towards keeping them in sorted order.

src/assets/scss/components/_oktaSigninWidget.scss
canac: Ah, sorry for any misunderstanding. I'm glad we're both on the same page.

src/common/components/signUpModal/signUpModal.component.js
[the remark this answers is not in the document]
dr-bizz: Looking at the way we currently sort the countries. The way you are suggesting is the correct way.

canac:
> Something else I've spotted while testing is when Google enters the data, the input borders get rounded. You should be able to fix it with this style.

@dr-bizz I'm not sure why autofill is triggering for you with the `autocomplete=true` set on those inputs, but I added some styles to override Okta's autofill styles just in case autofill gets triggered for some users.

dr-bizz (approved):
Great idea to make all places with regions and countries sorted the same

PR #1164, opened 2025-03-07
This updates some copy in the registerAccountModal based on feedback from USDS

rguinee:
The writing team from USDS helped review the copy in the modal and they made these small suggested changes.


canac (approved):
Thanks for the improvements!

PR #1165, opened 2025-03-07
2362 okta account creation flow

dr-bizz:
# Description
In this PR, I finish the coding for the sign up process. Up to now, we have the Okta form done, but in this PR, I handle saving the user data to the Give information, redirecting to Okta to ensure the user logs in to the browser, and, upon returning, handle the user matching.

You will also see that I've cleaned up some code in the registerAccountModal files.

## Sign up flows

_When testing, use your email address, but change it by adding the `+` symbol. such as `daniel.bisgrove+testEmail236@cru.org`_

### Error while saving data to Give DB (Cortex)
If there is an error saving the data on the sign-up form (Before redirecting to Okta), we send users to the contact-info modal. Once they fix the details and click save, we redirect the user to Okta to complete the registration.

**To Test:**
1. Change line 660 to `email: '111'` to cause an error while saving the email.
2. Sign up as a user (Entering your details)
3. After you enter the verification code and MFA options, you should be redirected to the contact info modal.
4. After fixing your email, you will be redirected to Okta to authentication.
5. Upon returning to the Give site, you will go through the user matching modal or be redirected to your previous page.
6. Revert changes to line 660.

### Catch user that has Okta account but no Give account
If there is an error saving the Give DB data on the sign-up form (Before redirecting to Okta), but the Okta account has been created, then the session is lost due to the user vanishing. We need to ensure when they log back in; we continue to finish off the signup process.

**To Test:**
1. Change line 660 to `email: '111'` to cause an error while saving the email.
2. Sign up as a user (Entering your details)
3. After you enter the verification code and MFA options, you should be redirected to the contact info modal.
4. Close your tab/session.
5. Open a new session on an incognito window and log into Okta with the account you just created.
6. Upon returning to the Give site, you should be sent to the contact info modal to fix errors with the account.
7. After this, you will be sent through the user-matching modals

### Go through user matching
If you enter data for the sign-up form, similar to another account's, it will show the user modal after being redirected from Okta in the registration process.

**To Test:**
1. Sign up as a new user, entering details similar to an existing account
3. Go through the email verification and MFA steps.
4. You will be redirected ti Okta. Login with your new account.
5. Upon returning to the Give site, you will go through the user matching modal.

## Tests
I have added tests to each function to ensure they work as expected.

## Loading and error pages
I still need to add error pages and better loading pages
- Show Error while loading contact info modal when there was an error saving the Cortex account.
- Show better loading when switching from the Signup modal to directing to Okta for login.
- Add error when Okta and Cortex doesn't work and we need to instruct the user to contact Cru.

### DIFF
--- a/src/common/components/registerAccountModal/registerAccountModal.component.js
+++ b/src/common/components/registerAccountModal/registerAccountModal.component.js
@@ -13,14 +13,6 @@ import userMatchModal from 'common/components/userMatchModal/userMatchModal.comp
 import contactInfoModal from 'common/components/contactInfoModal/contactInfoModal.component'
 import failedVerificationModal from 'common/components/failedVerificationModal/failedVerificationModal.component'
 
-import { Observable } from 'rxjs/Observable'
-import 'rxjs/add/observable/forkJoin'
-import 'rxjs/add/observable/of'
-import 'rxjs/add/operator/do'
-import 'rxjs/add/operator/map'
-import 'rxjs/add/operator/switchMap'
-import merge from 'lodash/merge'
-
 const componentName = 'registerAccountModal'
 
 class RegisterAccountModalController {
@@ -133,7 +125,7 @@ class RegisterAccountModalController {
     this.onSuccess()
   }
 
-  checkDonorDetails (signUpDonorDetails) {
+  checkDonorDetails () {
     // Show loading state
     this.stateChanged('loading-donor')
 
@@ -141,27 +133,7 @@ class RegisterAccountModalController {
     if (angular.isDefined(this.getDonorDetailsSubscription)) {
       this.getDonorDetailsSubscription.unsubscribe()
     }
-    this.getDonorDetailsSubscription = this.orderService.getDonorDetails().switchMap((donorDetails) => {
-      if (signUpDonorDetails && donorDetails['registration-state'] === 'NEW') {
-        // Save the contact info from signup
-        merge(donorDetails, signUpDonorDetails)
-
-        // Send each of the requests and pass donorDetails to the next step after the requests complete
-        return Observable.forkJoin([
-          this.orderService.updateDonorDetails(donorDetails),
-          this.orderService.addEmail(donorDetails.email, donorDetails.emailFormUri)
-        ]).map(() => donorDetails).do({
-          error: () => {
-            // If there was an error, save the donor details from sign up so that they will be added
-            // to the contact info form. The error handler below will change the step to contact-info.
-            this.signUpDonorDetails = signUpDonorDetails
-          }
-        })
-      }
-
-      // Pass donorDetails to the next step
-      return Observable.of(donorDetails)
-    }).subscribe({
+    this.getDonorDetailsSubscription = this.orderService.getDonorDetails().subscribe({
       next: (donorDetails) => {
         // Workflow Complete if 'registration-state' is COMPLETED
         if (donorDetails['registration-state'] === 'COMPLETED') {
--- a/src/common/components/registerAccountModal/registerAccountModal.component.js
+++ b/src/common/components/registerAccountModal/registerAccountModal.component.js
@@ -35,10 +35,11 @@ class RegisterAccountModalController {
   // 5. Complete User Match
 
   /* @ngInject */
-  constructor ($element, $rootScope, $window, cartService, orderService, sessionService, verificationService, envService, gettext) {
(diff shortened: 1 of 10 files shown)


src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -35,10 +27,11 @@ class RegisterAccountModalController {
   // 5. Complete User Match
 
   /* @ngInject */
-  constructor ($element, $rootScope, $window, cartService, orderService, sessionService, verificationService, envService, gettext) {
+  constructor ($element, $rootScope, $window, $document, cartService, orderService, sessionService, verificationService, envService, gettext) {
     this.element = $element[0]
     this.$rootScope = $rootScope
     this.$window = $window
+    this.$document = $document
```
canac, line 34:
```suggestion
    this.$document = $document
    this.$rootScope = $rootScope
    this.$window = $window
```

src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -196,6 +179,19 @@ class RegisterAccountModalController {
     modal.removeClass('modal-sm modal-md modal-lg')
     modal.addClass(`modal-${size}`)
   }
+
+  redirectToOktaForLogin () {
+    this.sessionService.signIn(this.lastPurchaseId).subscribe(() => {
+      const $injector = this.$injector
```
canac, line 185: Since the injector is only used once, can we create it here instead of using an instance variable? That minimizes the distance between where the variable is defined and where it is used and makes it easier to read the code.

```suggestion
      const $injector = angular.injector()
```

src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -344,13 +345,13 @@ describe('signUpForm', function () {
...
         $ctrl.$scope.streetAddress = user.streetAddress;
-        $ctrl.$scope.streetAddressExtended = user.streetAddressExtended;
+        $ctrl.$scope.streetAddressExtended = '';
         $ctrl.$scope.internationalAddressLine3 = 'internationalAddressLine3';
         $ctrl.$scope.internationalAddressLine4 = 'internationalAddressLine4';
         $ctrl.$scope.city = user.city;
         $ctrl.$scope.state = user.state;
         $ctrl.$scope.zipCode = user.zipCode;
-        $ctrl.$scope.primaryPhone = user.primaryPhone;
+        $ctrl.$scope.primaryPhone = '';
```
canac, line 354: Why are we using empty strings instead of the extended address and phone number?

canac (commented):
Just a few more questions and observations, but overall this still looks really good.

dr-bizz:
I tested the steps you outlined, and everything appears to be working great. In "Catch user that has Okta account but no Give account", the mailing address is not prefilled when you come back to the contact info modal, but I think that is inevitable because we only saved their address in memory. I'm going to review the tests now.

I had thought about this, but if we stored the address, I would want to limit the amount of time before the cookie expired to ensure security. However, the cookie will likely expire before they finish the sign up.

src/common/components/registerAccountModal/registerAccountModal.component.js
```diff
@@ -196,6 +179,19 @@ class RegisterAccountModalController {
...
+      if (!$injector.has('sessionService')) {
+        $injector.loadNewModules(['sessionService'])
```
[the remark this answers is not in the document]
dr-bizz: I believe we load it and send it with the window event so AEM can access the user's name.

src/common/components/registerAccountModal/registerAccountModal.component.js
dr-bizz: Good point

src/common/components/signUpModal/signUpModal.component.spec.js
dr-bizz: Thank you for catching that. I've fixed it now

src/common/components/registerAccountModal/registerAccountModal.component.js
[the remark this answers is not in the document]
canac: I don't think that snippet is needed in production either because `signInForm` injects `sessionService`. Unless I'm misunderstanding AngularJS' injection process, I don't see how `sessionService` could not be loaded inside of `signInForm`.

That doesn't necessarily need to be fixed here though.

canac (approved):
A test needs fixed, and there is one standardjs violation, but after that this looks great!

PR #1168, opened 2025-03-11
Fix deployment errors caused by PR #1154

dr-bizz:
# Description
A previous PR https://github.com/CruGlobal/give-web/pull/1154 broke the deployment action. The error occurred during the build script. This wasn't caught before merging it into the `master` branch, causing the deployment to production to break.

The error was due to the `MiniCssExtractPlugin` not being included as a plugin when building the app - https://github.com/CruGlobal/give-web/actions/runs/13790159211/job/38567765240

### Changes
**MiniCssExtractPlugin**
Having updated `MiniCssExtractPlugin` package in the previous PR, the `disabled` property was no longer available, so in https://github.com/CruGlobal/give-web/pull/1154, I had meant to conditionally load `MiniCssExtractPlugin` when building the app. However, I must have got confused as I loaded the plugin during development instead.
In this PR, I have downgraded it, so it's back at the original version and added back the `disable` property.

**StyleLintPlugin**
I have also conditionally added the StyleLintPlugin to only run during `development`. This plugin lints the style files, so it shouldn't be running during the build, as we have a CI that tests the style lint before we build.

**Added build CI test**
I've added a build CI test to avoid this happening again. 

## Why this wasn't caught?
1. I missed it. I had checked that `build` was working, but then did a number of changes and didn't recheck it.
2. The PR was passing all its tests.
3. We don't have a CI test that tests the app can be built without errors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build
+name: CI
 
 on:
   push: 
@@ -20,6 +20,7 @@ permissions:
 
 jobs:
   lint:
+    name: 🧹 Lint Check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -39,6 +40,7 @@ jobs:
         run: yarn lint:styles
 
   test:
+    name: ✅ Tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -80,6 +82,7 @@ jobs:
 
 
   deploy:
+    name: 🚀 Deploy app
     runs-on: ubuntu-latest
     environment:
       name: ${{ (github.ref == 'refs/heads/master' && 'production') || 'staging' }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,27 @@ jobs:
       
       - name: Upload Codecov Reports
         uses: codecov/codecov-action@v2 
-   
+
+  build:
+    name: 🏗️ Build app
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .tool-versions
+      
+      - name: Install Dependencies
+        run: yarn install --immutable --immutable-cache --ignore-optional
+      
+      - name: Build App
+        env:
+          S3_GIVE_DOMAIN: //${{ secrets.GIVE_WEB_HOSTNAME }}
+          ROLLBAR_ACCESS_TOKEN: ${{ secrets.ROLLBAR_ACCESS_TOKEN }}
+          DATADOG_RUM_CLIENT_TOKEN: ${{ secrets.DATADOG_RUM_CLIENT_TOKEN }}
+        run: yarn run build
+
+
   deploy:
     runs-on: ubuntu-latest
     environment:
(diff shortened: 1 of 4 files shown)


canac (commented):
Before the stylelint PR, the `MiniCssExtractPlugin` config was `disable: !isBuild`. So instead of always including the `MiniCssExtractPlugin` like this PR does, shouldn't we just flip the conditional to only include it if `isBuild` (instead of `!isBuild`)? I'm assuming we don't need this plugin to run in development and it may slow things down a little bit.

There are also some standardjs issues to fix.

dr-bizz:
> o instead of always including the `MiniCssExtractPlugin` like this PR does, shouldn't we just flip the conditional to only include it if `isBuild` (instead of `!isBuild`)? I'm assuming we don't need this plugin to run in development and it may slow things down a little bit.

it needs to be included in both. Otherwise, it will throw an error asking for it to be included.


I believe this is due to the rule:

```
      {
        test: /\.s?[ac]ss$/,
        include: path.resolve(__dirname, './src/assets'),
        use: [
          MiniCssExtractPlugin.loader,
          'css-loader?-url',
          {
            loader: 'sass-loader',
            options: {
              sourceMap: true,
              sassOptions: {
                quietDeps: true
              }
            }
          }
        ]
      },
```

dr-bizz:
Last night I had an idea of reverting `mini-css-extract-plugin` back to version `0.8.2`. After trying it this morning, I think it will work.

wrandall22:
Looks like the current CSS linting stuff never made it to `staging`.

wrandall22:
When I run `yarn`, I get a lock file that is out of sync.

dr-bizz:
Oh I see, it's adding the `^` on`mini-css-extract-plugin`

canac (approved):
Looks good! Was there a reason you upgraded `mini-css-extract-plugin` in the stylelint PR? I assumed the old version was incompatible, but was that not the case?

Also, looking at the code for that package, I don't think that `disabled` is even a valid option. It is referenced in [the code](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/v0.8.2/src/index.js) for version 0.8.2. I think it would be fine to put it back to version 1.x and always include it, just like you had in the first version of this PR. But the downgrade solution works too. Up to you.

dr-bizz:
> Was there a reason you upgraded mini-css-extract-plugin in the stylelint PR? I assumed the old version was incompatible, but was that not the case?


I think it hit an issue with `mini-css-extract-plugin`, and upgrading it solved the problem, but then I went a different direction and never downgraded it.

dr-bizz:
Also, looking at the code for that package, I don't think that disabled is even a valid option. It is referenced in [the code](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/v0.8.2/src/index.js) for version 0.8.2. I think it would be fine to put it back to version 1.x and always include it, just like you had in the first version of this PR. But the downgrade solution works too. Up to you.

When running `mini-css-extract-plugin` on version `1.x` it throws an error while running webpack due to `disable` not being a value option. I get this error when running version `1.6.2`
```
Options has an unknown property 'disable'. These properties are valid:
```

canac:
> When running `mini-css-extract-plugin` on version `1.x` it throws an error while running webpack due to `disable` not being a value option. I get this error when running version `1.6.2`
> 
> ```
> Options has an unknown property 'disable'. These properties are valid:
> ```

Definitely. I'm just saying `disable` doesn't appear to be a valid option in `0.8.2` either. It appears to be silently ignored.

PR #1171, opened 2025-03-13
Upgrading MiniCssExtractPlugin after PR #1168

dr-bizz:
# Description
In this PR, I upgraded the MiniCssExtractPlugin to use version 1.6.2. As Caleb pointed out in PR #1168, we are using the `disable` property, but it's silently erroring as `disable` was never a property.

Because of this, we can upgrade `MiniCssExtractPlugin` to the latest version that will run on Webpack 4 and remove the `disable` prop, as it has always run on `start` and `build` processors.


canac (approved):
Works great in staging!

PR #1174, opened 2025-03-13
[EP-2362] Rename street address label for consistency

canac:
This makes the first line "Street address" and the other lines "Address line 2/3/4". This looks better for international addresses, but do we need it to say "Street address line 2" for US addresses?


dr-bizz (approved):
We're good to change it to `Address line 2`

Thank yo for making this change.

PR #1175, opened 2025-03-13
[EP-2572] Only augment error message box once

canac:
Once or twice, I've seen two lines of "Please review the following fields: email." `afterError` was being called twice by Okta for some reason. The issue resolved itself once I closed and reopened the modal.  It's likely a bug in Okta's widget.

To prevent this, we need to make sure that `injectErrorMessages` can be safely called multiple times.

EP-2572


dr-bizz (approved):
Thank you for fixing this

PR #1177, opened 2025-03-14
[EP-2573] Prevent escape key from closing the sign in modal

canac:
Prevent the escape key from closing the sign in modal. This is to prevent confusion of users accidentally hitting escape (maybe when trying to close an autocomplete or dismiss an autofill suggestion), accidentally closing the sign in/sign up modal, and having to start over.

EP-2573


dr-bizz (approved):
Thank you!

PR #1179, opened 2025-03-14
[EP-2576] Improve Okta info line on narrow screens

canac:
Changes:
* Reduce the x-axis padding to reduce the amount of line-wrapping
* Vertically align Okta info items when they are forced to wrap

I worked with @rguinee on this.

Before:
Before

After:
After

EP-2576

### DIFF
--- a/src/assets/scss/components/_registerAccountModal.scss
+++ b/src/assets/scss/components/_registerAccountModal.scss
@@ -78,7 +78,7 @@ register-account-modal:has(sign-in-modal, sign-up-modal) {
     .cru-info {
       display: flex;
       justify-content: space-between;
-      padding: 1rem 2rem;
+      padding: 1rem;
       border-bottom: 2px solid $colorCru-gold;
       background-color: white;
       gap: 2rem;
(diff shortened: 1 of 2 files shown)


dr-bizz (commented):
Do we need the dot next to the Okta logo?

I understand why we have it on desktop as it's between 2 pieces of text, but on mobile, it seems a little redundant as the Okta logo splits up the text.

canac:
> Do we need the dot next to the Okta logo?
> 
> I understand why we have it on desktop as it's between 2 pieces of text, but on mobile, it seems a little redundant as the Okta logo splits up the text.

@rguinee Do you have thoughts? I feel like removing the bullet whether the text wraps or not looks OK. Also, if we drop the bullet, it's a lot easier to make "Powered by Okta" the same block (like below) instead of separate blocks. I think that change looks a little better.

Screenshot 2025-03-17 at 9 29 28 AM

canac:
I think it makes sense to wait on EP-2580 to see how the Okta info line gets changed on mobile (if at all), before proceeding with this PR.

dr-bizz (approved):
Great work

PR #1183, opened 2025-03-18
[EP-2362] Fix zip code height

canac:
Make the zip code input height the same as the state dropdown.

Introduced in #1181 

EP-2362

### DIFF
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -119,6 +119,10 @@ sign-up-modal {
         border: 1px solid transparent;
         border-radius: 0;
 
+        span {
+          line-height: 36px;
+        }
+
         div {
           top: 6px;
         }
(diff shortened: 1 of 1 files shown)


dr-bizz:
I also made this change by changing the line height. Could you either remove my changes and add yours or Vice versa?

`src/assets/scss/components/_oktaSigninWidget.scss`
```
#okta-sign-in.auth-container .o-form-select-fieldset #stateInput_chzn.chzn-container-single {
   ...
  .chzn-single span {
    line-height: 36px;
  }
}
```

dr-bizz (commented):
Review comment above

canac:
@dr-bizz A pro of your approach is that inputs and dropdowns are all 50px now, so I went with your approach. I adjusted the selector so that it applies to all dropdowns, not just the state one so that all dropdowns have a consistent height.

dr-bizz (approved):

PR #1185, opened 2025-03-19
2580-okta-modal-clean-up

rguinee:
The "one account to access" panel on the left is dense with information. During UAT we identified that removing the Okta disclosures at the bottom would make it less cluttered. We still have disclosures on the right panel for logging in/signing up, so we don't anticipate this causing confusion about Okta's role in account management. These disclosures will stay on mobile since the density of the panel reduces a lot on mobile.


src/common/components/registerAccountModal/registerAccountModal.tpl.html
```diff
@@ -59,37 +59,37 @@ <h2 translate>One account to access:</h2>
     </div>
   </div>
   <div class="inner-modal" ng-switch="$ctrl.state">
-    <sign-in-modal
```
canac, line 62: It looks like just whitespace was changed here. Can you revert those changes if they were unintentional?

src/common/components/registerAccountModal/registerAccountModal.tpl.html
```diff
@@ -42,7 +42,7 @@ <h2 translate>One account to access:</h2>
             <span class="benefit" translate>Events</span>
             <span translate>that empower you to journey together as you deepen your relationship with God and others.</span>
           </p>
-          <div class="okta-info">
+          <div class="okta-info u-hidden">
```
canac, line 45: Why are we hiding this element instead of removing it?

canac (commented):
Thanks for working on this. I tried it out and had a couple of questions.

src/common/components/registerAccountModal/registerAccountModal.tpl.html
rguinee: Ah yeah that's the format on save thing in the .vscode folder. I must have missed this one.

src/common/components/registerAccountModal/registerAccountModal.tpl.html
rguinee: Ha...spring break is getting the best of me. That div was supposed to be removed and the utility class was leftover from testing some things.

canac (changes requested):
Great work! We can clean up some of the styles now.

Can you remove the `.benefits .okta-info` rule?

Can you move the declarations from `&.mobile` nested inside `.okta-info` to directly into `.okta-info` since there aren't a mobile and non-mobile `.okta-info` elements anymore? I hope that makes sense.

canac (approved):
Looks great! 💯

canac:
Did you want to merge this? Before you do, can you add EP-2580 to the description? It would be good to change the title from the branch name to something like `[EP-2580] Okta modal cleanup`. Thanks for the descriptive branch name and commit messages!

rguinee:
@canac I do want to merge this...noticing a merge conflict in the Scss changes I made in the last commit. Looking into it now.

canac:
@rguinee FYI, the `height: 0.75rem` in `img` was accidentally reverted back to `height: 1rem`. I'll add a commit that switches it back.

rguinee:
Ah good catch, thank you!

PR #1187, opened 2025-03-25
EP-2568 - Ensure Okta redirects back to Give after logout

dr-bizz:
## Description
In this PR, I fix the redirection after signing out. Previously, the user would be kept on the Okta site when signing out. This fix ensures the user is redirected back to the Give site.

We are using the same link to sign out as we do on other applications like MPDX, so I know this will log them out of all instances of their Okta account on the browser before returning them to the Give site.

Jira Task: [EP-2568](https://jira.cru.org/browse/EP-2568)

### DIFF
--- a/src/common/services/session/session.service.js
+++ b/src/common/services/session/session.service.js
@@ -203,12 +203,13 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
         $window.sessionStorage.setItem(locationOnLogin, $window.location.href)
       }
 
-      return authClient.signOut({
-        postLogoutRedirectUri: redirectHome ? null : `${envService.read('oktaReferrer')}/sign-out.html`
-      }).catch(() => {
-        // If Okta sign out fails, redirect to Okta sign out page
-        $window.location.href = `${envService.read('oktaUrl')}/login/signout?fromURI=${envService.read('oktaReferrer')}`
-      })
+      // We no longer use the below code because it doesn't redirect the user back to the Give site
+      // Instead, it redirects to the Okta sign in page.
+      // return authClient.signOut({
+      //   postLogoutRedirectUri: redirectHome ? null : `${envService.read('oktaReferrer')}/sign-out.html`
+      // })
+
+      $window.location.href = `${envService.read('oktaUrl')}/login/signout?fromURI=${envService.read('oktaReferrer')}/sign-out.html`
     }
 
     return Observable.from($http({
--- a/src/common/services/session/session.service.js
+++ b/src/common/services/session/session.service.js
@@ -203,12 +203,7 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
         $window.sessionStorage.setItem(locationOnLogin, $window.location.href)
       }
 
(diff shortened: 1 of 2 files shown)


src/common/services/session/session.service.js
```diff
@@ -203,12 +203,13 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
...
-        postLogoutRedirectUri: redirectHome ? null : `${envService.read('oktaReferrer')}/sign-out.html`
-      }).catch(() => {
-        // If Okta sign out fails, redirect to Okta sign out page
-        $window.location.href = `${envService.read('oktaUrl')}/login/signout?fromURI=${envService.read('oktaReferrer')}`
-      })
+      // We no longer use the below code because it doesn't redirect the user back to the Give site
+      // Instead, it redirects to the Okta sign in page.
+      // return authClient.signOut({
+      //   postLogoutRedirectUri: redirectHome ? null : `${envService.read('oktaReferrer')}/sign-out.html`
+      // })
```
canac, line 210: I'd prefer not to leave commented out code in the codebase. Is there a reason this can't be removed?

src/common/services/session/session.service.spec.js
```diff
@@ -203,11 +203,8 @@ describe('session service', function () {
       it('makes a DELETE request to Cortex & sets postLogoutRedirectUri', done => {
         sessionService
           .signOut(false)
-          .subscribe((response) => {
-            expect(response.data).toEqual({})
-            expect(sessionService.authClient.signOut).toHaveBeenCalledWith({
-              postLogoutRedirectUri: `https://localhost.cru.org:9000/sign-out.html`
-            })
+          .subscribe(() => {
```
canac, line 206: Can you update the test names as well? Maybe something like "makes a DELETE request to Cortex redirects to Okta" and "should revoke all tokens and redirect to Okta".

canac (commented):
This is a major improvement for our users. Thanks for the fix!

src/common/services/session/session.service.js
dr-bizz: It's mainly so we don't try it again, but I can make it smaller

canac: That makes sense. Maybe we can delete the code, but leave a comment that says something like "Redirect instead of using authClient.signOut() because signOut leaves the user on the Okta dashboard"

src/common/services/session/session.service.spec.js
```diff
@@ -200,42 +200,23 @@ describe('session service', function () {
...
+            expect($window.location.href).toEqual(`${envService.read('oktaUrl')}/login/signout?fromURI=${envService.read('oktaReferrer')}/sign-out.html`)
             done()
           }, done)
         $httpBackend.flush()
       })
 
-      it('should still sign user out if error during signout', done => {
-        sessionService.authClient.signOut.mockRejectedValue()
-
+      it('should revoke all tokens & run redirect to Okta to sign out', done => {
```
canac, line 213:
```suggestion
      it('should revoke all tokens & redirect to Okta to sign out', done => {
```

canac (approved):
Thanks for the comment and wording tweaks!

PR #1190, opened 2025-03-25
[EP-2581] Only center register account modal

canac:
Instead of vertically centering all modals, only vertically center the register account modal.

EP-2581


dr-bizz (approved):
Thank you for doing this! This greatly improves the modal UX

PR #1191, opened 2025-03-25
EP-2583 - Prevent Okta from concealing the form.

dr-bizz:
## Description
Prevent Okta from concealing the form. Initially, I believed the form was not loading or that an error had occurred. However, it turned out that Okta was hiding the form instead.

Jira: [EP-2583](https://jira.cru.org/browse/EP-2583)
### Steps to replicate:
1. Open up the modal and complete the form, getting to step 3.
2. Close the modal and re-open it.
3. Step through to step 2, then go back to the previous step.
4. The form will no longer load.

### Changes
Added a style which ensures the form is displayed.

### DIFF
--- a/src/assets/scss/components/_oktaSigninWidget.scss
+++ b/src/assets/scss/components/_oktaSigninWidget.scss
@@ -49,6 +49,8 @@ sign-up-modal {
   }
 
   &.main-container {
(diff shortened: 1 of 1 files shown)


dr-bizz:
Closing as Cleb fixed the root cause in PR #1192

PR #1193, opened 2025-03-27
[HS-1332280] Expand designation editor input

canac:
Expand the designation editor input instead of letting it be set to `39px` by the `.form-control`.

The `import './designationEditor.scss'` wasn't doing anything. The `designationEditor.scss` styles weren't being included in the CSS bundle. I'm using a SCSS `@import` now instead.

Also, something is messed up with my staging account. I can't login with The Key locally and when I login with Okta in staging, I get a 403 Forbidden error when accessing the designation editor. I was only able to test by setting up local overrides pointing to the `give-web` `dist/` directory. Someone should probably test this in staging without overrides if possible.

https://secure.helpscout.net/conversation/2888678022/1332280

### DIFF
--- a/src/app/designationEditor/designationEditor.scss
+++ b/src/app/designationEditor/designationEditor.scss
@@ -1,3 +1,7 @@
 .designation-editor-paragraph-text{
   min-height: 90px;
 }
+
+.ta-editor {
+  height: auto;
+}
--- a/src/assets/scss/styles.scss
+++ b/src/assets/scss/styles.scss
@@ -31,3 +31,6 @@
 @import 'mobile';
 @import 'nav-cart';
 @import 'carousel';
+
+// TODO: Remove this manual import when we figure out why styles imported in components are not being added to the page
+@import '../../app/designationEditor/designationEditor';


wrandall22:
@jeff-clay do you happen to have any time/able to test designation editor in staging?

dr-bizz (commented):
Does this also mean that these aren't also working?

src/app/newsletterSubscription/newsletterModal/newsletter.modal.js
```
import './newsletterModal.scss'
```

src/app/designationEditor/textEditorModal/textEditor.modal.js
```
import './textEditor.scss'
```

If so, we should move all these SCSS files into a new folder `/assets/scss/app/` so it easy to see which page they belong and we have al the SCSS in one place.

dr-bizz:
I'm unable to test the UI as I can't open the esignation editor with my account

jeff-clay:
Are we talking about https://give-stage2.cru.org/0575182? I'm logged in and the edit button isn't popping up for me for some reason.

jeff-clay:
Screenshot 2025-03-27 at 11 30 13 AM

wrandall22 (approved):

canac:
> Can we make sure to updare staging with this change too

Good call

PR #1194, opened 2025-03-28
EP-2584, EP-2570, EP-2587 - Improving transition pages and loading screens

dr-bizz:
## Description
In this PR, I have improved the UX/UI of the `sign-out.html` and `okta-auth-callback.html` pages. This was mainly achieved by adding the `loading.scss` into the `/scss/assets` folder, as the loading component was not animating, but also updating the HTML.

## Jira Task
[EP-2584](https://jira.cru.org/browse/EP-2584)
[EP-2570](https://jira.cru.org/browse/EP-2570)
[EP-2587](https://jira.cru.org/browse/EP-2587)

## Changes
- Renamed `locationOnLogin` variable to `storedLocation` since we use it to store the previous location on login and logout.
- Added loading SCSS file to the `/assets/scss/components` folder.
- Added better UX/UI to the sign-out page
- Added better UX to the Okta authentication callback page.
- Ensure we show the loading component when registering for Cortexa and redirect to Okta, as this could take a few seconds.

## Test
- To test the signOut page, comment out lines `18` & `20`, then navigate to localhost.cru.org:9000/sign-out.html
- To test the oktaAuthCallback page, comment out lines `26-30`, then navigate to localhost.cru.org:9000/okta-auth-callback.html
- To test the loading screen, register for an account. Upon clicking "Continue" on the multiple MFA methods, it should show the loading screen.


src/app/signOut/signOut.tpl.html
wrandall22, line 1: We'll probably have to put this component onto the AEM page if we want to use it, or duplicate some things there.

dr-bizz: I know we currently have a sign-out page, but this component isn't linked to it. Can we link it to it without duplicating some thing?

What will be duplicated?

wrandall22: We could link it by creating an AEM component and dropping this in there, plus probably some bootstrapping. If we don't need any actual functionality on the page itself, we can just put the look/feel using existing AEM components onto the page.

dr-bizz: Would that mean the JS in the component runs, but we see a different UI? As we want to redirect the user to the previous page if they have one.

wrandall22: If we went the first route, we would see the AEM UI with this UI embedded into it.

src/app/oktaAuthCallback/oktaAuthCallback.tpl.html
```diff
@@ -10,7 +10,8 @@ <h3 class="text-center">
                   <div>
                     <loading></loading>
                     <br/>
-                    <p translate>{{$ctrl.noticeToUser}}</p>
+                    <span ng-if="!$ctrl.isReadyToRedirect" translate>Authenticating...</span>
```
canac, line 13: Based on other code in the repo, you can put the text inside `<loading>`. Once #1196 merges, there will be an animation with the loading bars.

src/assets/scss/components/_components.scss
```diff
@@ -7,5 +7,6 @@
 @import "./accountBenefits";
 @import "./failedVerificationModal";
 @import "./oktaSigninWidget";
+@import "./loading";
```
canac, line 10: This change and the rename of the loading stylesheet won't be necessary after #1196 merges.

src/app/oktaAuthCallback/oktaAuthCallback.tpl.html
dr-bizz: The other loaders have the words appear before or above the loading icon. Which might be needed elsewhere, but for this, I think we need to put the loading icon above the text, so this is the natural flow many other sites take (Showing the loading icon before the text), as the loading icon catches the eye of the user. Then the text gives the user more context.

dr-bizz:
> [Current logic on the Sign Out link in AEM](https://github.com/CruGlobal/giveaem/blob/main/ui.apps/src/main/content/jcr_root/apps/settings/wcm/designs/give/clientlibs/give-global/js/give-nav.js#L51-L63). ~Does this need to be removed in favor of pulling in the Angular component `sign-out`?~ In theory, we could put the redirect logic in the session service to make changes in AEM unnecessary.

The issue is that we need a page for Okta to redirect to after logout. This could be the homepage with a URL parameter. When the URL parameter is present, we implement the redirect logic we have for signout. However, this may confuse the user if they return to the homepage and click on a page, only to be redirected to another page without any warning message indicating that this is happening behind the scenes.

src/app/signOut/signOut.component.js
```diff
@@ -25,20 +26,15 @@ class SignOutController {
...
-    this.showRedirectingLoadingIcon = true
-    const locationToReturnUser = this.sessionService.getLocationOnLogin()
+    this.redirectToPreviousLocation = true
+    const locationToReturnUser = this.sessionService.getStoredLocation()
     if (locationToReturnUser) {
-      this.sessionService.removeLocationOnLogin()
+      this.sessionService.removeStoredLocation()
       this.$window.location.href = locationToReturnUser
     } else {
       this.redirectToHomepage()
```
canac, line 35: Wouldn't `this.redirectToPreviousLocation` need to be `false` in this branch? Maybe you can move `this.redirectToPreviousLocation = true` inside the `if`.

src/app/signOut/signOut.tpl.html
```diff
@@ -1,13 +1,24 @@
...
-        <p><translate>Successfully logged out.</translate></p>
-        <p><translate>Redirecting to your previous page...</translate></p>
-      </loading>
-      <button id="closeButtonTop" type="button" ng-click="$ctrl.closeRedirectingLoading()" class="btn btn-block btn-secondary" aria-label="{{'Close' | translate}}">
-        Close <span aria-hidden="true">×</span>
-      </button>
+  <div class="container">
+    <div class="row">
+      <div class="col-md-12 col-xs-12">
+        <div class="panel sign-in-panel">
```
canac, line 5: Is the `sign-in-panel` class a copy-paste error or just a name that's not completely accurate anymore?

src/common/services/session/session.service.js
```diff
@@ -276,19 +276,19 @@ const session = /* @ngInject */ function ($cookies, $rootScope, $http, $timeout,
     return envService.read('oktaUrl')
   }
 
-  function removeLocationOnLogin () {
-    $window.sessionStorage.removeItem(locationOnLogin)
+  function removeStoredLocation () {
```
canac, line 279: Thanks for renaming these! What do you think about `getRedirectLocation` and `clearRedirectLocation`? I that communicates a little better the intent of what the stored location is used for.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -742,6 +742,8 @@ class SignUpModalController {
   }
 
   saveDonorDetails () {
+    this.isLoading = true
+    this.oktaSignInWidget.remove()
```
canac, line 746: Do we need to clear the Okta listeners we do in `$onDestroy`?

src/app/oktaAuthCallback/oktaAuthCallback.component.js
```diff
@@ -37,7 +37,7 @@ class OktaAuthCallbackController {
   }
 
   initializeVariables () {
-    this.noticeToUser = 'Authenticating...'
+    this.isReadyToRedirect = false
     this.isLoading = true
```
canac, line 41: I don't see `this.isLoading` being read anywhere. Am I missing something?

canac (commented):
I see a few things that may be able to be improved. But nothing seems broken, so I'm happy to merge this into staging for our UAT session before this gets fully merged.

src/app/signOut/signOut.component.js
dr-bizz: Good catch

src/app/oktaAuthCallback/oktaAuthCallback.component.js
dr-bizz: It is being used on the HTML

canac: I don't see `isLoading` in `src/app/oktaAuthCallback/oktaAuthCallback.tpl.html`, just `isReadyToRedirect`.

dr-bizz: Sorry, you're correct. I was looking at the Sign Out component.

src/app/signOut/signOut.component.js
dr-bizz: It should be fixed. Let me check

dr-bizz: Yes, it is fixed. `this.redirectToPreviousLocation` is only set to `TRUE` when  `locationToReturnUser` is truthy.

canac: Sorry, I clicked the wrong thing and I was looking at the old version of that method. Looks great now!

canac (approved):
This all looks good to me. Was updating the widget to fix something in particular?

dr-bizz:
No, it was just that I saw there was an update

PR #1196, opened 2025-03-28
[EP-2585] Upgrade style-loader to v2

canac:
## Description

When #1171 upgraded `mini-css-extract-plugin`, it neglected to also update `style-loader` to v2, as recommended in the [release notes](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/CHANGELOG.md#-notice). The result was that stylesheets imported from JS files were not being injected into the DOM.

## Testing

* On `master`
  * Go to https://localhost.cru.org:9000/cart.html
  * See that the loading component for "Loading your gifts..." renders as unstyled text.
* On this branch
  * Go to https://localhost.cru.org:9000/cart.html
  * See that the loading component is correctly styled. The text should be centered and have animated bars.

EP-2585

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "sass": "^1.49.0",
     "sass-loader": "^10",
     "standard": "^16.0.4",
-    "style-loader": "^1.0.0",
+    "style-loader": "^2.0.0",
     "stylelint": "^16.14.1",
     "stylelint-config-idiomatic-order": "^10.0.0",
     "stylelint-config-standard": "^37.0.0",
--- a/src/app/designationEditor/designationEditor.scss
+++ b/src/app/designationEditor/designationEditor.scss
@@ -1,7 +1,3 @@
 .designation-editor-paragraph-text{
   min-height: 90px;
 }
-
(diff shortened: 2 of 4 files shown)


dr-bizz (approved):
Thank you for fixing this. I guess this means we can unable some of the changes done on PR #1193

wrandall22 (approved):

canac:
@wrandall22 I reverted the entire PR #1193 because it is no longer necessary with this fix. I'm still trying to get login access, but I tested it locally with local overrides.

```sh
mkdir -p 'overrides/localhost.cru.org:9000/bin/crugive'
echo "{}" > 'overrides/localhost.cru.org:9000/bin/crugive/designation.html?designationNumber=0864190'
echo "[]" > 'overrides/localhost.cru.org:9000/bin/crugive/image.html?designationNumber=0864190'
```

Now go to https://localhost.cru.org:9000/designation-editor.html?d=0864190. The editor should be full-height even with the explicit height style removed.

wrandall22 (approved):

PR #1198, opened 2025-04-09
2362 okta optional mfa click

dr-bizz:
## Description
Currently, Okta requires us to show you the MFA options during sign-up. Jason Buckner is trying to remove them, but in the meantime, I have made it so we can skip past this step, so the user doesn't get caught in this step while signing up.

We skip the Optional MFA enroll step by clicking the "Continue" button via JS and then hiding the MFA enroll HTML to not confuse the user.

### Testing
- Sign up as usual, entering your verification code.
- When you click "Submit", a loading screen will show, and then you will be taken to Okta to log in.

### DIFF
--- a/src/common/components/signUpModal/signUpModal.component.spec.js
+++ b/src/common/components/signUpModal/signUpModal.component.spec.js
@@ -909,6 +909,7 @@ describe('signUpForm', function () {
         jest.spyOn($ctrl, 'injectErrorMessages').mockImplementation(() => {});
         jest.spyOn($ctrl, 'injectBackButton').mockImplementation(() => {});
         jest.spyOn($ctrl, 'showVerificationCodeField').mockImplementation(() => {});
+        jest.spyOn($ctrl, 'skipOptionalMFAEnrollment').mockImplementation(() => {});
 
         $ctrl.$onInit()
       })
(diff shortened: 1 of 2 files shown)


src/common/components/signUpModal/signUpModal.component.spec.js
```diff
@@ -909,6 +909,7 @@ describe('signUpForm', function () {
         jest.spyOn($ctrl, 'injectErrorMessages').mockImplementation(() => {});
         jest.spyOn($ctrl, 'injectBackButton').mockImplementation(() => {});
         jest.spyOn($ctrl, 'showVerificationCodeField').mockImplementation(() => {});
+        jest.spyOn($ctrl, 'skipOptionalMFAEnrollment').mockImplementation(() => {});
```
wrandall22, line 912: Semicolons should be removed.

dr-bizz: When running `yarn lint:write`, it passes without making any edits or removing the semicolons.

I think we should establish a set rule, as we have some lines with semicolons and some without.

wrandall22: The rules are there, but specs are skipped from formatting.

canac (approved):
I tested and this seems to work. Thanks for the improvement!

src/common/components/signUpModal/signUpModal.component.spec.js
dr-bizz: @wrandall22 For now, prettier gives me an error when I remove the semicolon, so I will leave it there until we can improve the prettier and lint settings in T3.

PR #1199, opened 2025-04-10
Update the Build process and reduce repo size.

dr-bizz:
## Description
In this PR, I am cleaning up some files to prevent bloat on this repository.

**Removed labels**
If you ran `yarn build`, it was adding 60 label files to the assets folder, which we had previously removed. 
I updated the webpack to no longer copy these files during build.

**Removed unused images**
I removed 23 images that we no longer needed. I initially added them as I didn't know which ones we would need.


webpack.config.js
```diff
@@ -33,11 +33,7 @@ const brandedComponents = [
   'assets/scss/branded-checkout.scss'
 ]
 
-// Okta sign-in widget assets files
-const oktaSignInLabelsDestinations = [
-  'dist/assets/okta-sign-in/labels',
-  'src/assets/okta-sign-in/labels'
-]
+// Okta sign-in widget font files
```
canac, line 36: Can we delete `src/assets/okta-sign-in/labels` now?

src/assets/okta-sign-in/css/okta-sign-in.min.css
canac, line 1: Why did this file change? It is because the widget got updated in another PR?

webpack.config.js
canac, line 36: Can we delete `src/assets/okta-sign-in/labels` now?

canac (commented):
Thanks for going through this! Are we using these icons?

* `src/assets/okta-sign-in/img/icons/identifier/user-icon.svg`
* `src/assets/okta-sign-in/img/icons/login/tooltip_close.png`
* `src/assets/okta-sign-in/img/icons/mfa/*`

Should we ask Mike before removing the labels so that users with a browser default language set to something besides English can see them?

How it currently is, if the user has their default language set to Spanish, the field labels and country list will be translated into Spanish. After this change, all the labels will be in English. Maybe that's OK since the rest of the Give Site only supports English and Spanish, but it would be good to ask the stakeholders.

canac:
Is there a reason that anything in `src/assets/okta-sign-in` is committed to the repo? Could we delete everything and add that folder to `.gitignore`? The webpack copy plugin will still keep that directory in sync with `node_modules/@okta/okta-signin-widget`. So the directory will still be there on the filesystem but not in the git repo.

dr-bizz:
So upon further investigation. We could need the labels if we plug in the user's language option. It could grab the browser language settings. So I might add the labels back.

That is a good point, we could ignore that folder as we on't need to track it

.gitignore
```diff
@@ -12,3 +12,8 @@ stats.json
 
 # asdf
 .tool-versions
+
+
+# Okta sign in  assets
+src/assets/okta-sign-in/font
+src/assets/okta-sign-in/labels
```
canac, line 19: Is there a reason to keep the minified CSS in version control?

```suggestion
# Okta sign in assets
src/assets/okta-sign-in
```

dr-bizz: Yes, I tried to remove it, but since we included it in the files, it needs to exist so a build can be completed.

canac (approved):
Thanks for cleaning this up!

PR #1200, opened 2025-04-21
[no-Jira] Replace the update-staging workflow with merge-bot

canac:
The `update-staging` workflow successfully merges branches into staging, but it doesn't trigger a deployment because workflows can't trigger other workflows. This PR removes that workflow and enables the merge bot for this repo.


canac:
> I'm not sure how I could test this but it seems good to me.

@caleballdrin When I did this with the ERT in https://github.com/CruGlobal/conf-registration-web/pull/875, we had to merge this PR first. Then the next PR that we added the `On Staging` label to got automatically merged into staging. This is the exact same change that worked with the ERT, so I'm pretty confident it's right.

PR #1201, opened 2025-04-23
Sets overflow-y of .modal to scroll

rguinee:
Adding overflow-y attribute and setting it to `scroll` will override the current value `hidden` for overflow allowing the user to scroll to the "Add Payment Method" button.

### DIFF
--- a/src/assets/scss/_gift-config.scss
+++ b/src/assets/scss/_gift-config.scss
@@ -146,6 +146,7 @@ label.btn.btn-default-form.active {
 .modal {
   display: none;
   overflow: hidden;
+  overflow-y: scroll;
   position: fixed;
   right: 0;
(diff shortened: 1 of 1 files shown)


wrandall22 (approved):
This was interesting to test. Locally, I found that the way I typically test branded checkout brought in CSS that is not brought in for BCO on production. In stage, I found that going back no longer keeps the payment method saved, but gives an empty form again. I was able to determine this fixes the issue without changing standard checkout.

rguinee:
Will the current flow on stage where it does not save the payment method change the intended behavior in production after we push?

wrandall22:
I think it is probably a side effect of something that was non-intentional. However, I don't think it is a bad flow necessarily.

PR #1202, opened 2025-04-23
[EP-2593] Fix post-donation account creation stuck loading

canac:
When creating an account from the thank you page, the modal was getting stuck in a loading state after submitting the email verification code. This is because `redirect-to-okta-for-login` was not being passed to `<sign-up-modal>` from `sessionModal`. I fixed this by moving the logic in `redirectToOktaForLogin` into `sessionService.signIn` and adding `redirectToOktaForLogin` to the `signUpModal` controller to guarantee that it will always be available.

I also tweaked the padding around the account benefits modal to be more consistent and renamed `accountBenefits.scss` to `accountBenefitsModal.scss` since all of its styles are for `account-benefits-modal`.

Before (less padding on the top and extra padding on the bottom):

Screenshot 2025-04-23 at 4 41 31 PM

After (35px padding on all four sides):

Screenshot 2025-04-23 at 4 43 57 PM

Ryan requested that this not be merged into staging until after his user study concludes at 1:30pm ET tomorrow (Thursday, April 24).

EP-2593


src/common/services/session/session.service.spec.js
```diff
@@ -302,6 +303,35 @@ describe('session service', function () {
...
+        accessToken: {
+          accessToken: 'accessToken'
+        }
+      }))
+
+      const $injector = {}
+      jest.spyOn(angular, 'injector').mockReturnValue($injector)
+
+      const giveSignInSuccessCallback = jest.fn()
+      window.document.body.addEventListener('giveSignInSuccess', giveSignInSuccessCallback, { once: true })
```
dr-bizz, line 320: Is it best to make `giveSignInSuccess` a variable since we use it in a few places?

dr-bizz (commented):
Looking great! Thank you for making this change. 

I would like to test it in staging before we merge it into the `2362-okta` branch.

src/common/services/session/session.service.spec.js
dr-bizz: You're right, we no longer use it, so it might be useful to remove `LoginOktaOnlyEvent` in a later PR

canac: #1203

canac:
I just tested these flows in staging and everything is working correctly ✅

* Create an account from the Sign In button
* Anonymously submit a gift, then create an account from the Thank You page

dr-bizz (approved):
Thank you for making these changes! They look great.

One piece of feedback I was given is that it's not obvious the user will be redirected to Okta as Okta isn't mentioned anywhere when signing up from the Thank you page. how easy would it to be to add the Okta blurb?

Feel free to merge this PR and create a new one for this request.

Screenshot 2025-05-01 at 1 23 42 PM

PR #1203, opened 2025-04-24
[EP-2362] Remove unused LoginOktaOnlyEvent

canac:
Remove unused `LoginOktaOnlyEvent`

EP-2362


dr-bizz (approved):
Thank you for doing this. We used to use `LoginOktaOnlyEvent` in src/common/services/session/sessionHandleOktaRedirect.service.js. But that was removed some time ago, and we didn't realised we also needed to remove this.

PR #1205, opened 2025-05-01
EP-2598 - Add option to send email address to Okta on login

dr-bizz:
## Description
Add an option to allow an email address to be sent to Okta during login.

The reason for this is to eliminate a step in the sign-up process where we require users to re-enter their email address. We found that people were entering their email incorrectly and getting stuck, as the password field does not provide much guidance when an incorrect email or password is entered.


canac (approved):
This is amazing! It definitely improves the UX. I love that it didn't take much code to implement.

PR #1206, opened 2025-05-02
EP-2600 - Add repeat passcode field to sign up

dr-bizz:
## Description
In this PR, I have added a repeat password/passcode field to the sign-up process. The reason for this change is that 1. DSG has been advocating for it, and we found that many people were forgetting their password when they accessed Okta.

I was initially opposed to adding the repeat passcode as it introduces another field, and I do not want to save the password anywhere. However, I have discovered a method to implement it without storing the password. If we later find the repeat passcode field to be annoying, we can remove it.

### DIFF
--- a/src/common/components/signUpModal/signUpModal.component.js
+++ b/src/common/components/signUpModal/signUpModal.component.js
@@ -224,7 +224,7 @@ class SignUpModalController {
     ]
   }
 
-  getStep2Fields (schema) {
+  getStep2Fields () {
     // Retain the values entered by the user when navigating between steps.
     // Pre-populate the form fields with existing user details.
     return [
--- a/src/common/components/signUpModal/signUpModal.component.js
+++ b/src/common/components/signUpModal/signUpModal.component.js
@@ -174,16 +174,16 @@ class SignUpModalController {
   }
 
   getSteps (schema) {
-    const passwordInput = schema.find(field => field.name === 'credentials.passcode')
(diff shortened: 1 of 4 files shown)


dr-bizz:
Not for myself, I need to adjust the order radius on the input.

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -409,8 +414,21 @@ class SignUpModalController {
...
+    if (postData.credentials.passcode !== postData.passcodeVerification.repeatPasscode) {
+      errors.push({
+        errorSummary: this.translations.passwordDoesNotMatch,
+        property: 'passcodeVerification.repeatPasscode'
+      })
+    }
+    if (errors.length) {
+      this.injectErrorMessages(errors)
+      return
+    }
```
canac, line 427:
```suggestion
    if (postData.credentials.passcode !== postData.passcodeVerification.repeatPasscode) {
      const errors = [{
        errorSummary: this.translations.passwordDoesNotMatch,
        property: 'passcodeVerification.repeatPasscode'
      }]
      this.injectErrorMessages(errors)
      return
    }
```

canac (commented):
Looks great! I know you said you're going to fix the border radius on the repeat password input and its error message.

dr-bizz:
This is 98% there, but I'm not going to merge it as we are rolling live without it. However, it's part of Phase 2, so it could be finished later and deployed.

PR #1207, opened 2025-05-06
[EP-2362] Add label to loading indicator before redirecting

canac:
After the user submits the verification code, a loading indicator appears for a few seconds before redirecting to Okta. This PR adds a label to that loading indicator so that the user is aware of what is loading and is less confused when they are redirected.

@rguinee Do you have thoughts on the message I used?

### DIFF
--- a/src/assets/scss/components/_signUpModal.scss
+++ b/src/assets/scss/components/_signUpModal.scss
@@ -35,6 +35,10 @@ sign-up-modal {
 
   .okta-signup-loading {
     padding: 153px 0;
+
+    .loading-message {
+      padding-bottom: 0.5rem;
+    }
   }
 
   form {
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -394,7 +394,8 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     OKTA_FIRST_NAME_FIELD: 'first name',
     OKTA_LAST_NAME_FIELD: 'last name',
     OKTA_EMAIL_FIELD: 'email',
-    OKTA_PASSWORD_FIELD: 'password'
+    OKTA_PASSWORD_FIELD: 'password',
+    OKTA_REGISTERING: 'Creating your Okta account and preparing to redirect you to Okta to log in...'
   })
 
   $translateProvider.translations('es', {
@@ -617,7 +618,8 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     OKTA_FIRST_NAME_FIELD: 'first name',
     OKTA_LAST_NAME_FIELD: 'last name',
     OKTA_EMAIL_FIELD: 'email',
-    OKTA_PASSWORD_FIELD: 'password'
+    OKTA_PASSWORD_FIELD: 'password',
+    OKTA_REGISTERING: 'Creating your Okta account and preparing to redirect you to Okta to log in...'
   })
   $translateProvider.preferredLanguage('en')
 }
--- a/src/common/components/signUpModal/signUpModal.tpl.html
+++ b/src/common/components/signUpModal/signUpModal.tpl.html
@@ -54,8 +54,10 @@
         />
       </div>
     </div>
-    
-    <loading ng-if="$ctrl.isLoading" class="okta-signup-loading"></loading>
+
+    <loading ng-if="$ctrl.isLoading" class="okta-signup-loading">
+      <div ng-if="$ctrl.currentStep === 5" class="loading-message" translate>OKTA_REGISTERING</div>
+    </loading>
 
     <div id="osw-container"></div>
   </div>


rguinee:
Ideally the tone of the message is clear enough to be understood across a
spectrum of technical proficiency. I think this message leans toward a
greater level of technical proficiency and if we aim the other direction
towards little to none we’ll still be understood by technically adept users.

There are multiple layers to this particular part of the UX that I hoped we
would address:
- telling the user what’s happening now and giving them enough time to know
what’s happening next, making it clear that they aren’t about to re-enter
their info as part of account creation—they are signing in (this PR
achieves this although I wonder how quickly it moves past the message if
the response from Okta is fast)
- reassuring the user that their account registration was successfully
verified (including some easily recognizable, universally understood,
visual like a check mark, welcome message, etc.) I realize we tend to just
tell people if something goes wrong, no news good news, but this UX is
fragile because of the length of the process and the redirection/mismatched
branding. It’s good to reassure often.
- familiarizing the user with the appearance of Okta before they’re
redirected (Figma shows how this could be done on Desktop, mobile is
challenging)
- reassuring the user, again, that Okta is secure and we are intentionally
leading them to Okta (gimmicky, but the green shield or lock icon to
reinforce security)

dr-bizz:
@rguinee What you are mentioning isn't within the scope of this project. I like what you have created, but I feel some of this is phase 2 work. I don't want to delay the launch of this project any more than we have to.

> - telling the user what’s happening now and giving them enough time to know
> what’s happening next, making it clear that they aren’t about to re-enter
> their info as part of account creation—they are signing in (this PR
> achieves this although I wonder how quickly it moves past the message if
> the response from Okta is fast)

If we add the Okta logo to the loading page and redirect the page after 10 seconds with the current text, this will achieve this.

> - reassuring the user that their account registration was successfully
> verified (including some easily recognizable, universally understood,
> visual like a check mark, welcome message, etc.) I realize we tend to just
> tell people if something goes wrong, no news good news, but this UX is
> fragile because of the length of the process and the redirection/mismatched
> branding. It’s good to reassure often.

I'm good with us adding a message stating their account has been created and verified. @canac what do you think to changing it to the following text?

✅ Success!
Your account has been created and verified. You’ll be redirected to Okta to sign in [count down].
If you’re not redirected within 10 seconds, [click here to continue].


> - familiarizing the user with the appearance of Okta before they’re
> redirected (Figma shows how this could be done on Desktop, mobile is
> challenging)

This is out of scope.


> - reassuring the user, again, that Okta is secure and we are intentionally
> leading them to Okta (gimmicky, but the green shield or lock icon to
> reinforce security)

Done with the success message above.

dr-bizz (changes requested):
This is looking good, but could we add some of the edits I made in the commit above?

rguinee:
Sounds good @dr-bizz and @canac, thanks for considering the direction and finding a good compromise for launch!

src/common/components/loading/loading.component.js
```diff
@@ -36,6 +36,7 @@ class LoadingController {
   constructor () {
     this.type = 'block'
     this.inline = false
+    this.iconFirst = false
```
canac, line 39: Does this give the binding a default value? I'm not necessarily opposed, but I don't think it's necessary because we only add the `.icon-first` class if `$ctrl.iconFirst === 'true'`. This check evaluates the same whether or not `iconFirst` has a default value.

src/common/components/signUpModal/signUpModal.tpl.html
```diff
@@ -55,9 +55,32 @@ <h3 ng-if="$ctrl.isInsideAnotherModal" class="signup-header" translate>Create Yo
...
-    <loading ng-if="$ctrl.isLoading" class="okta-signup-loading">
-      <div ng-if="$ctrl.currentStep === 5" class="loading-message" translate>OKTA_REGISTERING</div>
-    </loading>
+    <loading ng-if="$ctrl.isLoading && $ctrl.currentStep !== 5" class="okta-signup-loading"></loading>
+
+    <div ng-if="$ctrl.currentStep === 5" class="welcome-message">
+      <loading class="redirect-loading" inline="true" icon-first="true" iconFirst="true">
+        <div class="loading-message" translate>Redirecting in {{$ctrl.redirectToOktaCountdown}}</div>
+      </loading>
+      <h2 translate>Welcome, {{$ctrl.scope.firstName}}</h2>
```
canac, line 64: I'm not sure that this is going to work for non-English languages. I think we'll need to put `Welcome, {{firstName}}` in a label in `app.config.js` and pass in the variables as described here: https://angular-translate.github.io/docs/#/guide/06_variable-replacement

canac (commented):
@dr-bizz Thanks for the help on this! Just leaving a few comments.

I'm realizing that anywhere we do `<p translate>Some human-readable message<p>` instead of `<p>TRANSLATION_KEY</p>`, it won't be translated for non-English languages. We'll need to come back and make sure all of our labels are doing it the second way if/when we want Spanish translations.

src/common/components/loading/loading.component.js
dr-bizz: I set it as false, so developers know it is a boolean and can quickly understand how to set the icon first.

canac: Got it. Since `@` bindings always create strings, do you think it would be better to set the default value to the string `"false"`? Or make it a `<` binding?

dr-bizz: That is a good point. Let's use `<`

canac: I went ahead switched it to `<`

src/common/components/signUpModal/signUpModal.tpl.html
canac: I went ahead and fixed this

src/common/components/signUpModal/signUpModal.component.js
```diff
@@ -508,14 +513,31 @@ class SignUpModalController {
...
-      const skipMFAButton = mfaOptions.querySelector('.button.skip-all')
-      skipMFAButton?.click()
+
+      this.oktaRedirectCountdown = 10
+      this.oktaRedirectInterval = this.$interval(() => {
+        --this.oktaRedirectCountdown
+        if (this.oktaRedirectCountdown === 0) {
+          this.completeRegistration()
+        }
+      }, 1000, 10)
```
canac, line 525: The third argument `10` here makes the interval automatically cancel after 10 times.

https://docs.angularjs.org/api/ngMock/service/$interval#$interval-arguments

src/common/app.config.js
```diff
@@ -620,8 +624,12 @@ export const appConfig = /* @ngInject */ function (envServiceProvider, $compileP
     OKTA_LAST_NAME_FIELD: 'last name',
     OKTA_EMAIL_FIELD: 'email',
     OKTA_PASSWORD_FIELD: 'password',
+    OKTA_REDIRECTING_IN: 'Redirecting in {{seconds}}...',
+    OKTA_REDIRECTING_NOW: 'Redirecting...',
+    OKTA_WELCOME: 'Welcome, {{name}}!',
     OKTA_ACCOUNT_CREATED: 'Your account with Cru was successfully created and verified. <br/><strong>Now let’s sign you in.</strong>',
-    OKTA_REDIRECT_EXPLANATION: 'For your security, you are being redirected to Okta to sign in with your new account.Okta is our secure login provider, helping keep your account safe. <strong>You’ll briefly visit the domain below, then we’ll bring you back after you sign in.</strong>'
+    OKTA_REDIRECT_EXPLANATION: 'For your security, you are being redirected to Okta to sign in with your new account. Okta is our secure login provider, helping keep your account safe. <strong>You’ll briefly visit the domain below, then we’ll bring you back after you sign in.</strong>',
+    OKTA_REDIRECT_NOW: 'Take me to Okta now'
```
dr-bizz, line 632: I think at some point we should have these translated into Spanish.

src/assets/scss/components/_signUpModal.scss
```diff
@@ -92,6 +92,14 @@ sign-up-modal {
     }
   }
 
+  .okta-url {
+    padding: 0.5rem;
+    border: 1px solid #b3b3b3;
```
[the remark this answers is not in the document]
canac: It wasn't running on SCSS files in directories of `src/assets/scss`. I fixed that.

canac:
@dr-bizz I had to fix some style lint issues in unrelated components. Do you want to give this another quick look over?

dr-bizz (approved):
Looks great to me

PR #1208, opened 2025-05-06
[EP-2362] Make the color of Okta dropdowns match text inputs

canac:
This PR makes the color of Okta dropdowns match the color of text inputs. I also moved the Okta CSS include into the SCSS file as an `@import` to ensure that our rules with the same specificity as Okta's (like the ones I added) will override Okta's rules.


dr-bizz (changes requested):
Thank you for updating the colour of the dropdowns.

They could have always been like this, but can you make the dropdown arrows more vertically aligned with the input

Screenshot 2025-05-07 at 8 46 43 AM

canac:
Good catch, thanks for noticing that. I used screenshots to verify that the dropdowns are in the same location as before, just with a different color now.

dr-bizz (approved):
Great work! Thank you for fixing this

PR #1210, opened 2025-05-14
[HS-1354646] Remove Grouping and Sorting unless Featured

caleballdrin:
https://secure.helpscout.net/conversation/2924841982/1354646?viewId=7669074

Problem:
Results are being grouped and ordered by date on a default search. This should only happen when in the featured section.

Changes:
* Check if the facet === 'featured' before grouping and ordering the results

### DIFF
--- a/src/app/searchResults/searchResults.tpl.html
+++ b/src/app/searchResults/searchResults.tpl.html
@@ -113,7 +113,10 @@
                     </ul>
                   </section>
 
-                  <div ng-repeat-start="(k,hits) in hits | orderBy:$ctrl.featuredGroupBy:(facet === 'featured' && $ctrl.featuredGroupBy === 'startMonth') | groupBy:$ctrl.featuredGroupBy">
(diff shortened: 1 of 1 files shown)


dr-bizz (commented):
This seems good. When testing on Staging I was seeing some ministries on the ministry search.

Screenshot 2025-05-16 at 10 15 26 AM

caleballdrin:
@wrandall22 good to know. They don't show up on production when searching for 'smith', so I assumed it was an issue with the data.

caleballdrin:
@dr-bizz see this comment to answer your question
https://github.com/CruGlobal/give-web/pull/1210#issuecomment-2887097087

wrandall22:
> To clarify, you are saying that it's not bad that people were shown on the ministry filter?

It is likely confusing, but it is technically correct. There is no way for us to filter those people out at this point in time.

dr-bizz:
I'm good with approving this then if that is the case. Best to wait until Monday to deploy.

dr-bizz (approved):

PR #1212, opened 2025-05-27
[EP-2603] Use register account modal on thank you page

canac:
## Description

On the thank you page, a modal pops up on page load with a "Register Your Account" button. This PR makes that button open the register account modal opened to the sign up form instead of a minimal sign up form.

This PR also does some cleanup in the logic in `accountBenefits.component.js`, correctly passes in `lastPurchaseId` to the register account modal, and removes some old Okta redirect logic in `accountBenefitsModal.component.js`.

EP-2603


dr-bizz (commented):
The code is looking good, but I haven't been able to test it as I can't get my local to work currently and staging doesn't show your edits.

dr-bizz:
Fixed issue on staging branch which prevented deployment

dr-bizz (approved):
Awesome work on this

PR #1220, opened 2025-06-23
switching functions so Prod can sign out fully

dr-bizz:
## Description
As AEM currently calls `downgradeToGuest` on sign out, we need to temporarily change the sign out function names so users get fully logged out of Okta.

`downgradeToGuest` -> `downgradeToGuestOld`
`signOut` -> `downgradeToGuest`
New function `signOut` that calls `downgradeToGuest`.

I had to update certain files to fix tests and ensure the proper functions were called.

### DIFF
--- a/src/app/signIn/signIn.component.js
+++ b/src/app/signIn/signIn.component.js
@@ -34,7 +34,7 @@ class SignInController {
   }
 
   checkoutAsGuest () {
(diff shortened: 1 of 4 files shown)


canac (approved):
This looks right

canac:
In the future, I believe we can remove the `redirectHome` argument since it is always `true` and delete the branches that check `if (!redirectHome)` since they will never run.

PR #1223, opened 2025-06-24
[no-Jira] Improve logging when sign-in fails

canac:
Since `error` is an object, it shows up in Rollbar (and probably in the UI) as `[object Object]`. This PR changes it to always show `unknownErrorMessage` to the user and log the error object in Rollbar. Unless we know how to extract a human-readable error message from `error`, this is going to be a better UX than showing them `[object Object]`.

### DIFF
--- a/src/app/oktaAuthCallback/oktaAuthCallback.component.js
+++ b/src/app/oktaAuthCallback/oktaAuthCallback.component.js
@@ -119,8 +119,8 @@ class OktaAuthCallbackController {
   }
 
   onSignInFailure (error) {
-    const errorMessage = error || unknownErrorMessage
-    this.$log.error(errorMessage)
+    const errorMessage = error instanceof Error ? error.message : unknownErrorMessage
+    this.$log.error(errorMessage, error)
     this.sessionService.clearRedirectLocation()
     this.errorMessage = errorMessage
   }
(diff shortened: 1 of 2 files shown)


src/app/oktaAuthCallback/oktaAuthCallback.component.js
```diff
@@ -119,10 +119,9 @@ class OktaAuthCallbackController {
   }
 
   onSignInFailure (error) {
-    const errorMessage = error || unknownErrorMessage
-    this.$log.error(errorMessage)
+    this.$log.error('Failed to authenticate user when redirecting from Okta', error)
     this.sessionService.clearRedirectLocation()
-    this.errorMessage = errorMessage
+    this.errorMessage = unknownErrorMessage
```
dr-bizz, line 124: Is it worth also adding the error from Okta, or is there no good data in it?

dr-bizz (commented):
Thank you for adding this! Great work spotting this

src/app/oktaAuthCallback/oktaAuthCallback.spec.js
```diff
@@ -73,26 +73,26 @@ describe('oktaAuthCallback', function () {
 
 
   describe('onSignInFailure()', () => {
-    it('should log error message', () => {
+    it('should log error message when error is an Error instance', () => {
```
canac, line 76: I removed two tests because I consolidated the tests and gave them more descriptive names

PR #1226, opened 2025-07-08
[no-Jira] Set up prettier and husky

canac:
Set up prettier and husky for precommit hooks. When reviewing, go commit by commit and skip "Apply automatic fixes". prettier caught some malformed HTML and SCSS, which is a nice perk!


dr-bizz (approved):
Thank you for adding this

PR #1227, opened 2025-07-15
 HS-1390414-CSS-updates

wjames111:
**Description**
This PR is for [HelpScout ticket 1390414](https://secure.helpscout.net/conversation/2976373523/1390414?viewId=8086747)
It's intended purpose is to make branded-checkout more friendly for WordPress environments by doing the following:

- Set `display` property to `block` on `branded-checkout` element
- Namespace overly generic selectors such as loading1, loading2, etc.
- Ensure selectors are nested within an applicable parent class such as branded-checkout. Adding specificity and reducing chances of generic styles incidentally being applied to branded checkout
- Reduce fade animations to a single animation
- Fix content shifts (This one might need some extra attention as `ng-if` edits were done to `product-config-form` which appear to be used by both branded and give)

Issues:

- For development branded-checkout styles need to be manually added to the branded-checkout component as the dev environment works differently then prod.
- It appears most of the content shifts were due to the having multiple different loading classes within product config component and elsewhere instead of wrapping all of branded checkout in them.
- Additionally it appears branded-checkout styles are already nested within branded-checkout element so not a whole lot was needed here unless I misread the request.

### DIFF
--- a/src/assets/scss/branded-checkout.scss
+++ b/src/assets/scss/branded-checkout.scss
@@ -25,6 +25,7 @@ branded-checkout{
     display: block;
   }
 
+  // Prevents layout shifts when loading
   &:has(loading) {
     position: relative;
     display: block;


src/assets/scss/branded-checkout.scss
```diff
@@ -1,6 +1,6 @@
 @use 'sass:math';
 
-branded-checkout{
+.branded-checkout{
```
wrandall22, line 3: Not quite sure what benefit this has. It changes the scope from the Angular element to the div directly underneath it.

src/app/productConfig/productConfigForm/productConfigForm.tpl.html
```diff
@@ -21,14 +21,6 @@ <h4 id="product-config-header" class="give-gift-header" translate>
     </div>
   </div>
 
-  <loading ng-if="$ctrl.loading">
-    <translate>{{'LOADING_GIFT_DETAILS'}}</translate>
-  </loading>
-
-  <div ng-if="$ctrl.errorLoading" class="alert alert-danger" role="alert">
-    <p translate='LOADING_ERROR' translate-values="{loadData: '$ctrl.loadData()'}" translate-compile></p>
-  </div>
```
wrandall22, line 30: Taking this out of here will remove the loading indicator from the standard checkout

src/app/branded/step-1/branded-checkout-step-1.tpl.html
```diff
@@ -3,109 +3,114 @@
...
+  </loading>
 <div ng-if="!$ctrl.loadingProductConfig && !$ctrl.errorLoadingProductConfig">
   <div class="panel">
-    <div class="panel-body">
-      <h3 ng-if="!$ctrl.useV3" class="panel-name" translate>{{'YOUR_INFORMATION'}}</h3>
-      <contact-info
+    <div class="panel-body give-loading-overlay-parent">
+      <loading ng-if="$ctrl.loadingProductConfig">
+        <translate>Loading gift details...</translate>
+      </loading>
```
wrandall22, line 14: Moving this within a check for `!$ctrl.loadingProductConfig` seems erroneous.

src/app/branded/step-1/branded-checkout-step-1.tpl.html
wrandall22, line 1: The indentions got messed up in the process of refactoring this file

clonge:
@wjames111 Not without research / knowing the project. Normally, you would hide / cover the interface until it's fully loaded and then reveal it all at once instead of letting it incrementally load in. This does come with potential issues of course. It can feel longer to the user, but better for performance. If js produces errors it could leave the loader on the screen even though the error may not be a critical failure.
Without looking into it this is the general idea:
```
branded-checkout:has(loader) {
  position: relative;
  height: 300px;
  overflow: hidden;
}
branded-checkout loader {
  position: absolute;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  background: #fff;
}
```
This would cover the entire branded-checkout, limit it to 300px without scrolling and cover it with the loader until the loader disappears. This is probably the quickest / easiest solution.
Now that I've thought about it, I could easily do this in the WP plugin. I just hadn't thought how to do it yet.

PR #1231, opened 2025-08-13
[HS-1424192] Initialize the amount radio buttons with the suggested amount

canac:
## Description

When adding a preselected amount in branded checkout v3, the item config would be correct, but the wrong amount would be selected (it was always showing the first amount as selected). This PR changes the radio button to show the correct amount as selected.

## Testing

Edit `branded-checkout.html` to enable v3 and have a preselected amount, for example:

```html
<branded-checkout
  use-v3="true"
  designation-number="2584059"
  amount="76"      
  show-cover-fees="true"
  ng-cloak>
</branded-checkout>
```

Verify that the $76 amount is selected in the UI, not $38.

https://secure.helpscout.net/conversation/3032947354/1424192/

### DIFF
--- a/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
+++ b/src/app/productConfig/productConfigForm/productConfigForm.tpl.html
@@ -136,8 +136,8 @@
                   name="suggestedAmount"
                   id="sa-{{$index}}"
                   ng-click="$ctrl.changeAmount(suggested.amount)"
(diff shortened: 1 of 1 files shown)


dr-bizz:
Stakeholders are in the middle of http://give-stage2.cru.org/ give site staging UAT and there were some issues giving EFT and CC yesterday.
We think they may have been caused by some deploys to web https://github.com/CruGlobal/give-web/actions.
If you could hold off changes that might break http://give-stage2.cru.org/ until we finish UAT, that will be appreciated.

dr-bizz (approved):

PR #1235, opened 2025-08-26
Adds needed build check for deploy job

zweatshirt:
- Adds build job to deploy dependencies/needs list. Deploy should not occur if build fails.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: ${{ (github.ref == 'refs/heads/master' && 'production') || 'staging' }}
-    needs: [lint, test]
+    needs: [lint, test, build]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: ${{ (github.ref == 'refs/heads/master' && 'production') || 'staging' }}
-    needs: [lint, test]
+    needs: [lint, test, build]
```
canac, line 90: What scenario does this prevent against? If the `build` job would fail, wouldn't the `yarn run build` step in this deployment job fail also and abort the deployment anyway?

dr-bizz: Thinking through this, I can go either way.

I will say our normal process is to build, and then once it's successful, we deploy. MPDX works this way.

Since we build on the CI when asking Amplify to deploy. Then Amplify runs a build to deploy.

canac (approved):

PR #1237, opened 2025-09-16
WIP: Brand Updates: colors and fonts

mandyshafer:
Background:
 - We need to update the styles to match the new brand
 - I also added a couple overrides for the new homepage working with cru.org components

Test page: https://author-p56256-e473998.adobeaemcloud.com/content/give/us/en/new-homepage.html?wcmmode=disabled

### DIFF
--- a/src/assets/crubrand/bootstrap-variations/_buttons.scss
+++ b/src/assets/crubrand/bootstrap-variations/_buttons.scss
@@ -56,6 +56,11 @@
     }
 }
 
+a.cmp-button:hover {
+  color: inherit;
+  text-decoration: none;
+}
+
 
 // Alternate buttons
 // --------------------------------------------------
--- a/src/assets/scss/_nav-cart.scss
+++ b/src/assets/scss/_nav-cart.scss
@@ -1,8 +1,8 @@
-$colorCru-gold: #f9b625;
-$colorCru-gray: #666062;
+$colorCru-gold: #ffd000;
+$colorCru-gray: #565652;
 $colorCru-mid-gray: #444244;
 $colorCru-white: #fff;
-$colorCru-blueBright: #3eb1c8;
+$colorCru-blueBright: #00c0d8;
 $colorGrayscale-background: #f7f7f7;
 $colorGrayscale-sand: #f4f4f4;
 $colorGrayscale-pebble: #f0f0f0;
@@ -10,14 +10,14 @@ $colorGrayscale-stone: #e7e7e7;
 $colorGrayscale-rock: #b9b9b9;
 $colorGrayscale-slate: #909b9c;
 $colorGrayscale-black: #000;
-$colorText: #3a3a3a;
+$colorText: #000;
 $sm: 992px;
 $header-height: 84px;
 $gutter: 30px;
 $query-unit: 16px;
 $nav-break-min: 400px;
 $fontWeight5: 500;
-$data-font: "Helvetica Neue", helvetica, sans-serif;
+$data-font: "Inter", sans-serif;
 $border-radius: 4px;
 
 
--- a/src/assets/cru-scss/_cru-settings.scss
+++ b/src/assets/cru-scss/_cru-settings.scss
@@ -40,18 +40,18 @@ $gold-bar-size: 3px;
 /**
  * Cru Colors
  */
-$colorCru-gold:       #f9b625;
-$colorCru-gray:       #666062;
+$colorCru-gold:       #ffd000;
+$colorCru-gray:       #565652;
 $colorCru-white:      #fff;
-$colorCru-orange:     #dd7d1b;
-$colorCru-blueDeep:   #007398;
-$colorCru-blueBright: #3eb1c8;
+$colorCru-orange:     #f08020;
+$colorCru-blueDeep:   #007890;
+$colorCru-blueBright: #00c0d8;
 
 
 /**
  * Type
  */
-$colorText:         #3a3a3a;
+$colorText:         #000;
 $colorText--muted:  #67747c;
 $colorText--washed: #abbabf ; // Uppercase and bold
 $colorText--link:   $colorCru-blueDeep;
--- a/src/assets/cru-scss/_slider.scss
+++ b/src/assets/cru-scss/_slider.scss
@@ -217,7 +217,7 @@ $slider-break-min: 1000px;
     }
 }
 
-.slider {
+.slider:not(.list) {
     position: relative;
     max-width: 1200px;
 
--- a/src/assets/crubrand/_overrides.scss
+++ b/src/assets/crubrand/_overrides.scss
@@ -23,7 +23,7 @@ $brand-warning: lighten($colorCru-orange, 10%);
 //** Background color for `<body>`.
 $body-bg:          #f7f7f7;
 //** Global text color on `<body>`.
-$text-color:       $colorGrayscale-text;
+$text-color:       $colorGrayscale-black;
 
 //** Global textual link color.
 $link-color:       $colorCru-blueDeep;
@@ -38,7 +38,7 @@ $link-hover-color: darken($link-color, 10%);
 //== Typography
 //
 //## Font, line-height, and color for body text, headings, and more.
-$font-family-base:     "Source Sans Pro", "Helvetica Neue", Arial, Helvetica, sans-serif;
+$font-family-base:     "Inter", sans-serif;
 
 $font-size-h1:         48px;
 $font-size-h2:         32px;
--- a/src/assets/crubrand/bootstrap-variations/_type.scss
+++ b/src/assets/crubrand/bootstrap-variations/_type.scss
@@ -26,6 +26,7 @@ h1, h2, h3, h4, h5, h6,
     letter-spacing: -0.02em;
     margin: 0 0  math.div($line-height-pixel, 2);
     line-height: $headings-line-height;
+    font-family: "Sora", sans-serif;
     
     small,
     .small {
--- a/src/assets/crubrand/variables/_colors.scss
+++ b/src/assets/crubrand/variables/_colors.scss
@@ -5,13 +5,13 @@
 /**
  * Cru Colors
  */
-$colorCru-gold:       #f9b625;
-$colorCru-gray:       #666062;
+$colorCru-gold:       #ffd000;
+$colorCru-gray:       #565652;
 $colorCru-mid-gray:   #444244;
 $colorCru-white:      #fff;
-$colorCru-orange:     #dd7d1b;
-$colorCru-blueDeep:   #007398;
-$colorCru-blueBright: #3eb1c8;
+$colorCru-orange:     #f08020;
+$colorCru-blueDeep:   #007890;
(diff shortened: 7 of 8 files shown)


mandyshafer:
I'm not seeing that particular issue with the sign-in page where the "Forgot Password?" link is separate from the other part of the dropdown text. I am however seeing an issue on screenwidths between about 728px and 992px where the entire dropdown text for "Need Help?" shows to the right of the "What is Okta?" dropdown option instead of below "Need Help?". I am seeing this issue on prod as well. Should I fix that to always be under the "Need Help?" text or is that outside the scope of this change?

wrandall22:
Yeah, the forgot password misalignment seems to be fixed now.

wrandall22:
We'll consider the broken styling between 728px and 992px to be out of scope. cc @rguinee

src/common/components/nav/navCartIcon.tpl.html
```diff
@@ -10,7 +10,8 @@
       class="nav-cart-icon nav-cart-icon-desktop"
       ng-click="$ctrl.miniCartViewAnalyticsEvent()"
     >
-      <span class="fal fa-shopping-cart">
+      <span class="material-icons-outlined">
+        volunteer_activism
         <span class="cart-item-count" ng-if="!$ctrl.cartOpen"></span>
```
wrandall22, line 15: Looks like the Figma has restyled the count bubble to be blue. It seems to be blue on dev-cloud.cru.org, but not on give-dev-cloud.cru.org

src/assets/scss/_nav-cart.scss
wrandall22, line 1: Looks like the buttons are not quite the right styling yet. The design has them shorter and wider:

Image

wrandall22 (changes requested):
Probably things you already are working on, but made a few comments from what I saw while poking around

src/assets/scss/_nav-cart.scss
```diff
@@ -44,23 +44,57 @@ $cru-tango-alt: #f5a055;
...
   border: none;
+  border-radius: 5px;
   margin: 0;
-  background-color: $cru-scorpion-light;
-  color: $cru-scorpion;
+  background-color: $colorCru-white;
+  box-shadow: 0 4px 4px 0 rgba(0, 0, 0, 0.25);
+  color: $colorGrayscale-black;
+
+  p {
```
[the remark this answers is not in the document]
mandyshafer: I'm assuming you mean the text-transform on the buttons. The new header doesn't have #cru-header but #cru23-header, so it wouldn't get that styling.

mandyshafer:
Since the button to open the cart modal popup and the anchor tag to the cart page are different elements, only the one for the <nav-cart-icon> element was updated in this PR. I have now updated the other one, but that update went in the giveaem codebase on my new-homepage-template branch/PR.

src/assets/scss/_form.scss
wrandall22, line 1: The account type on EFT payment types is still being pushed to 2 lines

rguinee:
We may want to consider adding this text-decoration style to match cru.org context navigation link styles. They are not underlined.
```
#cru-context-nav .first-level-inline li a {
    text-decoration: none;
}
```

wrandall22:
Some testing URLs:
- https://give-dev-cloud.cru.org
- https://give-dev-cloud.cru.org/new-homepage.html
- https://give-stage2.cru.org/

src/assets/scss/_form.scss
mandyshafer: What page is that on?

mandyshafer:
The text-decoration: none updates have been added, some of which are in https://github.com/CruGlobal/giveaem/pull/596

src/assets/crubrand/bootstrap-variations/_buttons.scss
```diff
@@ -17,8 +17,9 @@
     background-image: none; // Reset unusual Firefox-on-Android default style; see https://github.com/necolas/normalize.css/issues/214
     border: $btn-border-size solid transparent;
     white-space: nowrap;
+    text-decoration: none;
 
-    @include button-size($padding-base-vertical, $padding-base-horizontal, $btn-font-size, $line-height-base, $border-radius-base);
+    @include button-size(8px, 24px, $btn-font-size, $line-height-base, $border-radius-base);
```
rguinee, line 22: `$padding-large-horizontal` resolves to 24px. Unfortunately I don't see a variable that resolves to 8px for padding. Line height seems to be hardcoded to 29px in [src/assets/crubrand/variables/_type.scss](https://github.com/CruGlobal/give-web/blob/e0e4a9627d6bc4c3ebeaf5525aebeb5089d2e95c/src/assets/crubrand/variables/_type.scss#L11)

j2trumpet: I'm unsure what you're getting at.

src/common/components/nav/navCart/navCart.tpl.html
rguinee, line 1: I experienced some unexpected behavior at breakpoints above and below ~990. Attaching a video to show. 
1. I would have expected there to be a badge on the cart icon, but there isn't >991px.
2. I would have expected the `nav-cart-dropdown` <993px as was [designed](https://www.figma.com/design/4F9LG1YepsO0Y5nevweQxm/Cru-Style-Guide---Components--Jarrett-?node-id=22204-564&t=bg8PtSjpGITsMkjl-1). 

https://github.com/user-attachments/assets/e8878a7e-5c91-44a1-b6bd-a8cfded68669

src/assets/crubrand/bootstrap-variations/_buttons.scss
j2trumpet: I guess we could use that variable instead. It's no more fragile than the variables we were previously using.

src/common/components/nav/navCart/navCart.tpl.html
mandyshafer: The badge not showing up above 991 was a miss on my part. The fix for that is in my branding updates PR on the giveaem codebase. I put a question in slack regarding the functionality of the icon on mobile (keeping current header functionality vs following the designs).

dr-bizz:
Screenshot 2025-10-02 at 4 25 29 PM

dr-bizz:
Screenshot 2025-10-02 at 4 27 19 PM

wrandall22:
@dr-bizz those screenshots are from the old homepage which will be going away simultaneously (if I understand correctly, @mandyshafer let me know if that is not true).

mandyshafer:
As far as I know, the plan is to launch the new homepage at the same time as the branding updates. If that component is used anywhere else, I can make those updates, but if it is just on the homepage, it will be going away.

dr-bizz:
@wrandall22 What is the new homepage URL?

src/common/components/nav/navCartIcon.tpl.html
```diff
@@ -10,7 +10,8 @@
       class="nav-cart-icon nav-cart-icon-desktop"
       ng-click="$ctrl.miniCartViewAnalyticsEvent()"
     >
-      <span class="fal fa-shopping-cart">
+      <span class="material-icons-outlined">
+        volunteer_activism
```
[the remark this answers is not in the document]
mandyshafer: The way Google Material Icons work, we include the name of the icon as the "text" within the span, and then the font shows the icon related to that name.

mandyshafer:
Are we planning to use the old header/sub menu? The new homepage link also shows the new header, and the submenu links are not underlined there.

index.html
dr-bizz, line 1: It seems the Gift frequency has extra padding on the bottom making the text not vertically aligned.

Image

dr-bizz, line 1: On the checkout, the stars that replace the card number go onto a new line after 7 stars.

Image

wrandall22, line 1: FWIW, `index.html` is only used in local development.

j2trumpet (approved):
I don't know if we want @wrandall22, @dr-bizz, or @rguinee to sign off on this as well, but I'm thinking this is good to go.

mandyshafer:
It's ready for re-review. Generally, I put a thumbs up reaction on the comments that I have addressed.

wrandall22:
Still seeing some things pushed to 2 lines like on the cart page:

Screenshot 2025-10-16 at 9 58 08 AM

I also wonder if it is worth changing the verbiage from `Your Gift Cart` to something else while we're changing icons/language on the nav.

mandyshafer:
I know we are trying to move away from cart terminology. Do we have a consensus on what the terminology should be? I don't know who should say what the new content should be there.

wrandall22:
I don't think there is a consensus on the terminology.

j2trumpet:
I don't know the environment differences between `give-stage2.cru.org` and `give-dev-cloud.cru.org`, but the header on https://give-dev-cloud.cru.org/new-homepage.html is currently destroyed.

j2trumpet:
> Still seeing some things pushed to 2 lines like on the cart page:
> 
> Screenshot 2025-10-16 at 9 58 08 AM

I'm not sure there's much we can do about that instance on the cart page without changing the structure of the table that contains that information. What could be done without much effort is to hide `.sep` from browsers up to width 992px. I believe this could be done by changing the code [here](https://github.com/CruGlobal/give-web/blob/3dc1dc930494e1f1c16aaddc9c42c2421131afbe/src/assets/scss/_cart.scss#L97-L103) to this:

```
      @include respond-to(extrasmall) {
        text-align: left;
      }

      @include respond-to(small) {
        .sep {
          display: none;
        }
      }
```

j2trumpet:
@mandyshafer Good catch. I didn't realize that the `respond-to(small)` would have a `min-width` value.

dr-bizz (commented):
Awesome work on this. This is looking very close.

I spotted some things on the staging site. Nothing crazy.

1. On the Gift modal, the close wasn't showing up when I opened the first 2-3 Gift modals, but after a few, it loaded in. (_This could just be that an image loaded slowly_)
Image

2. I noticed that all the buttons and links on the give site don't have the normal click arrow. (_This might be my browser._)
Image


3. On the checkout, the credit/debit card icon was off the grey card.
Image

4. On the summary step on the checkout, I noticed that you show the text "Gift Amount" on mobile but not on tablet.

Image

Image

src/common/components/nav/navCart/navCart.tpl.html
```diff
@@ -52,26 +39,33 @@
...
+          </p>
         </div>
       </div>
     </div>
   </div>
-
   <div class="row repeating-row sub-total" ng-if="$ctrl.hasItems">
     <div class="col-xs-9">
       <p>
+        Total
```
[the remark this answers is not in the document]
mandyshafer: I don't see why not. Added!

wrandall22:
On the add payment modal, it looks like the security code font size is 14px compared to the 12px of the other labels.

Screenshot 2025-11-03 at 9 27 26 AM

wrandall22:
Going to approve even though there is an outstanding comment, so we can go live today.

wrandall22 (approved):

PR #1271, opened 2026-05-15
Detect partial-login state and fix shadowing cortex-role cookie

dr-bizz:
## Summary
- Users can land "partially logged in" — authenticated to Okta but with a stale `cortex-role` cookie scoped to `.cru.org` shadowing the `give.cru.org` one. `currentRole()` returns `IDENTIFIED` instead of `REGISTERED` and the sign-in modal pops on protected pages.
- Surfaces a warning in the sign-in form when Okta says signed in but Cortex role is anything other than `REGISTERED` (was previously only `PUBLIC` — missed the real-world `IDENTIFIED` case).
- The warning copy is now a self-service escalation ladder, ordered to maximize self-resolution and minimize support contacts: (1) **reload the page** — this triggers the self-heal below and silently fixes the common duplicate-cookie case; (2) if it persists, **clear cookies for `cru.org`** via the Chrome steps (kept intact, since Chrome's per-site wipe does not touch parent-domain cookies); (3) only if that still fails, **contact `eGift@cru.org`**.
- After a successful `/okta/login`, if the readable `cortex-role` JWT is still not `REGISTERED`, removes the parent-domain copy so the `give.cru.org` cookie wins. Legitimate `REGISTERED` `.cru.org` cookies (cross-property SSO) are left alone.
- Self-heals the shadow on every session refresh (page load + cookie-watcher refire), not just on a fresh `/okta/login`, via the exported `hasDuplicateCortexRole` helper — so users already stuck before this deploys recover on their next page load with no manual cookie clearing or re-login. This is why "reload the page" is the correct first step in the message above.

## Known limitation (accepted)
This cleans up the give-side shadow only. A side effect: you can no longer hold a staging **and** production session in the same browser profile — staging scopes `cortex-role` to the parent `.cru.org`, so loading a production page detects the duplicate and removes staging's copy (production always wins). This is not a regression — previously the same cookie pair produced the stuck-modal bug on production instead. Cross-environment testing should use separate browser profiles. The real root-cause fix is backend: scope staging `cortex-role` to `.give-stage2.cru.org` like production scopes to `.give.cru.org` (follow-up with the Cortex/Okta cookie owners).

## Test plan
- [ ] All 1821 unit tests pass locally
- [ ] Reproduce the bug (stale `.cru.org` `cortex-role`) → warning appears in the sign-in modal
- [ ] Verify the message escalation reads correctly: reload first, clear-cookies second, contact support last
- [ ] Confirm the Chrome instructions clear the right cookies
- [ ] Sign in cleanly → no stale parent-domain cookie remains after `/okta/login`
- [ ] Sign in while a legitimate `REGISTERED` `.cru.org` cookie exists → confirm it is NOT removed
- [ ] Load any give page while a duplicate `cortex-role` exists → duplicate is cleared and role settles to `REGISTERED` without re-login (the "reload" self-service step)

### DIFF
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -489,7 +489,7 @@ export const appConfig = /* @ngInject */ function (
     SESSION_MISMATCH_STEP_1:
       'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
     SESSION_MISMATCH_STEP_2:
-      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site and Shields Settings</strong>',
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site Settings</strong>',
     SESSION_MISMATCH_STEP_3:
       'Click <strong>View permissions and data stored across sites</strong>',
     SESSION_MISMATCH_STEP_4:
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -489,11 +489,13 @@ export const appConfig = /* @ngInject */ function (
     SESSION_MISMATCH_STEP_1:
       'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
     SESSION_MISMATCH_STEP_2:
-      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site Settings</strong>',
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Third-Party Cookies</strong>',
     SESSION_MISMATCH_STEP_3:
-      'Click <strong>View permissions and data stored across sites</strong>',
+      'Click <strong>See all site data and permissions</strong>',
     SESSION_MISMATCH_STEP_4:
       'Find <strong>cru.org</strong> and click the <strong>trash can icon</strong> to delete it',
+    SESSION_MISMATCH_RELOAD_AGAIN:
+      'Then reload the page and log in again.',
     SESSION_MISMATCH_CONTACT:
       'Still stuck? Email <a href="mailto:techhelp@cru.org">techhelp@cru.org</a> - we\'re happy to help!',
   });
@@ -808,11 +810,13 @@ export const appConfig = /* @ngInject */ function (
     SESSION_MISMATCH_STEP_1:
       'Haga clic en los <strong>tres puntos</strong> (esquina superior derecha) &rarr; <strong>Configuración</strong>',
     SESSION_MISMATCH_STEP_2:
-      'Vaya a <strong>Privacidad y seguridad</strong> &rarr; <strong>Configuración de sitios</strong>',
+      'Vaya a <strong>Privacidad y seguridad</strong> &rarr; <strong>Cookies de terceros</strong>',
     SESSION_MISMATCH_STEP_3:
-      'Haga clic en <strong>Ver permisos y datos almacenados en los sitios</strong>',
+      'Haga clic en <strong>Ver todos los datos y permisos de los sitios</strong>',
     SESSION_MISMATCH_STEP_4:
       'Busque <strong>cru.org</strong> y haga clic en el <strong>ícono de papelera</strong> para eliminarlo',
+    SESSION_MISMATCH_RELOAD_AGAIN:
+      'Luego vuelva a cargar la página e inicie sesión de nuevo.',
     SESSION_MISMATCH_CONTACT:
       '¿Aún tiene problemas? Escríbanos a <a href="mailto:techhelp@cru.org">techhelp@cru.org</a>; ¡con gusto le ayudaremos!',
   });
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -494,8 +494,7 @@ export const appConfig = /* @ngInject */ function (
       'Click <strong>See all site data and permissions</strong>',
     SESSION_MISMATCH_STEP_4:
       'Find <strong>cru.org</strong> and click the <strong>trash can icon</strong> to delete it',
-    SESSION_MISMATCH_RELOAD_AGAIN:
-      'Then reload the page and log in again.',
+    SESSION_MISMATCH_RELOAD_AGAIN: 'Then reload the page and log in again.',
     SESSION_MISMATCH_CONTACT:
       'Still stuck? Email <a href="mailto:techhelp@cru.org">techhelp@cru.org</a> - we\'re happy to help!',
   });
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -482,6 +482,20 @@ export const appConfig = /* @ngInject */ function (
     REMOVE_SPOUSE: 'Remove',
     SUGGESTED_AMOUNT_HELP: 'Suggested Gift Amounts. Tab for Custom Amount',
     CUSTOM_AMOUNT: 'Custom Amount',
+    SESSION_MISMATCH_RELOAD:
+      '<strong>Please reload this page and log in again.</strong>',
+    SESSION_MISMATCH_CLEAR_DATA:
+      "If it still doesn't work, you'll need to clear saved data for cru.org in Chrome:",
+    SESSION_MISMATCH_STEP_1:
+      'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
+    SESSION_MISMATCH_STEP_2:
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site and Shields Settings</strong>',
+    SESSION_MISMATCH_STEP_3:
+      'Click <strong>View permissions and data stored across sites</strong>',
+    SESSION_MISMATCH_STEP_4:
+      'Find <strong>cru.org</strong> and click the <strong>trash can icon</strong> to delete it',
+    SESSION_MISMATCH_CONTACT:
+      'Still stuck? Email <a href="mailto:techhelp@cru.org">techhelp@cru.org</a> - we\'re happy to help!',
   });
 
   $translateProvider.translations('es', {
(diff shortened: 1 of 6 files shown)


src/common/services/session/session.service.js
```diff
@@ -210,6 +228,27 @@ const session = /* @ngInject */ function (
           });
         })
         .then((response) => {
+          // /okta/login sets a give.cru.org-scoped cortex-role cookie with
```
wrandall22, line 236: Is this correctly handling the case where a user has an Okta login but not a donor account? I forget what role they get.

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -1,3 +1,38 @@
...
+  role="alert"
+  ng-if="$ctrl.showSessionMismatchMessage"
+>
+  <p translate>
+    It appears that you logged in, but the system got a bit confused due to
+    previous cookies. Currently you are only partially logged in.
+  </p>
+  <p translate>
+    <strong>First, try reloading this page</strong> and signing in again &mdash;
+    we&rsquo;ll attempt to straighten this out automatically.
```
wrandall22: Any reason not to just use `we'll`?

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -1,3 +1,38 @@
...
+      Find <strong>cru.org</strong> and click the trash can icon (&ldquo;Delete
+      site data and permissions for cru.org&rdquo;). Some of the stale cookies
+      are on the parent <em>cru.org</em> domain, not just <em>give.cru.org</em>,
+      so this is the level that needs to be cleared.
+    </li>
+  </ol>
+  <p translate>
+    Then reload this page and log in again. If it still doesn&rsquo;t work after
+    clearing your cookies, contact
+    <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.
```
wrandall22: Did this originate with Dev Services, Donation Services, or engineering? Is eGift aware of this new potential influx of tickets?

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -1,3 +1,38 @@
+<div
+  class="alert alert-warning"
+  role="alert"
+  ng-if="$ctrl.showSessionMismatchMessage"
+>
+  <p translate>
```
wrandall22: Lots of translatable text has been added here, but none of it has made it into the translation strings.

wrandall22 (changes requested):
This will require coordination with Sean McKenna in Dev Services as well. @joannacatanus-ux can you review this from a user experience perspective also?

joannacatanus-ux:
@dr-bizz I've been asked to review this from a UX perspective. Would you be able to replicate it on stage and send screenshots of what it looks like? Thank you!

dr-bizz:
@joannacatanus-ux So sorry it has taken me a while to get back to you.

This is a screenshot of the text:

Screenshot 2026-05-29 at 3 25 21 PM

joannacatanus-ux:
@dr-bizz @wrandall22 From a UX perspective, I think we can make the message more concise and focused on the fix. Providing unnecessary context can be confusing. They may not need to know why it appears that they are logged in- they just need to know how to log back in. It would also be helpful to have key parts of the instructions in bold. Also wanted to confirm -does this only happen on Chrome? Here's my suggestion for wording:

**Please reload this page and log in again.**

If it still doesn't work, you'll need to clear saved data for cru.org in Chrome:
1. Click the **three dots** (top-right corner) → **Settings**
2. Go to **Privacy and Security** → **Third-Party Cookies**
3. Click **See all site data and permissions**
4. Find **cru.org** and click the **trash can icon** to delete it

Then reload the page and log in again.

**Still stuck?** Email eGift@cru.org— we're happy to help!

dr-bizz:
Please reload this page and log in again.

If it still doesn't work, you'll need to clear saved data for cru.org in Chrome:

Click the three dots (top-right corner) → Settings
Go to Privacy and Security → Site and Shields Settings
Click "View permissions and data stored across sites"
Find cru.org and click the trash can icon to delete it
Then reload the page and log in again.

Still stuck? Email [eGift@cru.org](mailto:eGift@cru.org)— we're happy to help!

src/common/components/signInForm/signInForm.tpl.html
dr-bizz: Engineering, so I switched it to <a href="mailto:techhelp@cru.org">techhelp@cru.org</a>

src/common/app.config.js
```diff
@@ -474,6 +474,20 @@ export const appConfig = /* @ngInject */ function (
...
     SUGGESTED_AMOUNT_HELP: 'Suggested Gift Amounts. Tab for Custom Amount',
     CUSTOM_AMOUNT: 'Custom Amount',
+    SESSION_MISMATCH_RELOAD:
+      '<strong>Please reload this page and log in again.</strong>',
+    SESSION_MISMATCH_CLEAR_DATA:
+      "If it still doesn't work, you'll need to clear saved data for cru.org in Chrome:",
+    SESSION_MISMATCH_STEP_1:
+      'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
+    SESSION_MISMATCH_STEP_2:
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site and Shields Settings</strong>',
```
wrandall22: I don't have a `Site and Shields Settings` in Chrome. What I see is what Joanna suggested.

src/common/components/signInForm/signInForm.tpl.html
```diff
@@ -3,35 +3,15 @@
...
-      are on the parent <em>cru.org</em> domain, not just <em>give.cru.org</em>,
-      so this is the level that needs to be cleared.
-    </li>
+    <li translate="SESSION_MISMATCH_STEP_1"></li>
+    <li translate="SESSION_MISMATCH_STEP_2"></li>
+    <li translate="SESSION_MISMATCH_STEP_3"></li>
+    <li translate="SESSION_MISMATCH_STEP_4"></li>
   </ol>
-  <p translate>
-    Then reload this page and log in again. If it still doesn&rsquo;t work after
```
wrandall22: We lost this reload text

wrandall22 (changes requested):

src/common/app.config.js
dr-bizz: Ah, I'm on brave, I'll switch it back

src/common/services/session/session.service.js
dr-bizz: I addressed this in the new code

src/common/services/session/session.service.spec.js
```diff
@@ -177,6 +218,103 @@ describe('session service', function () {
...
+        staleIdentified,
+      );
+
+      $rootScope.$digest();
+
+      expect($cookies.remove).not.toHaveBeenCalledWith(
+        Sessions.role,
+        expect.objectContaining({ domain: cookieDomain }),
+      );
+      expect(jar.entries).toEqual([freshRegistered, staleIdentified]);
```
wrandall22, line 268: I don't see how this test is doing something different from `it('removes a shadowing parent-domain copy on session refresh and recovers the fresh role',`. The only difference I see on setup is the order you are passing the cookies to in `fakeCookieJar`, but that mock doesn't seem to care about order. I seem to be missing something about how the mock is working?

src/common/services/session/session.service.spec.js
```diff
@@ -177,6 +218,103 @@ describe('session service', function () {
...
+      expect(sessionService.session.cortexRoleShadowed).toBeUndefined();
+    });
+
+    it('does not treat a single non-REGISTERED cookie as a shadow', () => {
+      // An Okta session without a Cortex login (e.g. no donor account yet,
+      // or a failed /okta/login) has no duplicate cookie and must not
```
wrandall22: Has this been verified as true? Someone can be partially logged in without a cortex login (at least they could when the login was available on cru.org). Do we know they won't have a duplicate cookie in that case?

src/common/app.config.js
```diff
@@ -476,6 +476,20 @@ export const appConfig = /* @ngInject */ function (
...
+    SESSION_MISMATCH_RELOAD:
+      '<strong>Please reload this page and log in again.</strong>',
+    SESSION_MISMATCH_CLEAR_DATA:
+      "If it still doesn't work, you'll need to clear saved data for cru.org in Chrome:",
+    SESSION_MISMATCH_STEP_1:
+      'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
+    SESSION_MISMATCH_STEP_2:
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site Settings</strong>',
+    SESSION_MISMATCH_STEP_3:
+      'Click <strong>View permissions and data stored across sites</strong>',
```
wrandall22: This is still based on Brave.

src/common/app.config.js
```diff
@@ -483,7 +483,7 @@ export const appConfig = /* @ngInject */ function (
     SESSION_MISMATCH_STEP_1:
       'Click the <strong>three dots</strong> (top-right corner) &rarr; <strong>Settings</strong>',
     SESSION_MISMATCH_STEP_2:
-      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site and Shields Settings</strong>',
+      'Go to <strong>Privacy and Security</strong> &rarr; <strong>Site Settings</strong>',
```
wrandall22: You need to update the Spanish when you update the English

wrandall22 (changes requested):
I don't see any tests with fresh identified or fresh public cookies, are these worth testing?

src/common/services/session/session.service.spec.js
dr-bizz: I have updated it to make sure we take this into account

wrandall22 (approved):

PR #1282, opened 2026-06-10
EP-2490/EP-2491 - Don't fail entire payment method list when one instrument is malformed

dr-bizz:
## Jira
[EP-2490](https://jira.cru.org/browse/EP-2490) (Critical), [EP-2491](https://jira.cru.org/browse/EP-2491)

## Problem
One malformed payment instrument — missing `payment-instrument-identification-attributes` (seen with instruments created in checkout without a billing address) — threw a TypeError during list mapping. The error took down the whole observable, so:
- self-service showed "There was an error loading your payment methods. You may try again." (EP-2491)
- checkout step 2 silently fell back to the empty/new-payment form as if the donor had no saved payment methods (EP-2490)

## Fix
Optional-chaining guards at the five mapping sites:
- `profile.service.js`: `getPaymentMethods()`, `getPaymentMethod()`, `getPaymentMethodsWithDonations()`
- `order.service.js`: `getExistingPaymentMethods()`, `getCurrentPayment()`

Malformed entries are now mapped defensively (attribute fields absent) instead of failing the list; well-formed results are byte-identical (covered by full-list assertions).

## Tests
5 new specs written red-first (each reproduces the production TypeError without the fix). `profile.service.spec.js` + `order.service.spec.js`: 117 passing.

## Notes
- This fixes the **symptom**. The backend root cause (DELETE not persisted + NPE in `CruPaymentResourceHelpers.buildPaymentInstrumentAttributesEntity`) is tracked in [EP-2633](https://jira.cru.org/browse/EP-2633).
- Review follow-ups (not in scope): same unguarded read in `payment-methods.component.js:147` add-flow handler; several legacy specs in these files assert vacuously (failures swallowed by $q digest).

## How you can test it

1. Add multiple credit cards to your payment methods. Test cards are here - https://developer.globalpayments.com/resources/test-cards
2. Add a gift to your cart.
3. Ensure all show on step 2
Screenshot 2026-06-11 at 3 31 11 PM

### DIFF
--- a/src/common/services/api/order.service.js
+++ b/src/common/services/api/order.service.js
@@ -304,7 +304,7 @@ class Order {
           if (
             paymentMethod.description[
               'payment-instrument-identification-attributes'
-            ]['street-address']
+            ]?.['street-address']
           ) {
             paymentMethod.description.address = formatAddressForTemplate(
               paymentMethod.description[
@@ -339,7 +339,9 @@ class Order {
       .map((data) => {
         if (
           data &&
-          data['payment-instrument-identification-attributes']['street-address']
+          data['payment-instrument-identification-attributes']?.[
+            'street-address'
+          ]
         ) {
           data.address = formatAddressForTemplate(
             data['payment-instrument-identification-attributes'],
--- a/src/common/services/api/order.service.spec.js
+++ b/src/common/services/api/order.service.spec.js
@@ -1032,7 +1032,7 @@ describe('order service', () => {
       self.$httpBackend.flush();
     });
 
-    it('should not fail when the current payment is missing payment-instrument-identification-attributes', (done) => {
+    it('should pass when the current payment is missing payment-instrument-identification-attributes', (done) => {
       const clonedResponse = angular.copy(paymentMethodCreditCardResponse);
       const description =
         clonedResponse._order[0]._paymentinstrumentselector[0]._chosen[0]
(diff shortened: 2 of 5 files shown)


src/common/services/api/order.service.js
```diff
@@ -339,7 +339,9 @@ class Order {
       .map((data) => {
         if (
           data &&
-          data['payment-instrument-identification-attributes']['street-address']
+          data['payment-instrument-identification-attributes']?.[
+            'street-address'
+          ]
```
wjames111, line 344: Nice clean fix!

src/common/services/api/order.service.spec.js
```diff
@@ -998,6 +1031,31 @@ describe('order service', () => {
 
       self.$httpBackend.flush();
     });
+
+    it('should not fail when the current payment is missing payment-instrument-identification-attributes', (done) => {
```
wjames111:
```Suggestion
it('should pass when the current payment is missing payment-instrument-identification-attributes', (done) => {
```

wjames111 (approved):
This looks great, nice easy fix! I noticed the same pattern in [payment-methods.component.js:146-149](vscode-webview://1mc9osnqc9ipso0f614dadk5jg48iqvapfd5ua8f6i0pi3gaphmt/src/app/profile/payment-methods/payment-methods.component.js#L146-L149) we might want to add optional chaining there as well.

One other implication, `extractPaymentAttributes` always runs regardless, If Cortex returns the key present with a null/undefined value rather than omitting it, what would happen? It might be worth updating at the callsite directly.

src/common/services/api/order.service.js
dr-bizz: Thanks

PR #1283, opened 2026-06-10
EP-2320 - Don't fail product config form when designation has no AEM page

dr-bizz:
## Jira
[EP-2320](https://jira.cru.org/browse/EP-2320)

## Problem
`givingLinks()` in `designations.service.js` fetches the designation's AEM `.infinity.json` content with no error handling. When a designation has no AEM page (returns 300/404/500 — common for branded checkout designations on third-party sites), the error propagated through the merged observable in `productConfigForm.loadData()` and killed the entire gift config form (`errorLoading = true`).

The sibling `suggestedAmounts()` — which fetches the same URL — already recovers with `.catch(() => Observable.of([]))`.

## Fix
Apply the same catch to `givingLinks()`: an AEM content failure now yields an empty giving-links list and the form proceeds normally. Consumer already handles `[]` (`givingLinks || []`, length-gated).

## Tests
New red-first spec: 404 on the `.infinity.json` request → `givingLinks` emits `[]` instead of erroring. 96/96 passing across `designations.service.spec.js` + `productConfigForm.component.spec.js`.

## Review notes (flagged, out of scope)
- The catch is silent (matches the existing `suggestedAmounts` convention); a debug log could aid AEM-outage diagnosis.
- `getNextDrawDate()` is another unguarded member of the same merged stream (Cortex call, different failure class).

### DIFF
--- a/src/common/services/api/designations.service.js
+++ b/src/common/services/api/designations.service.js
@@ -19,12 +19,14 @@ class DesignationsService {
   constructor(
     $http,
     $location,
+    $log,
     envService,
     cortexApiService,
     hateoasHelperService,
   ) {
     this.$http = $http;
     this.$location = $location;
+    this.$log = $log;
     this.envService = envService;
     this.cortexApiService = cortexApiService;
     this.hateoasHelperService = hateoasHelperService;
@@ -269,7 +271,13 @@ class DesignationsService {
         }
         return suggestedAmounts;
       })
-      .catch(() => Observable.of([]));
+      .catch((error) => {
+        this.$log.error(
+          `Error loading suggested amounts for designation ${code}`,
+          error,
+        );
+        return Observable.of([]);
+      });
   }
 
   facebookPixel(code) {
@@ -311,7 +319,13 @@ class DesignationsService {
         }
         return givingLinks;
       })
-      .catch(() => Observable.of([]));
+      .catch((error) => {
+        this.$log.error(
+          `Error loading giving links for designation ${code}`,
+          error,
+        );
+        return Observable.of([]);
+      });
   }
 
   ministriesList(pagePath) {
--- a/src/common/services/api/designations.service.js
+++ b/src/common/services/api/designations.service.js
@@ -285,31 +285,33 @@ class DesignationsService {
     const path = this.generatePath(code, campaignPage);
     return Observable.from(
       this.$http.get(this.envService.read('publicGive') + path),
-    ).map((data) => {
-      const givingLinks = [];
-      if (data.data['jcr:content']) {
-        // Map giving links
-        if (data.data['jcr:content'].givingLinks) {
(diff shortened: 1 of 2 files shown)


src/common/services/api/designations.service.js
```diff
@@ -285,31 +293,39 @@ class DesignationsService {
...
-      }
-      return givingLinks;
-    });
+        return givingLinks;
+      })
+      .catch((error) => {
+        this.$log.error(
+          `Error loading giving links for designation ${code}`,
+          error,
+        );
```
wjames111:
```Suggestion
  if (error.status === 404) {
    this.$log.info(`No AEM page for designation ${code}`);
  } else {
    this.$log.error(`Error loading giving links for designation ${code}`, error);
  }
```

wjames111: Should not having an AEM page technically be considered an error? It would be good info to have but it might fill up rollbar with a lot of noise.

src/common/services/api/designations.service.js
```diff
@@ -269,7 +271,13 @@ class DesignationsService {
         }
         return suggestedAmounts;
       })
-      .catch(() => Observable.of([]));
+      .catch((error) => {
```
wjames111, line 274: I might do the same check here too and use  `this.$log.info` for designations with no AEM page.

src/common/services/api/designations.service.js
```diff
@@ -285,31 +293,39 @@ class DesignationsService {
...
+              }
+            });
+          }
         }
-      }
-      return givingLinks;
-    });
+        return givingLinks;
+      })
+      .catch((error) => {
```
wjames111, line 332: Not sure if its worth the trouble but we could turn into a helper.

wjames111 (approved):
Nice catch finding this, it'll be great to know which designations don't have AEM pages!

src/common/services/api/designations.service.js
wrandall22: The designation may be inactive, in which case the giving will fail. It is more likely that the search incorrectly brought back a result (stale) than the page didn't properly get created in the first place.

dr-bizz: Implemented in edfb1c16 — 404s now log as info ("No AEM page for designation …") and other failures stay at error level. The Rollbar $log decorator forwards info-level too, so we keep visibility on the inactive/stale-search cases without them landing in the error stream.

src/common/services/api/designations.service.js
dr-bizz: Done — both catches now go through a shared logAemContentError helper with the same 404-vs-error split.

src/common/services/api/designations.service.js
dr-bizz: Extracted logAemContentError, used by both suggestedAmounts and givingLinks.

PR #1284, opened 2026-06-10
EP-2481 - Fix un-interpolated frequency labels and annual gift day display

dr-bizz:
## Jira
[EP-2481](https://jira.cru.org/browse/EP-2481)

## Problems
1. **Un-interpolated labels:** `giftSummaryView.tpl.html` had `translate` attributes on elements containing interpolation bindings, so angular-translate rendered the raw template text (e.g. `On the {{$ctrl.gift.transactionDay | ordinal}} of...`) in the Stop a Gift dialog and delete-payment-method modal. The identical bug was fixed in `giftDetailsView` by 399bc8b2 (EP-2383) — this template was missed.
2. **Annual gifts showed no day label at all:** both templates switched on `ng-switch-when="Annually"`, but the model's frequency value is `'Annual'` (see `<option value="Annual">` in giftUpdateView and the `frequency === 'Annual'` gates). The `displayRateTotals` "Annually" strings are Cortex rate-display values — a different domain, untouched.

## Fix
Removed the three `translate` attributes (mirroring 399bc8b2 — those keys were never in the translation tables, so no i18n change) and corrected the switch values to `Annual` in both templates.

## Tests
6 new template-rendering specs, written red-first (reproduced both symptoms: raw binding text for quarterly, missing day label for annual). `src/common/components/giftViews`: 20/20 passing.

Screenshot 2026-06-11 at 4 26 51 PM

### DIFF
--- a/src/common/components/giftViews/giftDetailsView/giftDetailsView.component.spec.js
+++ b/src/common/components/giftViews/giftDetailsView/giftDetailsView.component.spec.js
@@ -1,5 +1,6 @@
 import angular from 'angular';
 import 'angular-mocks';
+import moment from 'moment';
 
 import module from './giftDetailsView.component';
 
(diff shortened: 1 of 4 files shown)


canac (approved):
Thanks for fixing this!

PR #1285, opened 2026-06-10
EP-2378 - Require a valid amount before submitting recurring gifts

dr-bizz:
## Jira
[EP-2378](https://jira.cru.org/browse/EP-2378) (High)

## Problem
The "Edit my recurring donations" flow allowed submitting recurring gifts with no amount:
- Step 1 Continue was only gated on `allPaymentMethodsValid()` — amount never checked.
- Step 3 ("Set up your new recurring gifts") Continue had **no `ng-disabled` at all** — the $50 default could be cleared and submitted blank.
- A cleared amount became `updated-amount: undefined`, which `hasChanges()` treated as a change, flowing a blank/$0.00 amount to the confirmation step.

## Fix
- New `RecurringGiftModel.hasValidAmount()` mirroring the giftUpdateView input rules (pattern, ≥ $1, ≤ $10,000,000) — single shared rule for both steps.
- Step 1: Continue now also requires `allAmountsValid()`. The no-gifts case (`every([]) === true`) is deliberately preserved — the template invites users with no gifts to continue and add new ones.
- Step 3: Continue now gated on `allGiftsValid()` (payment method + valid amount per addition).
- `hasChanges()` ignores `updated-amount: undefined` (defense-in-depth; the gates are the primary fix).
- Audited the remaining steps: step 2 only selects recipients (no amounts); step 4 was already gated.

## Out of scope
The broader pattern (outer modal buttons not consuming inner `giftForm` validity, affecting giveOneTimeGift/restartGift/redirectGift) is tracked in [EP-1996](https://jira.cru.org/browse/EP-1996).

## Tests
14 new specs written red-first (model validity rules, both step gates, hasChanges with cleared amount). Targeted suites: 161/161. Regression across all RecurringGiftModel consumers: 305/305.

### DIFF
--- a/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
+++ b/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
@@ -42,6 +42,10 @@ class EditRecurringGiftsController {
   allPaymentMethodsValid() {
     return every(this.recurringGifts, (gift) => gift.paymentMethod);
   }
+
+  allAmountsValid() {
+    return every(this.recurringGifts, (gift) => gift.hasValidAmount());
+  }
 }
 
 export default angular
--- a/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
+++ b/src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
@@ -39,12 +39,15 @@ class EditRecurringGiftsController {
     }
   }
 
-  allPaymentMethodsValid() {
-    return every(this.recurringGifts, (gift) => gift.paymentMethod);
-  }
-
-  allAmountsValid() {
-    return every(this.recurringGifts, (gift) => gift.hasValidAmount());
+  allGiftsValid() {
+    const hasPaymentMethods = every(
+      this.recurringGifts,
+      (gift) => gift.paymentMethod,
+    );
+    const hasValidAmounts = every(this.recurringGifts, (gift) =>
+      gift.hasValidAmount(),
+    );
+    return hasPaymentMethods && hasValidAmounts;
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/common/models/recurringGift.model.spec.js
```diff
@@ -182,6 +182,54 @@ describe('recurringGift model', () => {
...
+    });
+
+    it('should return false for an amount greater than 10000000', () => {
+      giftModel.amount = 10000001;
+
+      expect(giftModel.hasValidAmount()).toEqual(false);
+    });
+
+    it('should return false for an amount with more than two decimal places', () => {
+      giftModel.amount = 50.123;
```
dr-bizz, line 221: In Brazil, they use decimals instead of commas

dr-bizz: AngularJS ng-model on a type="number" input parses it, so we are good.

src/app/profile/yourGiving/editRecurringGifts/step1/editRecurringGifts.component.js
```diff
@@ -42,6 +42,10 @@ class EditRecurringGiftsController {
   allPaymentMethodsValid() {
     return every(this.recurringGifts, (gift) => gift.paymentMethod);
   }
+
+  allAmountsValid() {
+    return every(this.recurringGifts, (gift) => gift.hasValidAmount());
+  }
```
canac: What about combining these into a single `allGiftsValid` method? That makes it impossible to accidentally forget to check one.

src/common/models/recurringGift.model.js
```diff
@@ -256,8 +266,11 @@ export default class RecurringGiftModel {
   }
 
   hasChanges() {
+    // updated-amount is undefined when the amount input was cleared or invalid
+    // (ng-model sets the model to undefined), so it isn't a change to submit
     return (
-      this.gift['updated-amount'] !== '' ||
+      (this.gift['updated-amount'] !== '' &&
```
canac: optional: What about `!!this.gift['updated-amount']` or `typeof this.gift['updated-amount'] === 'number'`? Is the amount is populated it will be a number, and validation rejects zeros.

canac (approved):
I tested this locally, and it seems to fix it. Thanks for fixing this!

PR #1286, opened 2026-06-10
EP-2335 - Block order submission when the selected credit card has expired

dr-bizz:
## Jira
[EP-2335](https://jira.cru.org/browse/EP-2335) (High)

## Problem
Expired saved credit cards could be submitted at checkout. Step 2 blocks selecting expired cards (`validPaymentMethod` filter + disabled radios), but step 3 (review) never re-validated:
- Deep-linking to `?step=review` skips step 2 entirely, so a card selected in a previous session that has since expired submits as-is.
- A card can also expire between step 2 and submit at a month boundary.
- The donor sees a successful submission, but the gift won't process (see the Rollbar example on the ticket).

## Fix
- New `isCardExpired()` in step-3, reusing the **same** `validPaymentMethod` helper step 2 uses (one source of truth; month-granularity rules identical, bank accounts unaffected).
- `canSubmitOrder()` now requires `!isCardExpired()` — submit button disables live via the existing digest-driven binding.
- `submitOrder()` itself also early-returns on an expired card, guarding the direct call paths that bypass button state (the `submitOrderEvent` listener, the branded-checkout submit binding, and month-boundary expiry after the last digest).
- Template: `alert-danger` in the payment panel — "The credit card you selected has expired. Please choose or add another payment method." — with a CHANGE button reusing the existing `changeStep({newStep: 'payment'})` navigation. Translate key added to both en and es tables.
- A card missing expiry attributes is treated as expired — identical to step 2's behavior with the same helper, so step 3 can never be more permissive than step 2.

## Tests
7 new specs written red-first (no card / bank account / valid / expired / missing-expiry / canSubmitOrder gate / submitOrder guard), using `jest-date-mock` like the existing validPaymentMethods specs. 80/80 passing across step-3, existingPaymentMethods, and paymentHelpers suites.

## Note
Server-side expiry validation at submit remains backend work (per ticket, the backend doesn't always reject at submit).

## How to test

First test on his branch on localhost, then on production to see the error.

1. Add item to cart.
2. Either have a credit card pre-selected or go through the steps, but stop at step 3
3. In the **Network** tab. locate the request that brings in your **pre-selected credit card** details.
4. On the request URL name, **right-click** and select **Override content**
5. Edit the expiration year to 2022.
6. Refresh the page.
Screenshot 2026-06-11 at 5 00 29 PM

### DIFF
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -136,6 +136,20 @@
                   ></display-address>
                 </div>
               </div>
+              <div class="col-md-12" ng-if="$ctrl.isCardExpired()">
+                <div class="alert alert-danger" role="alert">
+                  <p translate>{{'REVIEW_SELECTED_CARD_EXPIRED_ERROR'}}</p>
+                  <button
+                    id="expiredCardChangePaymentButton"
+                    class="btn btn-default"
+                    ng-click="$ctrl.changeStep({newStep: 'payment'})"
+                    aria-label="Update payment method"
+                    translate
+                  >
+                    {{'CHANGE'}}
+                  </button>
+                </div>
+              </div>
             </div>
             <loading ng-if="$ctrl.loadingCurrentPayment"></loading>
           </div>
--- a/src/app/checkout/step-3/step-3.tpl.html
+++ b/src/app/checkout/step-3/step-3.tpl.html
@@ -141,7 +141,7 @@
                   <p translate>{{'REVIEW_SELECTED_CARD_EXPIRED_ERROR'}}</p>
                   <button
                     id="expiredCardChangePaymentButton"
-                    class="btn btn-default"
+                    class="btn btn-primary"
                     ng-click="$ctrl.changeStep({newStep: 'payment'})"
                     aria-label="Update payment method"
                     translate
--- a/src/common/app.config.js
+++ b/src/common/app.config.js
@@ -395,6 +395,8 @@ export const appConfig = /* @ngInject */ function (
     REVIEW_SUBMITTING_PAYMENT_ERROR:
       'There was an error submitting your payment. Verify that your payment info is correct or try adding it again. If you are still seeing this message, please contact <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.',
     REVIEW_CARD_EXPIRED_ERROR: 'Your card is expired.',
+    REVIEW_SELECTED_CARD_EXPIRED_ERROR:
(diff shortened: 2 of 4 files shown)


src/app/checkout/step-3/step-3.tpl.html
```diff
@@ -136,6 +136,20 @@
                   ></display-address>
                 </div>
               </div>
+              <div class="col-md-12" ng-if="$ctrl.isCardExpired()">
+                <div class="alert alert-danger" role="alert">
+                  <p translate>{{'REVIEW_SELECTED_CARD_EXPIRED_ERROR'}}</p>
```
wrandall22, line 141: We'll want to make sure to run this by dev services before pushing.

src/common/app.config.js
```diff
@@ -395,6 +395,8 @@ export const appConfig = /* @ngInject */ function (
     REVIEW_SUBMITTING_PAYMENT_ERROR:
       'There was an error submitting your payment. Verify that your payment info is correct or try adding it again. If you are still seeing this message, please contact <a href="mailto:eGift@cru.org">eGift@cru.org</a> for assistance.',
     REVIEW_CARD_EXPIRED_ERROR: 'Your card is expired.',
+    REVIEW_SELECTED_CARD_EXPIRED_ERROR:
+      'The credit card you selected has expired. Please choose or add another payment method.',
```
wrandall22, line 399: What about editing the expiration date on the card they chose? Not sure if that is too much language for this error message.

dr-bizz: I think most cards get a completely new number when the card renews

src/app/checkout/step-3/step-3.tpl.html
dr-bizz: I'll flag this for them. Since this is a bug fix, not a new feature, I wouldn't think I need approval to fix it, right?

wrandall22: It changes the UX and assumes wording, so I would recommend getting approval.

src/common/app.config.js
wrandall22: Not according to Gemini. It has been awhile since I've gotten a new card, though, so I forget.

dr-bizz: Checked — renewals typically keep the card number (only expiry/CVV change), and the CHANGE button lands on step 2 where saved cards have an edit link to update the expiration. I included an alternate wording ("update the card's expiration date or choose another payment method") in the approval request to dev services.

dr-bizz:
This has been approved by stakeholders in UAT

wjames111 (approved):
Looks good to me!

PR #1287, opened 2026-06-10
EP-2382 - Tighten phone number validation to block malformed input

dr-bizz:
## Jira
[EP-2382](https://jira.cru.org/browse/EP-2382)

## Problem
The shared phone regex in `app.constants.js` (unchanged since 2019) was **not anchored at the start** and **required no actual digits**, so input like `abc!!!1234567`, `call me at 5551234`, or seven hyphens passed client validation and reached the backend phone validator — which crashes non-gracefully on them (see Rollbar links on the ticket).

## Fix
New pattern (same constant, so both consumers — profile self-service and checkout contact info — benefit):
```
/^((?:011[\s\-.]*)?\+?(?:[\s\-().]*\d){7,15})(?:\s*(?:#|x\.?|ext\.?|extension)\s*(\d+))?$/
```
- Anchored; counts **7–15 actual digits** (ITU E.164 max), with separators (space, hyphen, parens, dot) allowed between them.
- Leading `+` and `011` international access prefix supported; the 011 prefix doesn't count toward the digit cap.
- Extension suffix and capture groups preserved verbatim.
- Verified compatible with every phone format in existing specs/fixtures (`(909) 337-2433`, `555-555-5555`, `541-967-0010`, `1234567890`, etc.) plus international and extension forms.
- Reviewed for catastrophic backtracking: bounded quantifiers, disjoint character classes; pathological inputs ≤ 0.2 ms.

## Newly rejected
`abc!!!1234567`, `call me at 5551234`, `-------`, 16+ digit strings, extension marker without digits.

## Tests
New `app.constants.spec.js` (32 tests, written red-first — 7 failed on the old regex). 116/116 green across constants, profile, and contactInfo suites.

## Note
This blocks the malformed input most likely to crash the backend parser; hardening the backend validator itself to return 400 instead of crashing remains Cortex work (per ticket).

### DIFF
--- a/src/common/app.constants.js
+++ b/src/common/app.constants.js
@@ -29,9 +29,10 @@ const mobileBreakpoint = 575;
 
 // Generic phone number regex, matches + and 011 prefixes.
 // Anchored at both ends, allows digits/spaces/hyphens/parentheses/dots,
-// requires 7-15 digits (ITU E.164 max), and supports an optional extension.
+// requires 7-15 digits (ITU E.164 max, excluding a leading 011 international
+// access prefix), and supports an optional extension.
 const phoneNumberRegex =
-  /^(\+?(?:[\s\-().]*\d){7,15})(?:\s*(?:#|x\.?|ext\.?|extension)\s*(\d+))?$/;
+  /^((?:011[\s\-.]*)?\+?(?:[\s\-().]*\d){7,15})(?:\s*(?:#|x\.?|ext\.?|extension)\s*(\d+))?$/;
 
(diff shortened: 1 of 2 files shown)


canac:
> which crashes non-gracefully on them (see Rollbar links on the ticket)

Where are the Rollbar links? Ethan and I tried to improve this on the backend. But I don't think we tested for non-digits.

canac (approved):
Thanks for improving this!

dr-bizz:
It looks like the links are now dead on Rollbar.

PR #1288, opened 2026-06-10
EP-1593 - Search in place instead of full page reload

dr-bizz:
## Jira
[EP-1593](https://jira.cru.org/browse/EP-1593)

## Problem
Both search forms on the results page were native GET submits (`<form method="get" ...>` with no `ng-submit`), so every keyword/name search did a full page reload. The type-filter tabs already searched in place — only form submits reloaded.

## Fix
- Both forms now submit via `ng-submit` handlers: `submitKeywordSearch()` (clears name params, keeps the active type filter) and `submitNameSearch()` (clears keyword, forces `type=people`).
- The URL is kept shareable/bookmarkable via `$location.search({q, fName, lName, type})` — same param names `$onInit` reads, falsy values stripped.
- Back/forward works in place: a `$locationChangeSuccess` listener re-reads params and re-searches, with (a) an equality guard so the component's own URL writes and the bootstrap event don't double-search, and (b) a route-exit guard so navigating away doesn't fire a stray search on the dying scope.
- Name inputs got real `ng-model` bindings (they were one-way `value` interpolations — an in-place name search was impossible without this).
- Removing the hidden `type` input also fixes a pre-existing bug: the old name form submitted the current filter type, and the service silently drops `first_name`/`last_name` unless `type=people` — so advanced name search used to be ignored whenever a non-people filter was active.

## Behavior change to note for QA
A new keyword search now **preserves the active type filter** (the old native submit dropped it and reset to All). This matches the visible tab state, but it is a change from long-standing behavior — easy to flip if product prefers reset-to-All.

## Tests
19/19 in `src/app/searchResults` (7 new, red-first), including mutation-tested specs for the single-search-per-submit cycle (real digest fires the location event) and the route-exit guard.

### DIFF
--- a/src/app/searchResults/searchResults.component.js
+++ b/src/app/searchResults/searchResults.component.js
@@ -82,6 +82,12 @@ class SearchResultsController {
   // Re-read the URL params and re-run the search if they changed
   // (e.g. browser back/forward). No-op for our own URL updates.
   syncFromLocation() {
+    // When navigating away, $locationChangeSuccess fires before this scope is
+    // destroyed. Ignore location changes that leave the search-results page so
+    // we don't wipe the params and fire a stray search request.
+    if (!includes(this.$location.path(), 'search-results')) {
+      return;
+    }
     const params = this.$location.search();
     if (
       (params.q || undefined) === (this.searchParams.keyword || undefined) &&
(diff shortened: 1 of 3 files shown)


dr-bizz:
@wrandall22 Do you know if there was a reason why we initially had the page refresh on search?

wrandall22:
Not sure about page refresh specifically, but I imagine we need to keep `q` in the URL for analytics purposes, which may go away if you remove the normal form submission?

dr-bizz:
I'm going to close this out, as I also thought that we would lose the abilty to track searches

PR #1289, opened 2026-06-10
EP-2480 - Trace branded checkout cortex requests in DataDog RUM

dr-bizz:
## Jira
[EP-2480](https://jira.cru.org/browse/EP-2480)

## Problem
`datadog.config.js` built `allowedTracingUrls` from the environment `apiUrl` inside an Angular `.config()` block — which runs at bootstrap, **before** the `<branded-checkout api-url="...">` override is applied in the component's `$onInit`. So on third-party sites DataDog watched `give.cru.org/cortex` while all the real requests went to e.g. `brandedcheckout.jesusfilm.org/cortex` — branded checkout network requests were never instrumented.

## Fix
- At config time, read the `api-url` attribute straight from the DOM (`document.querySelector('branded-checkout')`) — safe because the loader bootstraps after the page is loaded.
- The attribute may omit the protocol (per README), so `normalizeApiUrl` was extracted from `branded-checkout.component.js` into a shared `common/lib/normalizeApiUrl.js` (verbatim; the component delegates to it) and matching is done protocol-relatively via a DataDog match **function** (supported by the installed @datadog/browser-rum v5.16 — `MatchOption` accepts `(url) => boolean`).
- The branded URL is **added alongside** the environment cortex URL rather than replacing it, so a stray branded element can never silently disable main-site tracing.
- Main give site behavior unchanged (no element → env apiUrl only; the appended `/cortex` also enforces the host boundary against prefix collisions).

## Tests
4 new specs written red-first, with a `wouldTrace` helper that mirrors DataDog's actual `matchList` semantics (verified against the installed v5.16 source). 87/87 green across the datadog and branded suites.

### DIFF
--- a/src/app/branded/branded-checkout.component.js
+++ b/src/app/branded/branded-checkout.component.js
@@ -17,6 +17,7 @@ import checkoutService from 'common/services/checkoutHelpers/checkout.service';
 import brandedAnalyticsFactory from './analytics/branded-analytics.factory';
 
 import 'common/lib/fakeLocalStorage';
+import normalizeApiUrl from 'common/lib/normalizeApiUrl';
 
 import template from './branded-checkout.tpl.html';
 
@@ -80,20 +81,7 @@ class BrandedCheckoutController {
   }
 
   normalizeApiUrl(url) {
-    if (!url) {
-      return url;
-    }
-
-    // Remove trailing slashes
-    url = url.replace(/\/+$/, '');
-
-    // Remove protocol if present
-    url = url.replace(/^https?:/, '');
-
-    // Remove slash before query parameters if present
-    url = url.replace(/\/\?/, '?');
-
-    return url.startsWith('//') ? url : '//' + url;
+    return normalizeApiUrl(url);
   }
 
   formatDonorDetails() {
--- a/src/common/datadog.config.js
+++ b/src/common/datadog.config.js
@@ -11,11 +11,15 @@ const dataDogConfig = /* @ngInject */ function (envServiceProvider) {
     const brandedCheckoutElement = document.querySelector('branded-checkout');
     const brandedCheckoutApiUrl =
       brandedCheckoutElement && brandedCheckoutElement.getAttribute('api-url');
-    const apiUrl = brandedCheckoutApiUrl
-      ? normalizeApiUrl(brandedCheckoutApiUrl)
-      : envServiceProvider.read('apiUrl');
+    // Always trace the env cortex URL; additionally trace the branded one when present
+    const apiUrls = [envServiceProvider.read('apiUrl')];
+    if (brandedCheckoutApiUrl) {
+      apiUrls.push(normalizeApiUrl(brandedCheckoutApiUrl));
+    }
     // The api-url attribute may omit the protocol, so match request urls protocol-relatively
-    const cortexUrl = apiUrl.replace(/^https?:/, '') + '/cortex';
+    const cortexUrls = apiUrls.map(
+      (apiUrl) => apiUrl.replace(/^https?:/, '') + '/cortex',
+    );
     const config = {
       applicationId: '3937053e-386b-4b5b-ab4a-c83217d2f953',
       clientToken,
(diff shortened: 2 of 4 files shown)


src/common/datadog.config.js
```diff
@@ -1,16 +1,37 @@
...
+    const brandedCheckoutApiUrl =
+      brandedCheckoutElement && brandedCheckoutElement.getAttribute('api-url');
+    // Always trace the env cortex URL; additionally trace the branded one when present
+    const apiUrls = [envServiceProvider.read('apiUrl')];
+    if (brandedCheckoutApiUrl) {
+      apiUrls.push(normalizeApiUrl(brandedCheckoutApiUrl));
+    }
+    // The api-url attribute may omit the protocol, so match request urls protocol-relatively
+    const cortexUrls = apiUrls.map(
+      (apiUrl) => apiUrl.replace(/^https?:/, '') + '/cortex',
```
canac: Why are we stripping the protocol? The `apiUrl`s in `app.config.js` include a protocol.

src/common/datadog.config.js
```diff
@@ -1,16 +1,37 @@
...
+    );
     const config = {
       applicationId: '3937053e-386b-4b5b-ab4a-c83217d2f953',
       clientToken,
       site: 'datadoghq.com',
       service: 'give-web',
       env: envServiceProvider.get(),
-      allowedTracingUrls: [envServiceProvider.read('apiUrl') + '/cortex'],
+      allowedTracingUrls: [
+        (url) =>
```
canac: So this is a function that determines whether the URL should be traced? Is there a reason why we can't just pass `cortexUrls` to `allowedTracingUrls`?

src/app/branded/branded-checkout.component.js
```diff
@@ -80,20 +81,7 @@ class BrandedCheckoutController {
...
-    url = url.replace(/\/+$/, '');
-
-    // Remove protocol if present
-    url = url.replace(/^https?:/, '');
-
-    // Remove slash before query parameters if present
-    url = url.replace(/\/\?/, '?');
-
-    return url.startsWith('//') ? url : '//' + url;
+    return normalizeApiUrl(url);
```
canac: This this method just delegates to `normalizeApiUrl`, I'd remove it and call `normalizeApiUrl` instead of `this.normalizeApiUrl`.

src/common/lib/normalizeApiUrl.js
```diff
@@ -0,0 +1,18 @@
+// Normalizes an API URL into a protocol-relative URL without trailing slashes,
+// e.g. 'https://give.domain.com/' or 'give.domain.com' -> '//give.domain.com'
```
canac, line 2: I'm not sure if now is the time to change it, but protocol-relative URLs are an anti-pattern now. If a branded checkout page allowed access over HTTP, all of their personal information would be sent to Cortex Gateway unencrypted. We want to use HTTPS whenever it is available.

canac (commented):
This looks like it works. I'd like to advocate for moving away from protocol-relative API URLs. I didn't realize we were doing that until this PR.

Did you get a chance to test this change? Maybe by deploying to staging and setting up a local override in the HTML to point JesusFilm's branded checkout at the staging script?

Misconfigured tracing URL whitelists broke things one time, so I want to be careful.

src/common/datadog.config.js
dr-bizz: Reworked in 97df25ee — the env apiUrl protocol is no longer touched. Only the branded api-url attribute (which may omit a protocol per the README) gets normalized, and it is now resolved to an absolute https URL instead of being matched protocol-relatively.

src/common/datadog.config.js
dr-bizz: Done — allowedTracingUrls now gets plain URL strings. I checked the installed v5.16 source: the tracer calls matchList(..., useStartsWith: true), so string entries prefix-match request URLs and the match function was unnecessary.

src/app/branded/branded-checkout.component.js
dr-bizz: Removed — the call site uses the imported function directly, and the specs moved to src/common/lib/normalizeApiUrl.spec.js.

src/common/lib/normalizeApiUrl.js
dr-bizz: Agreed. This PR does not change the checkout runtime behavior, but the DataDog config now resolves the branded URL to https rather than matching protocol-relatively. Moving branded checkout itself off protocol-relative URLs seems like a separate ticket.

dr-bizz:
Review comments addressed in 97df25ee — allowedTracingUrls is now plain URL strings and the delegating method is gone. Re: live testing — I have not run it against a real branded checkout page yet; planning to verify on staging with a local override pointing a branded page at the staging script before merging.

PR #1290, opened 2026-06-10
EP-1933 - Stack one-time gift modal buttons on small screens

dr-bizz:
## Jira
[EP-1933](https://jira.cru.org/browse/EP-1933)

## Problem
On iOS, the bottom-of-screen Safari menu target area hides the side-by-side (`col-xs-6`) CTA buttons in the 1-Time Bulk Gift modal, making it hard to continue. The gift configuration modal was already fixed years ago by stacking full-width buttons (`btn-block-xs` pattern in `_cart.scss`), but this modal never got it.

## Fix
Applied the existing stacked pattern to **all four** footers of the giveOneTimeGift modal (step 1 select/search, step 2 amounts, and the error-state footer):
- < 550px: buttons stack full-width with the primary action on top (above Safari's bar) and Cancel at the bottom, with the existing 10px sibling spacing.
- ≥ 550px: visual layout unchanged (secondary left, primary right; the error footer keeps Cancel | Retry, Continue).
- SCSS: widened the float reset to cover `pull-left` (was `pull-right` only — safely additive, no other co-occurrence in the codebase) and restored Bootstrap's 5px gap between two floated right-side buttons in the error footer.
- All `ng-click`/`ng-disabled`/`translate`/aria attributes preserved verbatim.

## Notes for QA / follow-up
- Final verification on an actual iOS device/simulator recommended (specs are controller-only; layout verified against the compiled CSS rules).
- Tab order now puts the primary CTA first while desktop shows it on the right — a deliberate trade-off so the primary stacks on top on mobile (mild WCAG 2.4.3 note).
- The review inventoried ~16 other yourGiving modal footers (editRecurringGifts, stopStartRecurringGifts) with the same `col-xs-6` pattern — same iOS issue, candidates for a follow-up ticket.

## Tests
30/30 across the giveOneTimeGift suites (controller specs; templates aren't rendered by existing specs).

Screenshot 2026-06-12 at 9 35 21 AM

Screenshot 2026-06-12 at 9 35 16 AM

### DIFF
--- a/src/app/profile/yourGiving/giveOneTimeGift/giveOneTimeGift.modal.tpl.html
+++ b/src/app/profile/yourGiving/giveOneTimeGift/giveOneTimeGift.modal.tpl.html
@@ -53,20 +53,25 @@
     </div>
   </div>
   <div class="modal-footer" ng-switch-when="errorLoadingRecentRecipients">
-    <div class="col-xs-6">
-      <a class="btn btn-default" ng-click="$ctrl.dismiss()" translate>
-        Cancel
-      </a>
-    </div>
-    <div class="col-xs-6 text-right">
-      <a
-        class="btn btn-primary"
-        ng-click="$ctrl.loadRecentRecipients()"
-        translate
-        >Retry</a
-      >
-      <a class="btn btn-default" ng-click="$ctrl.next()" translate>Continue</a>
-    </div>
+    <a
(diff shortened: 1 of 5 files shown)


wjames111 (approved):
This looks great, really nice improvement over the left and right buttons!

PR #1291, opened 2026-06-10
EP-2118 - Make the Other amount option visible and labeled

dr-bizz:
## Jira
[EP-2118](https://jira.cru.org/browse/EP-2118)

## Problem
Per the ticket, 30–50% of new donors couldn't find the "Other" amount option: it was a bare text input whose only label was a `placeholder` — which vanishes on focus, is low-contrast, and doesn't read as a peer of the suggested-amount buttons.

## Fix
All **three** template variants in `suggestedGiftAmounts.tpl.html` (default button-style, v3 $-prefixed, non-v3 radio list) now have a visible, persistent **"Other amount"** label styled consistently with the amount options:
- New `OTHER_AMOUNT` translate key in both en and es tables ("Otra cantidad"); the old placeholder relied on an `'Other'` key that never existed in either table, so this also fixes a latent untranslated-in-Spanish bug.
- Accessibility: `aria-labelledby` replaces the placeholder/aria-label as the accessible name everywhere.
- Radio variant: the label's `for` now targets the amount input (was implicitly targeting the decorative radio), so clicking the label text focuses the input and correctly drives the existing `customInputActive` wiring — fixing a latent visual-desync bug.
- Orphaned `OTHER_PLACEHOLDER`/`CUSTOM_AMOUNT` keys removed (zero remaining references).
- SCSS scoped to the live `.suggested-gift-amounts` selector (review caught that the pre-existing `.give-modal-panel .custom-amount` block was dead CSS — deleted); active-state label color matches the existing gold-border/light-background active treatment.
- Behavior/controller untouched — existing focus/radio wiring reused.

## QA note
A quick visual smoke test of the gift-config modal (default variant) at desktop and <768px is recommended, especially with the longer Spanish label.

## Tests
108/108 across `src/app/productConfig` (4 new template-render specs: label presence + aria wiring + placeholder removal per variant, plus a behavioral spec for focus → customInputActive).

### DIFF
--- a/src/app/productConfig/giftPanels/suggestedGiftAmounts/suggestedGiftAmounts.component.spec.js
+++ b/src/app/productConfig/giftPanels/suggestedGiftAmounts/suggestedGiftAmounts.component.spec.js
@@ -35,8 +35,9 @@ describe('suggestedGiftAmounts', () => {
       $rootScope = _$rootScope_;
     }));
 
+    let scope;
     const compileComponent = (scopeValues) => {
-      const scope = $rootScope.$new();
+      scope = $rootScope.$new();
       Object.assign(
         scope,
         {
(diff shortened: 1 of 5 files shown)


dr-bizz:
Closing this as I think this has already been addressed

PR #1292, opened 2026-06-10
EP-2114 - Warn when adding a designation already in recurring gifts

dr-bizz:
## Jira
[EP-2114](https://jira.cru.org/browse/EP-2114)

## Problem
Donors can add a designation to a new recurring donation when they already give to that designation in another recurring gift. Siebel stops these by placing the gift on Hold, but the frontend gave no hint — the donor finds out via a held gift.

## Fix (non-blocking warning — backend remains authoritative)
In the "Edit my recurring donations" step 2 pickers, using a `designationNumber` intersection with the donor's existing recurring gifts (already in memory from step 1):
- **Recent recipients picker:** a per-row `label-warning` badge — "Already in your recurring gifts" — visible before selection, transcluded into `gift-list-item` (same pattern steps 1/3 already use; no shared-component changes).
- **Search picker:** results render in the shared `giftSearchView`, so instead of touching the shared component, selecting a duplicate shows an `alert-warning` listing the duplicated recipients; it clears on deselection.
- Continue is untouched — purely informational, no flow changes.
- New `ALREADY_IN_RECURRING_GIFTS` translate key in en + es.
- Handles the no-existing-gifts path safely (no badge, no errors).

## Open question for backend folks (from review)
The warning checks **active** recurring gifts only (step 1's `managerecurringdonations` list). If Siebel's duplicate/Hold logic also counts held/suspended/future gifts, the warning should be fed a wider gift-type list — easy follow-up once confirmed.

## Tests
8 new controller specs written red-first (detection logic both pickers, undefined-gifts safety, binding config). 70/70 green across the editRecurringGifts suites.

### DIFF
--- a/src/app/profile/yourGiving/editRecurringGifts/step2/searchRecipients.tpl.html
+++ b/src/app/profile/yourGiving/editRecurringGifts/step2/searchRecipients.tpl.html
@@ -29,6 +29,16 @@
           select-label="{{'Add' | translate}}"
         >
         </gift-search-view>
+        <div
+          class="alert alert-warning"
+          role="alert"
+          ng-if="$ctrl.duplicatedRecipients.length > 0"
+        >
+          <p ng-repeat="gift in $ctrl.duplicatedRecipients">
+            {{gift.designationName}} &mdash; {{'ALREADY_IN_RECURRING_GIFTS' |
+            translate}}
+          </p>
+        </div>
       </div>
     </div>
   </div>
--- a/src/app/profile/yourGiving/editRecurringGifts/editRecurringGifts.modal.tpl.html
+++ b/src/app/profile/yourGiving/editRecurringGifts/editRecurringGifts.modal.tpl.html
@@ -101,6 +101,7 @@
   <step-2-add-recent-recipients
     ng-switch-when="step2AddRecentRecipients"
     recent-recipients="$ctrl.recentRecipients"
+    recurring-gifts="$ctrl.recurringGifts"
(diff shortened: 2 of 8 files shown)


dr-bizz:
I have no way to test whether this works, as I can't reach this step. Only a few people can

src/app/profile/yourGiving/editRecurringGifts/step2/searchRecipients.tpl.html
```diff
@@ -29,6 +29,16 @@ <h3 translate>Give a new recurring gift</h3>
           select-label="{{'Add' | translate}}"
         >
         </gift-search-view>
+        <div
```
canac, line 32: optional: Should we put this warning closer to the staff person's card? If search returns lots of entries, the user won't see this warning until they scroll all the way to the bottom. Although they have to scroll to the bottom to continue, so maybe it's fine as-is.

canac (approved):
The change looks good. Should we loop in Sean McKenna for this user-facing change?

src/app/profile/yourGiving/editRecurringGifts/step2/searchRecipients.tpl.html
dr-bizz: Keeping it as-is for now — search results render inside the shared giftSearchView component, so a per-row badge there would mean changing the shared component, which this PR deliberately avoids. And as you noted, users have to scroll to the bottom to hit Continue anyway. Can revisit if donors miss it.

